PR #109, opened 2014-11-22
Experiment: Run tasks on servers asynchronously

gordalina:
Currently deployer deploys in a synchronous manner, this works well with few servers, but when the more servers we are deploying, the longer it takes.

Running every task in every server asynchronously will make this much faster. Initial experiments gave me a `1.5x` speed up (3.46 seconds compared to 5.46 seconds) with only two servers, this speed up increases with every server added.

**Issues**
- [ ] I don't think it is even possible to install dependencies on windows because it depends on [spork](https://github.com/phresque/spork) which requires `ext-pcntl` and `ext-posix`. This issue is easily solved (no async on windows)
- [ ] Tasks which require user input are also run asynchronously, which makes the cli interface unintuitive. This can also be solved. 
- [ ] Other issues that I'm not aware of?


gordalina:
Closing because its already implemented

PR #130, opened 2015-01-08
Use mustRun instead of run and catch exception

fran6co:
It does the same thing but it gives out a better exception object.


elfet:
Thanks!

PR #131, opened 2015-01-09
Added branch to git clone execute

cafferata:
Expected by documentation that when the branch is set this wil be checkout
- http://deployer.in/stages.html
- http://deployer.in/tasks.html#using-input-options


elfet:
Done.

PR #135, opened 2015-01-15
Changed phpseclib dependency to php5 branch

ArjandeV:
The phpseclib php5 branch autoloads according to PSR-4 namespacing so the client encryptionalgorithms array isn't empty.


elfet:
Cool! Thanks!

PR #143, opened 2015-01-20
Added possibility to add arguments/options

BenjaminPaap:
We felt the need to add some arguments/options to the console command during our development of our recipes. Maybe others need this functionality, too. This would allow you to do something like this:

In a recipe:
```php
option('tag', array(
    'description' => 'Tag to checkout during `git clone`'
));
```

When deploying:
```bash
deployer.phar --server=prod --tag=v1.0
```

/cc @mabrahamde

### DIFF
--- a/src/Deployer.php
+++ b/src/Deployer.php
@@ -103,6 +103,43 @@ class Deployer
     }
 
     /**
+     * Adds a global argument
+     *
+     * @param string $name
+     * @param array  $config
+     */
+    public function addArgument($name, $config)
+    {
+        $argument = new Console\Input\InputArgument(
+            $name,
+            $config['mode'],
+            $config['description'],
+            $config['default']
+        );
+
+        $this->console->getDefinition()->addArgument($argument);
+    }
+
+    /**
+     * Adds a global option
+     *
+     * @param string $name
+     * @param array  $config
+     */
+    public function addOption($name, $config)
+    {
+        $option = new Console\Input\InputOption(
+            $name,
+            $config['shortcut'],
+            $config['mode'],
+            $config['description'],
+            $config['default']
+        );
+
+        $this->console->getDefinition()->addOption($option);
+    }
+
+    /**
      * @return Console\Input\InputInterface
      */
     public function getInput()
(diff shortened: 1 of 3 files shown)


src/Deployer.php
```diff
@@ -103,6 +103,56 @@ public function addConsoleCommands()
...
 
     /**
+     * Adds a global argument
+     *
+     * @param string $name
+     * @param array  $config
+     */
+    public function addArgument($name, array $config)
+    {
+        $config = array_merge(array(
```
elfet: Minimum PHP version 5.4. So `[...]` is better than `array(...)`.

src/Deployer.php
```diff
@@ -103,6 +103,56 @@ public function addConsoleCommands()
     }
 
     /**
+     * Adds a global argument
+     *
+     * @param string $name
+     * @param array  $config
+     */
+    public function addArgument($name, array $config)
```
elfet: Is `Deployer` proper place for that?

src/functions.php
```diff
@@ -420,3 +420,37 @@ function env($name = null, $value = null)
...
+ *
+ * @param string $name
+ * @param array  $argument
+ *
+ * @return mixed
+ */
+function argument($name, array $argument = null)
+{
+    if (null === $argument) {
+        return Deployer::get()->getInput()->getArgument($name);
```
elfet: Better use `input()` than `Deployer::get()->getInput()`. It takes `Input` from current tasks context.

elfet:
Also what about parallel execution? Can you implement passthru for options/arguments to worker?
See [ParallelExecutor](https://github.com/deployphp/deployer/blob/master/src/Executor/ParallelExecutor.php#L60) and [Worker](https://github.com/deployphp/deployer/blob/master/src/Console/WorkerCommand.php).

src/Deployer.php
BenjaminPaap: Any suggestions on a new place for that?

src/functions.php
BenjaminPaap: When doing this I have to inform every `TaskCommand` about the created arguments/options in order to have them available in that specific context.

elfet:
Will implement it myself.

PR #144, opened 2015-01-20
Fixed command configuration

BenjaminPaap:
The symfony console has a `configure`-method which should be used to configure the command.


elfet:
Great! :beer: Can you fix other commands too?

Issue #146, opened 2015-01-21
Set custom timeouts for Server::run()


mabrahamde:
Deploying using `Deployer\Server\Local` Server i am running into a timeout.

Tasks like composer install can take really long. However in `Symfony\Component\Process` the method `run()` uses a default timeout of 60 seconds. Sometimes 60s is too short. I suggest two solutions for this:

a) set higher default values `Deployer\Server\Local`:

```
    const TIMEOUT = 300;
```

b) extend `Deployer\Server\ServerInterface` and add a new parameter to the `run()` method inside `functions.php`:

```
    public function run($command, $timeout = 60)
```

elfet:
I think `a` is better.

EkAndreas:
Please add this option for runLocally...

maximTarleckiy:
If you want run deployment task for local server which already used run() function, turns out that you should change task to use runLocally('command...', 0) to disable timeout for command. Maybe make sense to set timeout option for particular server?

peterjaap:
+1 to add custom timeouts to `run()` as well, just like `runLocally()`. We pretty often need to import database into our staging environment, which we control through Deployer.

pluseg:
@elfet please stick it with https://github.com/deployphp/deployer/issues/924 (seems as full duplicate) and https://github.com/deployphp/deployer/pull/974.

elfet:
Command (remote/locale) timeout refactoring is planned for v5. Have some ideas implement to.

elfet:
Done in #1092

PR #147, opened 2015-01-21
Set local process timeout to a higher value

mabrahamde:
Tasks like `composer install` can take really long. However in `Symfony\Component\Process` the method `run()` uses a default timeout of 60 seconds. Sometimes this value is too low. 

In #146 we discussed just to set a higher timeout value.

### DIFF
--- a/src/Server/Local.php
+++ b/src/Server/Local.php
@@ -11,7 +11,7 @@ use Symfony\Component\Process\Process;
 
 class Local implements ServerInterface
 {
-    const TIMEOUT      = 300;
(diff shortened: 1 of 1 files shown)


src/Server/Local.php
```diff
@@ -11,6 +11,8 @@
 
 class Local implements ServerInterface
 {
+    const TIMEOUT      = 300;
```
elfet: Please, remove extra spaces.

elfet:
Thanks! :100:

Issue #148, opened 2015-01-23
Submodule init problem


tomaj:
Hi

I have problem in task deploy:update_code when my repo is using some submodules.
Deploy skript is trying to run something like this:
```
cd /home/user/mytest && git clone --recursive -q git@bitbucket.org:owner/repo.git /home/user/mytest/releases/2015012324229
```
But it fails :-(
```
[RuntimeException]
  Cloning into 'htdocs/somefolder1'...
  Cloning into 'htdocs/somefolder2'...
  Cloning into 'htdocs/somefolder3'...
  Cloning into 'htdocs/somefolder4'...
```

I am using *php ssh2* lib. I have tried to debug this issue and it fails in ```vendor/elfet/php-ssh/src/Ssh/Exec.php:30``` when command returns something to output it will consider it as fail and. git command outputs information with "Cloning into" for submodules (event you pass -q variable). There is no permission problem, I can run command that I mentioned on the top and it runs ok (but output information about initialising submodules)

Anyone have any ideas how to fix this?

elfet:
Please, run deploy with -vvv attr and show output.

tomaj:
Here it is:

```
$php vendor/bin/dep deploy:update_code -vvv

Updating code
Run task deploy:update_code on server main
[mytest.sk] cd /home/forge/mytest && git clone --recursive -q git@bitbucket.org:owner/streamapp.git /home/forge/mytest/releases/2015012329915
Run task deploy:rollback on server main
Rolling back to previous release on server main
[mytest.sk] cd /home/forge/mytest && rm -rf /home/forge/mytest/releases/2015012329915
[mytest.sk] ::



  [RuntimeException]
  Cloning into 'htdocs/somefolder1'...
  Cloning into 'htdocs/somefolder2'...
  Cloning into 'htdocs/somefolder3'...
  Cloning into 'htdocs/somefolder4'...



Exception trace:
 () at /var/www/streamapp/htdocs/Nette/libs/vendor/elfet/php-ssh/src/Ssh/Exec.php:28
 Ssh\Exec->run() at /var/www/streamapp/htdocs/Nette/libs/vendor/deployer/deployer/src/Server/Ssh2.php:94
 Deployer\Server\Ssh2->run() at /var/www/streamapp/htdocs/Nette/libs/vendor/deployer/deployer/src/functions.php:115
 run() at /var/www/streamapp/htdocs/Nette/libs/vendor/deployer/deployer/recipe/common.php:64
[...]

PR #151, opened 2015-01-29
Change output for cloning repository

tomaj:
Changed output when cloning repository, because of initialization of submodule. Git always output some text, and this is actually considered as error. Update provide possibility to deploy repository with submodules.


tomaj:
This fixes issue #148

elfet:
Thanks a lot!

PR #159, opened 2015-02-04
change elfet/php-ssh to original repo herzult/php-ssh

sidz:
Changed not existing elfet/php-ssh repo on herzult/php-ssh


composer.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "elfet/deployer",
+    "name": "sidz/deployer",
```
sidz: it's wrong commit

PR #163, opened 2015-02-06
Shallow clone

KennedyTedesco:
`--depth` create a shallow clone with a history truncated to a limited number of revisions.


elfet:
Thanks!

PR #164, opened 2015-02-06
PHPUnit improvements

KennedyTedesco:
https://phpunit.de/manual/current/en/textui.html#textui.clioptions

Before:

image

After:

image


elfet:
Nice!

Issue #128, opened 2015-01-01
Don't download composer.phar for every release if it is globally available


Flyingmana:

elfet:
Yes, it better to check it first.

elfet:
But how to tets it?

Flyingmana:
first solution which comes into my mind, test what the command ```composer.phar --version``` returns, maybe also allow to configure how the executable is named, and where it should be (I always install it as part of my environment directly after php gets installed)

elfet:
Done.

PR #166, opened 2015-02-07
Composer - Try to use the global installation

KennedyTedesco:
Like suggested here: https://github.com/deployphp/deployer/issues/128

What you guys think? I have tested and worked fine.

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -131,13 +131,23 @@ task('deploy:writeable', function () {
  */
 task('deploy:vendors', function () {
     $prod = get('env');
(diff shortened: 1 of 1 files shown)


elfet:
I need some time to merge it without conflicts with my new common.php.

tomzx:
Why download composer.phar to the `release_path` when you could/should download it to the `shared_path` instead?

Also, it'd be nice if we could not pollute commands with framework specific environment variables (such as `SYMFONY_ENV`).

KennedyTedesco:
@tomzx I'm ok with that. :+1:

elfet:
Done!

KennedyTedesco:
Awesome! :smile:

PR #173, opened 2015-02-13
Use symbolic links instead of rm/mv to set the current release.

tomzx:
Create shared directories in the release path.
Create writable directories (if needed).


gordalina:
:+1: for atomic deploy

elfet:
Thanks! Will merge manually! Now have my own commits for `common.php` with some refactoring. Thanks you for your work!

elfet:
Done!

Issue #177, opened 2015-02-14
Why was the cd() function removed?


tomzx:
I think that the following syntax is much nicer than the second one.

```php
cd('{some_path}');
run('my_super_command'); // in {some_path}
```

```php
run('cd {some_path} && my_super_command');
```

Thus my question: why was cd() removed in v3?

elfet:
I removed `cd` function to simplify `run` function. I think `run('cd {some_path} && my_super_command')` more readable than. But if is possible to bring back `cd` function.

tomzx:
Most tasks generally occur within a specific directory. I think the `cd` syntax follows the DRY (Don't Repeat Yourself) principle which is a good thing. Furthermore, since your preference (`cd`-ing on every command) still would work with `cd` support, it would probably be better to have it than not.

```php
cd('{some_path}');
run('my_super_command');
run('my_super_command');
run('my_super_command');
run('my_super_command');
```

```php
run('cd {some_path} && my_super_command');
run('cd {some_path} && my_super_command');
run('cd {some_path} && my_super_command');
run('cd {some_path} && my_super_command');
```

I do understand that the removal of `cd` (and the working path paradigm) was to simplify the `run` function. I guess the problem is that the phpseclib (and I assume the ssh2 extension) both run one-off commands where you are back to the home directory after each command, which means if we want emulate a working directory, we have to do it ourselves.

elfet:
Ok, will bring back `cd` func. :+1:

elfet:
`cd` is back!

PR #180, opened 2015-02-14
Add cd (change directory) and within (run commands within a directory and revert to the previous workingPath)

tomzx:
Here's my take on the `cd` function. I've also added a `within` function which allows the user to provide a callback:

```php
<?php

task('test', function() {
	cd('my_dir');
	within('{some_dir}', function() {
		run('some command');
	});
	// back in my_dir
});
```

After the `within` function is called, the environment is reverted to the previous working path.

This current implementation only support going to an absolute path. We may consider changing the implementation to support relative navigation if that is something people would want (I don't think that is the case though).

Reference: #177.

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -141,6 +141,40 @@ function after($it, $that)
 }
 
 /**
+ * Change the current working directory.
+ *
+ * @param string $path
+ */
+function cd($path)
+{
+    env('working_path', $path);
+}
+
+/**
+ * Execute a callback within a specific directory and revert back to the initial working directory.
+ *
+ * @param string $path
+ * @param callable $callback
+ */
+function within($path, $callback)
+{
+    $lastWorkingPath = workingPath();
+    env()->set('working_path', $path);
+    $callback();
+    env()->set('working_path', $lastWorkingPath);
+}
+
+/**
+ * Return the current working path.
+ *
+ * @return string
+ */
+function workingPath()
+{
+    return env()->get('working_path', env()->get(Environment::DEPLOY_PATH));
+}
+
+/**
  * Run command on server.
  *
  * @param string $command
@@ -149,7 +183,10 @@ function after($it, $that)
 function run($command)
 {
     $server = Context::get()->getServer();
-    $command = Context::get()->getEnvironment()->parse($command);
+    $command = env()->parse($command);
+    $workingPath = workingPath();
+
+    $command = "cd $workingPath && $command";
 
     if (isVeryVerbose()) {
         writeln("<comment>Run</comment>: $command");


src/functions.php
```diff
@@ -140,6 +140,24 @@ function after($it, $that)
     $afterScenario->addAfter($scenario);
 }
 
+function cd($path)
```
elfet: Add PhpDoc here, please.

src/functions.php
```diff
@@ -140,6 +140,24 @@ function after($it, $that)
...
+{
+    env('workingPath', $path);
+}
+
+function within($path, $callback)
```
elfet: Add PhpDoc here, please.

src/functions.php
```diff
@@ -140,6 +140,24 @@ function after($it, $that)
...
+{
+    $lastWorkingPath = workingPath();
+    env()->set('workingPath', $path);
+    $callback();
+    env()->set('workingPath', $lastWorkingPath);
+}
+
+function workingPath()
```
elfet: Add PhpDoc here, please.

src/functions.php
```diff
@@ -140,6 +140,24 @@ function after($it, $that)
     $afterScenario->addAfter($scenario);
 }
 
+function cd($path)
+{
+    env('workingPath', $path);
```
elfet: Lets rename `workingPath` to `working_path` (like `deploy_path`).

src/functions.php
```diff
@@ -149,7 +167,10 @@ function after($it, $that)
 function run($command)
 {
     $server = Context::get()->getServer();
-    $command = Context::get()->getEnvironment()->parse($command);
+    $command = env()->parse($command);
+    $workingPath = workingPath();
+
+    $command = "cd $workingPath && $command";
```
elfet: Now it does not possible to use `cd` like this:
```php
cd('{deploy_path}');
run('command');
```

tomzx: Interestingly enough, it does because `workingPath()` calls `get` on the environment, which does the following before returning the value `return $this->parse($value);`.

elfet: Correct. Sorry. Forgot :smiley:

tomzx:
I've forced pushed the changes you've asked. Please take the time to review them.

Thanks!

elfet:
Done :beer:

elfet:
Hmmm.

It's not possible to write like this:
```
cd('{release_path}');
```

only like this:
```
cd(env('release_path'));
```
@tomzx  Can you fix it?

tomzx:
@elfet Pretty sure that should work (at least it did when I wrote the functionality). I'll pull the latest changes and see if that still works.

elfet:
Hmmm, can you look at https://github.com/deployphp/deployer/blob/master/recipe/common.php#L134
Not working there :(

tomzx:
I assume you're getting this error: `PHP Fatal error:  Maximum function nesting level of '100' reached, aborting!`?

elfet:
Yes.

Issue #190, opened 2015-02-20
Replace phar build script with box


tomzx:
https://github.com/box-project/box2

elfet:
Why?

tomzx:
We could benefit from a few things such as gzip compression.

It would also convert a build script into a simple configuration file.

tomzx:
Closing, see discussion @ #193.

PR #193, opened 2015-02-22
Replace phar build script with box-project/box2 build tool.

tomzx:
Adds a box.json file that should replace the configuration used in the build file.
Removes the old build file (since it is replaced by the configuration file).

File size goes from 3 MB to ~485 KB.

Reference: #190


elfet:
What if i add `$phar->compressFiles(Phar::GZ);` to my build script? 
Now deployer.org site uses my `build` script, so we need to refactoring them too.
What is the benefit ща migration to `box2`?

elfet:
New size of `deployer.phar` 825KB now.

I think we need to reduce number of dependencies.

gordalina:
Nice job @tomzx 
Box really makes it simple to package an app. 
 
I would like to add that enabling gzip compression could lead to php installations (without gzip extension enable), to fail loading the phar. Also Composer [had an issue](https://github.com/composer/composer/blob/1.0.0-alpha9/src/Composer/Compiler.php#L131) with this before.

Regarding the size of the phar, I guess we (developers) have reasonably fast bandwidth to transfer a larger file, thus making it less relevant.

tomzx:
@elfet Why do you prefer your own build script to using a library that is used/tested by many? What would make you change your mind?

elfet:
Hmmm, i'll think about it :)

gpupo:
I agree with @tomzx. Nice job

elfet:
Box awesome project, but i do not what to touch what's already good working. 
Also i do not see any cons of this.

PR #196, opened 2015-02-25
Create path to shared dir in release dir if it does not exist.

tomzx:
If the shared paths do not exist in the release_path directory and those shared directories have a depth > 1, the directory structure must be created before symlinking can be done.


elfet:
Cool!

PR #197, opened 2015-02-25
Parse the target path given to cd to prevent recursive loop.

tomzx:
If cd was given {release_path} as target path, it would attempt to run a command to figure out what the value of {release_path} is, which, in the process, would also want to know the value of {release_path} and so on.

The fix simply parses the value as it is given, thus making the result static instead of a closure.


elfet:
Cool :+1: :smiley:

PR #199, opened 2015-02-26
Check the exit status of the command instead of stderr

cjsewell:
Composer now outputs non-error messages to stderr. The `run` function should be check the exit status of the command, not the error output.


elfet:
Cool. Thanks. Now work with errors?

cjsewell:
Yep it was annoying but some applications use std error even though its not an error. Checking the exit status is a more reliable way to tell if the command was successful. 0 means Ok, any other number means there was an error

elfet:
Cool! Thanks!

PR #203, opened 2015-02-27
Extract $releaseDatetime in order to share a common release datetime between parallel deployments.

tomzx:


recipe/common.php
```diff
@@ -71,16 +71,10 @@
...
-
+$release = date('YmdHis');
+task('deploy:release', function () use ($release) {
     $releasePath = "{deploy_path}/releases/$release";
 
-    $i = 0;
-    while (is_dir(env()->parse($releasePath)) && $i < 42) {
-        $releasePath .= '.' . ++$i;
-    }
-
```
elfet: Why you delete this?

elfet:
Yes, this is good feature. But it will not working in parallel mode. I will implement distributed Deployer parameters, so it will be possible to implement correct solution for this in parallel deployment mode.

recipe/common.php
tomzx: Is it really needed? And why stop at 42?

elfet: Yes, it's needed in tests. Tree time runs `deploy:release` and check all was correct. 
[42](https://simple.wikipedia.org/wiki/42_(answer))

elfet:
Build fails.

tomzx:
@elfet With the latest change, I've created an global variable called `release_identifier`, which is more generic than before, and allows users to redefine how they want their release identifier to work so they could use the old numbering system if they prefer it).

KingCrunch:
Or the VCSs revision/hash. Just saying :smile:

tomzx:
If I'm not mistaken, the way Capistrano does it is to append the release at the end of a file containing the list of releases. Whatever the identifier is, it'll be possible to figure out which release came before/after which one based on the ordering in the file. It shouldn't be too hard to implement this feature if enough people want it.

KingCrunch:
Just tried it myself: Capistrano3 keeps a log of revision it deployed, but that seems to be just a log.

```bash
$ cat revisions.log 
Branch master (at bdd236b) deployed as release 20150318214035 by sebastian
```
It still creates the actual deployments into folders based on the timestamp and delete them in order. However, that doesn't me `deployer` can't do it differently :smile:

elfet:
Will be implemented in Deployer v4.

PR #209, opened 2015-03-04
Return $result if getStdError() returns an empty string.

tomzx:
This is done to support applications which return a non-zero exit status but do not output error messages to stderr.


elfet:
Do you know application like that?

tomzx:
[Grunt](http://gruntjs.com/) for example.

gquemener:
I'd say better always display both stdout and stderr when throwing the RuntimeException.

gordalina:
:+1:

PR #217, opened 2015-03-18
Updated to PHP SSH 1.1.1

gquemener:
This version improves command execution error detection

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1,0 +1,1024 @@
+{
+    "_readme": [
+        "This file locks the dependencies of your project to a known state",
+        "Read more about it at http://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
+        "This file is @generated automatically"
+    ],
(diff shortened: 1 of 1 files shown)


gquemener:
See https://github.com/Herzult/php-ssh/pull/44

gquemener:
Hello @elfet,
Could you review this one please?

Main problem with php-ssh@1.1.0 is that it relies on stderr to check if executed command failed or not BUT since a few weeks, composer writes a lot of things to stderr that are not errors (see my previous comment for the reasons).

Basically, all deployer users that uses composer during deployment (so a good majority of its users) will get a RuntimeException during composer install step.

I've fixed the problem on php-ssh, all you have to do now is to update deployer, and then we can share a :beers:

Djeg:
:+1:

Shivoham:
:+1:

wysow:
:+1:

elfet:
Great! Sorry about delay. :beer:

Issue #211, opened 2015-03-06
Bash-syntax mistakenly taken for environment variable substitution


KingCrunch:
Version `3.0.0-beta.5`

```php
run('mkdir -p ./{last,current,new}');
```

This is valid bash syntax and should create the three folders at once, but it ends up in an exception.

```bash
[RuntimeException]                                         
  Environment parameter `last,current,new` does not exists.  
                                                             


Exception trace:
 () at phar:///usr/local/bin/deployer.phar/src/Server/Environment.php:55
 Deployer\Server\Environment->get() at phar:///usr/local/bin/deployer.phar/src/Server/Environment.php:93
 Deployer\Server\Environment->parse() at phar:///usr/local/bin/deployer.phar/src/functions.php:187
 run() at /home/krebs/Workspace/api-wrapper/deploy.php:28
 {closure}() at n/a:n/a
 call_user_func() at phar:///usr/local/bin/deployer.phar/src/Task/Task.php:52
 Deployer\Task\Task->run() at phar:///usr/local/bin/deployer.phar/src/Executor/SeriesExecutor.php:40
 Deployer\Executor\SeriesExecutor->run() at phar:///usr/local/bin/deployer.phar/src/Console/TaskCommand.php:105
 Deployer\Console\TaskCommand->execute() at phar:///usr/local/bin/deployer.phar/vendor/symfony/console/Symfony/Component/Console/Command/Command.php:253
 Symfony\Component\Console\Command\Command->run() at phar:///usr/local/bin/deployer.phar/vendor/symfony/console/Symfony/Component/Console/Application.php:874
 Symfony\Component\Console\Application->doRunCommand() at phar:///usr/local/bin/deployer.phar/vendor/symfony/console/Symfony/Component/Console/Application.php:195
 Symfony\Component\Console\Application->doRun() at phar:///usr/local/bin/deployer.phar/vendor/symfony/console/Symfony/Component/Console/Application.php:126
 Symfony\Component\Console\Application->run() at phar:///usr/local/bin/deployer.phar/src/Deployer.php:92
 Deployer\Deployer->run() at phar:///usr/local/bin/deployer.phar/bin/dep:62
```

elfet:
Hmmm. Well, need to use another syntax for env vars.

Now
```php
run('cd {deploy_path}/release & ...');
```

Maybe like that:
```php
run('cd {{deploy_path}}/release & ...');
```

oanhnn:
Vote +1 for 
```php
[...]

PR #218, opened 2015-03-20
Change syntax for env vars

oanhnn:
In this Pull request, I will implement feature of issue #211 
Change syntax for env vars to `{{ ... }}` and support allow whitespaces ( `{{deploy_path}}` and `{{ deploy_path }}` return same value).

### DIFF
--- a/src/Server/Environment.php
+++ b/src/Server/Environment.php
@@ -88,13 +88,20 @@ class Environment
     public function parse($value)
     {
         if (is_string($value)) {
-            if (preg_match_all('/\{(.+?)\}/', $value, $matches)) {
-                foreach ($matches[1] as $name) {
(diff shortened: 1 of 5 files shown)


src/Server/Environment.php
```diff
@@ -88,13 +88,20 @@ public static function setDefault($name, $value)
...
         return $value;
     }
+    
+    /**
+     * Replace env values callback for parse
+     *
+     * @param array $matchs
+     * @return mixed
+     */
+    private function _parse($matchs)
```
elfet: Please do not user under scope `_` for private methods. Rename it to `parseCallback`.

elfet:
Good one!

oanhnn:
Sorry @elfet ,
I try make all commits to single commit by `rebase` but it is failure. 
Please implement this feature yourself. I think single is best ^^

PR #219, opened 2015-03-23
Change syntax for env vars

oanhnn:
In this Pull request, I will implement feature of issue #211 
Change syntax for env vars to {{ ... }} and support allow whitespaces ( {{deploy_path}} and {{ deploy_path }} return same value).

Sorry by create other pull request.
This pull request is rebase of PR #218

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -33,13 +33,13 @@ task('rollback', function () {
     $releases = env('releases_list');
 
     if (isset($releases[1])) {
-        $releaseDir = "{deploy_path}/releases/{$releases[1]}";
+        $releaseDir = "{{deploy_path}}/releases/{$releases[1]}";
 
         // Symlink to old release.
-        run("cd {deploy_path} && ln -nfs $releaseDir current");
+        run("cd {{deploy_path}} && ln -nfs $releaseDir current");
 
         // Remove release
-        run("rm -rf {deploy_path}/releases/{$releases[0]}");
+        run("rm -rf {{deploy_path}}/releases/{$releases[0]}");
 
         if (isVerbose()) {
             writeln("Rollback to `{$releases[1]}` release was successful.");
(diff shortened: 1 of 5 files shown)


elfet:
Ok :)

oanhnn:
Dear @elfet ,
My pull request #219 has a problem.
With old parser (pattern `/\{(.+?)\}/`) detected all cases: 
```
{abcxyz}
{abc-xyz}
{abc_xyz}
{abc.xyz}
{abc@zyx}
```
but new parser (pattern `/\{\{\s*(\w+)\s*\}\}/`) only detected cases:
```
{{abcxyz}}
{{abc_xyz}}
```
I can fix this, but i want specific requirements for env var.
I think we should only allow:
```
{{abcxyz}}
{{abc_xyz}}
{{abc.xyz}}
```

gordalina:
I vote for same pattern as normal variables (implemented on #219).
Having a dot `.` on a variable could be used in the future to be a property of an array.

oanhnn:
@gordalina, Is your attention?
```php
env('abc', array('xyz' => 1));

echo env()->parse('{{abc.xyz}}');
```
output:
```
1
```

gordalina:
I didn't know about `parse`.
Still, I keep my opinion, and would suggest making `parse()` private and `env()` would call `parse()` [here](https://github.com/deployphp/deployer/blob/master/src/Server/Environment.php#L55)

This way, properties are part of the DSL and the user doesn't need to know wether it is necessary to parse them or not.

What are your thoughts @oanhnn @elfet?

oanhnn:
I think:
1. Method `parse` will be private. It is only call when using `run()`.
2. When call `env($abc, $xyz)` or `env()->set($abc, $xyz)`, variable `$abc` will check and only allow a-z, A-Z, 0-9 and underscore character ( `_` )
3. When call `env($abc)` or `env()->get($abc)`, variable `$abc` will check and only allow a-z, A-Z, 0-9, underscore character ( `_` ) and dot character ( `.` ) . If having a dot on variable `$abc` could be used in the future to be a property of an array.

gordalina:
@oanhnn exactly!

Issue #152, opened 2015-01-29
Generated shared file is empty


dirkluijk:
Hi,

I have a configuration file, say ``config.php`` which is shared among releases. The first time I deploy, it is generated in the ``/shared`` directory but it is still empty instead of having the original contents.

If this is a bug, I could provide a fix for this.

update: it looks like the same is happening with shared directories. They are empty as well. It looks like it was meant for generated files/directories, but I have some initial content stored in git.

elfet:
Yeah, i think it is feature good to have :) If you can create it - will be amazing! :gem:

dirkluijk:
Nice, will have a look at it soon, in a few days. For now I have moved the initial contents manually.

oanhnn:
I have same problem, and my solution is write a task `configure`
```php

/**
 * Make shared_dirs and configure files from templates
 */
task('configure', function () {
    
    /**
     * Compiler template of configure files
     * 
     * @param string $contents
     * @return string
     */
    $compiler = function ($contents) {
        if (preg_match_all('/\{\{(.+?)\}\}/', $contents, $matches)) {
            foreach ($matches[1] as $name) {
                $value = env()->get($name);
                if (is_null($value) || is_bool($value) || is_array($value) || is_string($value)) {
                    $value = var_export($value, true);
                }
                $contents = str_replace('{{' . $name . '}}', $value, $contents);
            }
        }
        return $contents;
    };
    
    // Create shared dir if does not exist
    foreach (get('shared_dirs') as $dir) {
        run("mkdir -p {deploy_path}/shared/$dir");
    }
    
    $finder = new \Symfony\Component\Finder\Finder();
    $iterator = $finder
        ->files()
        ->name('*.tpl')
        ->in(__DIR__ . '/shared');

    $tmpDir = sys_get_temp_dir();
    
    /*@var $file \Symfony\Component\Finder\SplFileInfo */
    foreach ($iterator as $file) {
        $success = false;
        // Make tmp file
        $tmpFile = tempnam($tmpDir, 'tmp');
        if (!empty($tmpFile)) {
            try {
                $contents = $compiler($file->getContents());
                $target   = preg_replace('/\.tpl$/', '', $file->getRelativePathname());
                // Put contents and upload tmp file to server
                if (file_put_contents($tmpFile, $contents) > 0) {
                    upload($tmpFile, 'shared/' . $target);
                    $success = true;
                }
            } catch (\Exception $e) {
                $success = false;
            }
[...]

PR #220, opened 2015-03-30
[task deploy:shared] Copy shared files/directory

dirkluijk:
See https://github.com/deployphp/deployer/issues/152

Summary:
* Copies shared directories from source recursively, without overwriting existing files
* Copies shared files from source, if the file doesn't exists an empty one is created

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -90,7 +90,7 @@ task('deploy:rollback', function () {
 
 
 /**
- * Create symlinks for shared directories and files
+ * Create symlinks for shared directories and files and copy existing contents
  */
 task('deploy:shared', function () {
     $basePath = config()->getPath();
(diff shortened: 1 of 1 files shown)


oanhnn:
@dirkluijk , I think you should implement this feature as optional, copy or not.

gordalina:
I agree with @oanhnn, there should be an option to enable this. Current method should be the default.

dirkluijk:
Hm, flexibility is great, but I couldn't imagine a situation where you want empty files/directories if the files and directories already exist in source control? Because that would still be possible of they don't exist.

Of course I could make it optional, but I can't figure out how. It seems not possible to add additional options to a ``TaskCommand`` generated by ``task``. And separate it from the current task is also not an option, because this currently removes to existing files/directories before creating the symlinks.

oanhnn:
Dear @dirkluijk ,

Source code (checkout from git) is deployed on stages. So that, configure files (`shared_files`, `shared_dirs`) are modified when deploy to a special stage.
When re-deploy on stage, if copy original configure files, you have modify it again. Are you want this?
I think we should generate configure files in first deploy and re-generate it when stage have change. 
And we need a task to generate configure files instead of copy it from source code.

You can implement this feature as option, using env vars. Eg `env('copy_original_shared')`. 

```php
env('copy_original_shared', false);

task('deploy:shared', function(){
    // ...
    if (env('copy_original_shared')) {
        // copy original shared files and dirs
    }
    // ...
});
```

dirkluijk:
@oanhnn: the files are only copied the first time, the second time the files will not be overwritten thanks to the ``-n`` flag. I think that's the desired behaviour.

elfet:
Can you make these changes to `master` branch? Not more features will be implemented in 2.x branch.

PR #224, opened 2015-04-08
Atomic symlink swap

trauma2u:
Creating symlink to release, the symlink swap is now a **non-atomic** operation.
To use "ln -fsn" is not bad, good enough. But '-n' option is non-standard and so is not recommended. Above all, it's not sure "ln -f" is atomic.
It's sure "mv" is an **atomic** operation.
**"mv -Tf"** - with '-T' option - may be the best bet.

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -208,8 +208,7 @@ task('deploy:vendors', function () {
  */
 task('deploy:symlink', function () {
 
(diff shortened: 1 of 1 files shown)


elfet:
Cool. Will try it and then merge.

trauma2u:
Thanks a lot.
I'm looking forward to 3.0 :)

elfet:
Merged.

PR #226, opened 2015-04-17
Add namespace on functions.php

luciano-jr:
For Laravel 5 compatibility, because the function info() is previously declared in laravel/framework/src/Illuminate/Foundation/helpers.php


luciano-jr:
Why the tests has a undefined functions?

luciano-jr:
So sorry, I make a misstake

PR #242, opened 2015-05-08
Run commands as specific user [feature] [idea] [discuss]

m038:
This PR changes the run functionality a bit and allows for commands to be executed as a specific user. 
The runAs() function, which is basically the old run() function with a few changes, allows a second parameter which is the username. It accepts the _special_ username _sudo_, it will be prepend as is to the command. If a _false_ value (null, '', false, etc.) is given, nothing will be prepended. Any other string values will be treated as usernames. A check will be performed to see if the username exists on the server (using the _getent passwd \<username\>_ command). If not a RunTimeException is thrown. For a valid username, all commands will be prepended with _sudo -u \<username\>_.
The run() function is wrapper for the new runAs() function. It will use the environment variable _run_as_ as a username or null if that is not set.
The runAsCurrent() function is a wrapper for runAs() and will force the command to be executed as the current user.

Via the environment variable _run_as_ it's possible to change the user for all run commands.

Please share your ideas about this here.

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -201,18 +201,53 @@ function workingPath()
 }
 
 /**
- * Run command on server.
+ * Run command on server as default user. This function respects the run_as env
+ * variable.
  *
  * @param string $command
  * @return Result
  */
 function run($command)
 {
+    return runAs($command, env()->get('run_as', null));
+}
+
+/**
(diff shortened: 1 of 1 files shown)


m038:
#245 is a replacement with proper branch names.

PR #245, opened 2015-05-11
Run commands as specific user

m038:
This PR changes the run functionality a bit and allows for commands to be executed as a specific user. 
The runAs() function, which is basically the old run() function with a few changes, allows a second parameter which is the username. It accepts the special username sudo, it will be prepend as is to the command. If a false value (null, '', false, etc.) is given, nothing will be prepended. Any other string values will be treated as usernames. A check will be performed to see if the username exists on the server (using the getent passwd <username> command). If not a RunTimeException is thrown. For a valid username, all commands will be prepended with sudo -u <username>.
The run() function is wrapper for the new runAs() function. It will use the environment variable run_as as a username or null if that is not set.
The runAsCurrent() function is a wrapper for runAs() and will force the command to be executed as the current user.

Via the environment variable run_as it's possible to change the user for all run commands.

Please share your ideas about this here.

Replaces #242.

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -201,18 +201,53 @@ function workingPath()
 }
 
 /**
- * Run command on server.
+ * Run command on server as default user. This function respects the run_as env
+ * variable.
  *
  * @param string $command
  * @return Result
  */
 function run($command)
 {
+    return runAs($command, env()->get('run_as', null));
+}
+
+/**
+ * Run command on server as ssh user. This ignores the run_as env variable.
+ *
+ * @param string $command
+ * @return Result
+ */
+function runAsCurrent($command)
+{
+    return runAs($command, null);
+}
+
+/**
+ * Run command on server.
+ *
+ * @param string $command
+ * @param string $username
+ * @return Result
+ */
+function runAs($command, $username)
+{
     $server = Context::get()->getServer();
     $command = env()->parse($command);
     $workingPath = workingPath();
+    $runAs = '';
 
-    $command = "cd $workingPath && $command";
+    if ($username == 'sudo') {
+        $runAs = 'sudo';
+    } elseif ($username) {
+        $userExists = runAsCurrent("getent passwd $username");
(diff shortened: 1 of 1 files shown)


m038:
@oanhnn i agree it doesn't force this security restriction. However it's also not always possible to ssh as the user under which your webserver software runs. That could lead to permissions troubles. E.g. when wanting to delete older releases, etc.
Next to that in your current common recipe you also allow users to run the writables command as sudo and the directories are chmodded to 777. That gave me the idea that you would allow your software to be more flexible, then to strictly enforce secure methods. 
So this is just giving the user more options and it's still their responsibility to handle security in a good manner.

elfet:
I think bringing bad practice to lib it unacceptable. You need to configure your server right, instead of using `sudo` on all of commands.

src/functions.php
```diff
@@ -201,18 +201,53 @@ function workingPath()
...
+ * @return Result
+ */
+function runAs($command, $username)
+{
     $server = Context::get()->getServer();
     $command = env()->parse($command);
     $workingPath = workingPath();
+    $runAs = '';
+
+    if ($username == 'sudo') {
```
elfet: Magic username is bad solution.

elfet:
Well, i do not see use case of this feature.

threesquared:
On some servers I manage there is one user that is used to log in. This user is then able to run commands with `sudo -u username` to deploy various websites. This allows a single ssh user but simultaneously allows websites to be owned by multiple other users (for mpm.itk). So it would be very useful to be able to globally prefix commands with `sudo -u username` for me :)

diegocastro:
@threesquared I have the same setup on my servers. Every website have a user, sudo permission to them is a no-no.

It would be very useful to run a command or task as another user. Like [ansible become](http://docs.ansible.com/ansible/become.html).

elfet:
May be we need some more testable and safe solution. Maybe connect to server as another user to run task?

PR #247, opened 2015-05-11
Fix restore working path after task deploy:writable

oanhnn:
After task `deploy:writable`, we should restore working path.


elfet:
Is it a bug? Can you add some tests?

Issue #248, opened 2015-05-11
Shell commands should probably include bang


pbudzon:
Hi!

The deployer 2.X doesn't work when trying to deploy on server that does not use sh/bash as a default shell (like many *BSD machines). Because of that, commands like this:
```
run("if [ ! -d $(echo $basePath) ]; then mkdir $basePath; fi", true);
```
die with errors like
```
[RuntimeException]      
  Illegal variable name.  
````

This can be simply prevented by adding bang (`#!/bin/sh `) before the actual command.
For example:
```
run("#!/bin/sh \n if [ ! -d $(echo $basePath) ]; then mkdir $basePath; fi", true);
```

runs perfectly fine :)

elfet:
Ok. Cool one :+1: I thing need to add some parameter with pre_run name or similar.

pbudzon:
It doesn't work with version 3 though :( 

To prevent this from not working for people, I'd suggest doing something like `echo $SHELL` before running the commands to make sure the shell we are connected to is compatible with `sh`. For example `/bin/tcsh` is not :) 
This while issue can be simply solved by the user by setting a default shell for the user to something compatible (sh, bash, etc). But giving people a warning would be nice.

elfet:
Can you create PR with shell cheking and warning? This will be so great! :green_apple:

pbudzon:
Will do!

elfet:
Thanks you a lot! :)

Max-Might:
I just noticed that the exception is thrown when `{{deploy_path}}` does not exist on the server.
This will confuse people. Maybe we should check with `/tmp` or `/home` or something else.

elfet:
Done: https://github.com/deployphp/deployer/blob/master/recipe/common.php#L61

PR #249, opened 2015-05-11
Added simple check if the shell on remote server is POSIX-compliant.

pbudzon:
As described in issue #248 - the commands for deployment can fail with mysterious errors when remote server is running a non-POSIX shell (like tcsh, common on BSD systems). 
Added a simple check with expression - if it fails, user will see more friendly message explaining to switch to sh/bash/similar.


elfet:
Thanks a lot!

PR #262, opened 2015-05-19
Changed the symfony recipe to support the v3 dir structure

ZeeCoder:
In deployer v2 I had to copy the whole recipe to support the new symfony dir structure.
Migrating to v3, it feels very unprofessional, so I made some changes in it,
which makes it very easy to support projects adopting the new dir structure early on.

Usage example from my `deployer.php` file:

```php
require 'recipe/symfony.php';

// Switching to the sf3 dir structure
set('shared_dirs', ['var/logs']);
set('writable_dirs', ['var/cache', 'var/logs']);
set('bin_dir', 'bin');
set('var_dir', 'var');
```

Awaiting you opinions. :)

### DIFF
--- a/recipe/symfony.php
+++ b/recipe/symfony.php
@@ -30,13 +30,17 @@ set('auto_migrate', false);
 env('env_vars', 'SYMFONY_ENV=prod');
 env('env', 'prod');
 
+// Adding support for the Symfony3 directory structure
+set('bin_dir', 'app');
+set('var_dir', 'app');
+
 
 /**
  * Create cache dir
  */
 task('deploy:create_cache_dir', function () {
     // Set cache dir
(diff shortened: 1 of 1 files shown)


elfet:
Nice one, thanks! :+1: I updateв docs too.
Need to think for a some syntax for parameters, like this maybe:

``` php
run('php {{release_path}}/{{%bin_dir}}/console assetic:dump --env={{env}} --no-debug');
```

ZeeCoder:
A bigger issue though is that you didn't update the downloadable phar.
(Tried it ~5-10 minutes ago.)
So now the documentation is out of sync with the executable. :/

elfet:
Phar will automatically will be updated in next 10 mins.

ZeeCoder:
Didn't know that, it's my mistake then. :)

PR #267, opened 2015-05-19
Added some modifications into the global functions, and fixed a spelling mistake

ZeeCoder:
- Fixed the success sentence, which was outputted when a new phar build was finished,
 - Added automatic environment variable registration to the `server` function ("host" and "name" variables),
 - Added a `has` function, so that the existence of a parameter can be checked.

I'm awaiting your opinions.

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -425,6 +425,15 @@ function get($key)
 }
 
 /**
+ * @param string $key
+ * @return boolean
+ */
+function has($key)
+{
+    return Deployer::get()->parameters->has($key);
+}
+
+/**
  * @param string $message
  * @param string|null $default
  * @return string
--- a/src/functions.php
+++ b/src/functions.php
@@ -51,10 +51,6 @@ function server($name, $host = null, $port = 22)
     $deployer->servers->set($name, $server);
     $deployer->environments->set($name, $env);
 
-    // Setting environment variables about the server automatically
-    $env->set('name', $name);
-    $env->set('host', $host);
-
     return new Builder($config, $env);
 }
 
--- a/src/functions.php
+++ b/src/functions.php
@@ -35,12 +35,12 @@ use Symfony\Component\Console\Output\OutputInterface;
  * @param int $port
  * @return Builder
  */
-function server($name, $domain = null, $port = 22)
+function server($name, $host = null, $port = 22)
 {
     $deployer = Deployer::get();
 
     $env = new Environment();
-    $config = new Configuration($name, $domain, $port);
+    $config = new Configuration($name, $host, $port);
 
     if ($deployer->parameters->has('ssh_type') && $deployer->parameters->get('ssh_type') === 'ext-ssh2') {
         $server = new Remote\SshExtension($config);
@@ -50,6 +50,10 @@ function server($name, $domain = null, $port = 22)
 
     $deployer->servers->set($name, $server);
     $deployer->environments->set($name, $env);
+
+    // Setting environment variables about the server automatically
+    $env->set('name', $server);
+    $env->set('host', $host);
 
     return new Builder($config, $env);
 }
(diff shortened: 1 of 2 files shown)


ZeeCoder:
Forgot to mention, if this gets merged in, then the documentation should definitely mention the automatically generated env variables. (host/name)
Same with the `has` function, it must be listed in the "Functions" section of the docs.

src/functions.php
```diff
@@ -51,6 +51,10 @@ function server($name, $domain = null, $port = 22)
     $deployer->servers->set($name, $server);
     $deployer->environments->set($name, $env);
 
+    // Setting environment variables about the server automatically
+    $env->set('name', $server);
+    $env->set('host', $host);
```
elfet: I think it's better to put this lines inside `Builder` class, as there are `localServer` function needs this lines too.

Also can you add some test to check after function `server`/`localServer` call env vars a created?

elfet:
And check new issues here: https://scrutinizer-ci.com/g/deployphp/deployer/inspections/fe07aa87-2431-4754-ab40-813f64962e04

src/functions.php
```diff
@@ -51,6 +51,10 @@ function server($name, $domain = null, $port = 22)
     $deployer->servers->set($name, $server);
     $deployer->environments->set($name, $env);
 
+    // Setting environment variables about the server automatically
+    $env->set('name', $server);
```
elfet: `$server` isn't string, it's `ServerInterface`. I think you mention: 
``` php
$env->set('name', $name);
```

elfet:
BTW, its better to separate PR into small, independent ones. So it's simple to merge them.

ZeeCoder:
I'll keep that in mind next time. :)
I'll check these issues as soon as I can.

ZeeCoder:
I refactored some things and played around the "host" and "name" environment settings.
I discarded the latter idea because the first - naive - implementation would certainly cause issues in the future.

I'll open a separate issue about the env variables.

oanhnn:
After 7 commits, this PR has:
- Spelling fix. Fixed the success sentence, which was outputted when a new phar build was finished,
- Added a `has` function, so that the existence of a parameter can be checked.
I think clear more if you re-base before merging code.

elfet:
Do we really need `has` function?

ZeeCoder:
I can break this PR into two separate one if you think, one for the spelling and an other for the has function.

I added that global function, since I wanted to check inside a task, whether a global configuration array was set or not. I wanted to implement proper error handling, which is as informative as it can be.
The current `get` implementation throws an exception if the given key does not exist in the collection, which is fine, but I wanted a different error message.
I can of course put the `get` call in a try-catch block, and throw my own exception, but then there's the chance of catching exceptions not related to the existence of the global variable I was trying to access.

The task I was trying to implement was a slack notification task, and with the `has` function I could do the following:

```php
// Checking for the existence of the slack task configuration.
// If it doesn't exist, I throw a custom exception with detailed information
// (Actually it could be even more detailed, by giving the expected array keys,
// but this is only an example.)
if (!has('slack')) {
    throw new \RuntimeException('Missing "slack" parameter for the "slack:push:success" task. Please provide it using the "set(\'slack\', [])" function.');
}

// The slack global variable now surely exist. If this get call throws
// something, that will not be related to the existence of the variable.
$config = get('slack');

// Validating the given configuration options goes here 
// ...
```

In short: I wanted to implement proper error handling inside a given task, and I think a `has` function can make that very simple.

elfet:
Looks good. I'm merging.

ZeeCoder:
Thanks. :)
On a related note: would you consider adding a simple "slack" task?
Because I could share what I have.

elfet:
Take a look at: https://github.com/deployphp/recipes

ZeeCoder:
Now that slack implementation is a perfect example why I wanted to access the "host" server configuration.
Does it work in v3?
It uses the `config()` function which I think does not exist anymore:
https://github.com/deployphp/recipes/blob/master/recipes/slack.php

elfet:
Yeahh, need to create issue there.

ZeeCoder:
On it. :)

Issue #270, opened 2015-05-20
There should be a way to be able to access the server configurations (host, name, etc.)


ZeeCoder:
I would need information about the server I'm currently deploying to, so that I can output proper messages if I need to.

I do understand why it's not ideal to give full access to the Configuration object through a global `config()` function, because then any task could - in theory - modify the server configuration values.

So in short, I think there should be a way to "get" the configuration values, but it should not expose "setters".

Here's an example of a 3rd party recipe to show an actual use-case for the old `config()` function:
https://github.com/deployphp/recipes/blob/master/recipes/slack.php

elfet:
I think better place for this data is env. 
`env('host')`

ZeeCoder:
Yeah, the problem was with that approach is that if we follow that naming pattern, then it can easily conflict with other env variable names.

Like: `env('name')` -> this doesn't tell you much about what's name is it, and it also conflicts with some tests, where the `env('name', 'value')`.

I thought about prefixing them, like:
`env('server-name', '...')`,
but it starts to feel like there's just too much magic involved, that would happen under the hood.

What do you think?

elfet:
It’s good approach. Env already support this. 
You can do this:
```php
env(‘config’, [
    ’name’ => ‘prod’,
    ...
]);
```

Now possible:

```php
echo env(‘config.name’); // prod
```

ZeeCoder:
Oh, I see, so I could implement the following code in the Builder constructor:

```php
// ...
$env->set('server', [
    'name' => $config->getName(),
    'host' => $config->getHost(),
    // ...
])
// ...
```

So that later in a task, I could access them as follows:
```php
$serverName =  env('server.name');
$serverHost =  env('server.host');
```

I would like that approach.
One question though: can I get the full array just by saying: `env('server')`?

elfet:
Honestly, i don't know :sunglasses: I don't implemented this feature.

ZeeCoder:
Okay, I'll check out it later then.
But about the Builder class: does that look okay to you?

Because if it does, I'll be sure to implement that feature, and if that's done,
I think I'll create another pull request, to fix the 3rd party slack recipe too. :)

elfet:
Yes, look good for me.

ZeeCoder:
I'm closing this issue in favor of the #272 pull request.

PR #272, opened 2015-05-20
Automatic registration of server env vars

ZeeCoder:
I implemented the solution we aggreed upon in issue #270.
The problem is, that it breaks an existing test, and I can't figure out why.

Can someone help me out with this?
(If this issue is resolved I'll add tests for the new functionality too.)

Here's the output of phpunit:

```bash
PHPUnit 4.6.6 by Sebastian Bergmann and contributors.

Configuration read from /home/zeecoder/work/deployer/phpunit.xml

............................................F.................... 65 / 74 ( 87%)
.........

Time: 2.13 seconds, Memory: 12.75Mb

There was 1 failure:

1) Deployer\Server\BuilderTest::testEnv
Expectation failed for method name is equal to <string:set> when invoked 1 time(s)
Parameter 0 for invocation Deployer\Server\Environment::set('server', Array (...)) does not match expected value.
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'name'
+'server'

/home/zeecoder/work/deployer/src/Server/Builder.php:36
/home/zeecoder/work/deployer/test/src/Server/BuilderTest.php:120
/usr/share/php/PHPUnit/TextUI/Command.php:186
/usr/share/php/PHPUnit/TextUI/Command.php:138

FAILURES!
Tests: 74, Assertions: 208, Failures: 1.
```

### DIFF
--- a/src/Server/Builder.php
+++ b/src/Server/Builder.php
@@ -28,11 +28,11 @@ class Builder
         $this->config = $config;
         $this->env = $env;
 
-        $env->set('server', [
+        $env->setAsProtected('server', [
             'name' => $config->getName(),
             'host' => $config->getHost(),
             'port' => $config->getPort(),
-        ], true);
+        ]);
     }
 
     /**
--- a/src/Server/Builder.php
+++ b/src/Server/Builder.php
@@ -27,6 +27,12 @@ class Builder
     {
         $this->config = $config;
         $this->env = $env;
+
+        $env->set('server', [
+            'name' => $config->getName(),
+            'host' => $config->getHost(),
+            'port' => $config->getPort(),
+        ]);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


elfet:
It's fail down because of (BuilderTest.php#L115)[https://github.com/deployphp/deployer/blob/master/test/src/Server/BuilderTest.php#L115]
In this test `Deployer\Server\Environment` expected to be called once. And you add second call in this PR.

ZeeCoder:
I think it should be okay now.

elfet:
Please, before i merge this PR, add docs to deployphp/docs too. Thank a lot!

elfet:
Now we need set as private here.

ZeeCoder:
I had to redo this due to incompatibilities with the resent changes.
Also it contained 5 commits for something really simple.

PR #274, opened 2015-05-20
Shared directories and files is not getting called in Laravel recipe

lasselehtinen:
Seems like the deploy:shared was forgotten from the Laravel recipe.


elfet:
Thanks!

PR #275, opened 2015-05-21
Realtime output in runLocally function

yannschepens:
…play debug output with this mode

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -314,21 +314,21 @@ function runLocally($command, $timeout = 60)
 
     $process = new Symfony\Component\Process\Process($command);
     $process->setTimeout($timeout);
-    $process->run();
+    $process->run(function($type, $buffer){
+        if (isDebug()) {
+            if ('err' === $type) {
+                write("<fg=red>></fg=red> $buffer");
+            } else {
+                write("<fg=green>></fg=green> $buffer");
+            }
+        }
+    });
 
     if (!$process->isSuccessful()) {
         throw new \RuntimeException($process->getErrorOutput());
     }
-
-    $output = $process->getOutput();
-
-    if (isDebug() && !empty($output)) {
-        writeln(array_map(function ($line) {
-            return "<fg=red>></fg=red> $line";
-        }, explode("\n", $output)));
-    }
-
-    return new Result($output);
+    
+    return new Result($process->getOutput());
 }
 
 /**


src/functions.php
```diff
@@ -314,7 +314,15 @@ function runLocally($command, $timeout = 60)
 
     $process = new Symfony\Component\Process\Process($command);
     $process->setTimeout($timeout);
-    $process->run();
+    $process->run(function($type, $buffer){
+        if (isVeryVerbose()) {
```
elfet: May be it's better use `isDebug()` here?

src/functions.php
```diff
@@ -322,7 +330,7 @@ function runLocally($command, $timeout = 60)
 
     $output = $process->getOutput();
 
-    if (isDebug() && !empty($output)) {
+    if (!isVeryVerbose() && isDebug() && !empty($output)) {
```
elfet: And delete this duplicate output?

src/functions.php
yannschepens: As you wish, the twice are ok for me. I do the modification.

src/functions.php
yannschepens: Yes

yannschepens:
Updated

src/functions.php
```diff
@@ -322,7 +330,7 @@ function runLocally($command, $timeout = 60)
 
     $output = $process->getOutput();
 
-    if (isDebug() && !empty($output)) {
+    if (!isDebug() && !empty($output)) {
```
elfet: Now this will be printing output all the time. Lets delete this block completely.

src/functions.php
```diff
@@ -314,7 +314,15 @@ function runLocally($command, $timeout = 60)
 
     $process = new Symfony\Component\Process\Process($command);
     $process->setTimeout($timeout);
-    $process->run();
+    $process->run(function($type, $buffer){
+        if (isDebug()) {
+            if ('err' === $type) {
+                write('<fg=red>'.$buffer.'</fg=red>');
```
elfet: Maybe it's better to keep this format?
```php
"<fg=red>></fg=red> $line"
```

src/functions.php
yannschepens: Ok

src/functions.php
yannschepens: You're the master, I change it. I do the same for the next line.

PR #276, opened 2015-05-21
Added the possibility of 'protected' env parameters. (+test).

ZeeCoder:
We didn't talk about this, but I think this feature is a natural addition.

Basic env parameters - especially ones that could be mentioned in the docs - should be protected from future changes, since many 3rd party and custom tasks can rely on these values.

Consider the env parameters set by the Builder (#272).
If a 3rd party task later desides to set the "server" env var, there's nothing to stop him in doing so.

Now if a developer just include that task in his deployment process without digging in the code and checking every detail, it can cause side-effects, since after that task runs, any following task that relies on the "server" env var will fail in some way.

At this point, the developer can easily misjudge the situation and believe that the depoyer code itself is buggy, while in reality the 3rd party task is the one causing unwanted side-effects.

If this pull request gets accepted, I'll create another one, which sets the aforementioned "server" env parameter as protected.

### DIFF
--- a/src/Server/Environment.php
+++ b/src/Server/Environment.php
@@ -25,7 +25,14 @@ class Environment
      * @var \Deployer\Type\DotArray
      */
     private $values = null;
-    
+
+    /**
+     * Values represented by their keys here are protected, and cannot be
+     * changed by calling the `set` method.
+     * @var array
+     */
+    private $protectedValueKeys = [];
+
     /**
      * Constructor
      */
@@ -37,10 +44,19 @@ class Environment
     /**
      * @param string $name
      * @param bool|int|string|array $value
+     * @param bool $isProtected
      */
-    public function set($name, $value)
+    public function set($name, $value, $isProtected = false)
     {
+        if (in_array($name, $this->protectedValueKeys)) {
+            throw new \RuntimeException("The environment parameter `$name` has already been set, and is protected from changes.");
+        }
+
         $this->values[$name] = $value;
+
+        if ($isProtected === true) {
+            $this->protectedValueKeys[] = $name;
+        }
     }
 
     /**
--- a/src/Server/Environment.php
+++ b/src/Server/Environment.php
@@ -31,7 +31,7 @@ class Environment
      * changed by calling the `set` method.
      * @var array
      */
-    private $protectedValueKeys = [];
+    private $protectedNames = [];
 
     /**
      * Constructor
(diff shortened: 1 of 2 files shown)


oanhnn:
Were you think this case ?
```php
$env = new Environment();
$env->set('protected', ['a' => 1, 'b' => 2], true);
$env->set('protected.a', 'some-other-value');
```

ZeeCoder:
Good point, I didn't think of the dot notation.
I'll refactor the logic so it would handle such a situation.

elfet:
Good addition. Must have feature. I suggest туче syntax:
```php
$env->setProtected('key', ...);
```

But it has to be accessible by dot notation:
```php
$env->setProtected(''server', [
            'name' => $config->getName(),
            'host' => $config->getHost(),
            'port' => $config->getPort(),
        ]);

$env->set('server', ...); // Thrown an exception.

$env->set('server.name', ...); // Thrown an exception.
```

ZeeCoder:
I like this, one thing though: `setProtected` should be `setAsProtected`.
What are your thoughts on that?

elfet:
I think `setProtected($key, $value)` better. But `setAsProtected` is like *set as protected this key* `setAsProtected($key)`.

ZeeCoder:
Well if you really want, I can implement it as `setProtected` it just sounds a bit off to me.

Anyway, here's a longer explanation why I think "setProtected" would be slightly off.

`setProtected(...)` in a proper sentence would be: "Set this key protected." (or even better: "Set this key to be protected.")
This implies that there's an already existing key, which I now want to set to be protected.

What the method really should mean, is that I want to set a new key as a protected key.
So the `setAsProtected(...)` would then mean: "Set this key as protected." which properly implies, what I really meant.

Here's some example which further illustrates what I think the difference is between the two names:
```php
/**
 * This function sets an existing key to be protected.
 * @param string $key The key which needs to be protected
 */
function setProtected($key)
{
    // ...
}

/**
 * Sets a new environment parameter as protected.
 * @param string $key The parameter's key
 * @param string|bool|int|array $value The key's value
 */
function setAsProtected($key, $value)
{
    // ...
}
```

Originally when you proposed the `setProtected` method, and I still didn't read the code examples, I thought you meant this:

```php
$env->set('server', [
    'name' => $config->getName(),
    // ...
]);

$env->setProtected('server');
```

**Then again:** I can roll with `setProtected` if you really want. :+1:

elfet:
Well, you convinceв me. `setAsProtect` is better.

ZeeCoder:
Thanks! ^^
I'll get to it as soon as I have some free time then. :)

ZeeCoder:
I added the modifications and tests.

One thing though:
the logic of the new tests is to set something as protected with the `setAsProtected` method, then check if the `set` method would work after that.

Now this of course doesn't test what happens, if I set new parameters with the `setAsProtected` method, and then try to use it again instead of `set`.

Of course since internally the `setAsProtected` method calls the `set` method, it shouldn't be a problem, but I thought it worth mentioning anyway.

ZeeCoder:
Note: if this gets accepted, I'll write the documentation for this and #272 the next week.

elfet:
Awesome! Nice work! Merged! :beer:  :beers:

elfet:
Will wait for docs...

oanhnn:
@ZeeCoder , @elfet 
Did you test with case:
```php
    public function testContainingProtectedParam()
    {
        $env = new Environment();
        $env->set('not_protected', []);
        $env->setAsProtected('not_protected.protected', 'value');

        // throw exception => ok
        $env->setAsProtected('not_protected', 'value');

        // throw exception => error ???
        $env->setAsProtected('not_protect', 'value');
    }
```
Cause is https://github.com/deployphp/deployer/blob/master/src/Server/Environment.php#L87
Please fix this problem !!!!

ZeeCoder:
hmm, I can't fix it until monday, I only have my mobile on the weekend. :/

oanhnn:
@ZeeCoder ,
I fixed in PR #288

elfet:
Thanks!

elfet:
I have a new idea. That is all env vars "private" by default. And if you want you can override it with `$env->override(key, value)`? 
@ZeeCoder @oanhnn @gordalina what do you think?

oanhnn:
@elfet Are all variables can be override ? What was so different `set` function?

ZeeCoder:
I thought about that too, because it's just so easy to implement it.
But I think protected env vars are protected for a reason.
Implementing an override method would let the users break things, exactly what we want tot avoid.
I think everything can be solved easily without overriding protected env vars.

Also: I think this discussion can be closed, the only thing needs to be done is adding this to the documentation. (Which I will do.)

PR #288, opened 2015-05-23
Fix bug of pull request #276: add feature protected env params

oanhnn:
This pull request is fixing bug of pull request #276 .
> Note: This code using 
```php
if($len < $length && '.' === $name[$len] && 0 === strpos($name, $protectedName)) {
```
> it's performance better than 
```php
if(0 === strpos($name, $protectedName . '.')) {
```


elfet:
Thanks a lot!

PR #289, opened 2015-05-23
update contributing

oanhnn:
This project has many pull requests. I think it would be better if contributor squash multiple intermediate commits (while developing) before create pull request. To be best, one pull request only has one commit.


elfet:
That's purpose for this? Why contributors has to be restricted?

ZeeCoder:
I do agree with the one PR/feature mentality.
I think squashing the commits should not be mandatory.
Whether it's is necessary should be decided based on the particular PR itself I think.

I'm open-minded about the issue though.

elfet:
@oanhnn if you have any arguments, lets us know. For now i'm closing this PR.

PR #290, opened 2015-05-24
Refactored PR based on #272 (automatic setting of the "server" env var)

ZeeCoder:


elfet:
Awesome! This very helpful indeed. :)

PR #282, opened 2015-05-22
Add progress indicator to the upload function, when uploading to a di…

gStanev:
Add progress indicator to the upload function, when uploading to a directory

PR #293, opened 2015-05-25
Use the Symfony progress bar when uploading directories.

Max-Might:
As discussed in pull request #282.

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -365,12 +365,12 @@ function upload($local, $remote)
 
         /** @var $file \Symfony\Component\Finder\SplFileInfo */
         foreach ($files as $file) {
-            $progress->advance();
-
             $server->upload(
                 $file->getRealPath(),
                 $remote . '/' . $file->getRelativePathname()
             );
+
+            $progress->advance();
         }
 
         $progress->finish();
--- a/src/functions.php
+++ b/src/functions.php
@@ -361,6 +361,7 @@ function upload($local, $remote)
             ->in($local);
 
         $progress = new ProgressBar(output(), count($files));
+        $progress->start();
 
         /** @var $file \Symfony\Component\Finder\SplFileInfo */
         foreach ($files as $file) {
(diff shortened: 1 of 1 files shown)


elfet:
Test fails, can you check and fix it before i merge your PR?

Max-Might:
The problem seems to be in symfony/console.
I have implemented the progress bar as per the docs:
http://symfony.com/doc/current/components/console/helpers/progressbar.html


The error is:
```
There was 1 error:

1) Deployer\FunctionsTest::testUpload
Argument 1 passed to Symfony\Component\Console\Helper\Helper::strlenWithoutDecoration() must implement interface Symfony\Component\Console\Formatter\OutputFormatterInterface, null given, called in /home/koce/Projects/deployer/vendor/symfony/console/Symfony/Component/Console/Helper/ProgressBar.php on line 545 and defined
```

ZeeCoder:
Is there any progress here @Max-Might?
Did you try to fix the issue?

Max-Might:
Yes but I can't find where the problem is :(
I have implemented the progress bar as per the docs.
I get the output stream from deployer and pass it to the progress bar.
The error is in the symfony code.

ZeeCoder:
It's not the Symfony Component that's at fault here, look at the error more closely: https://travis-ci.org/deployphp/deployer/jobs/64213311#L264
It has something to do the way the tests work.

For some reason during that test the output you're passing to the Progress Bar by calling `output()` giving you `null`  instead of a proper `OutputFormatterInterface` object when the component calls the `getFormatter` method on it : https://github.com/symfony/Console/blob/2.6/Helper/ProgressBar.php#L545

I hope this helps. :)

Max-Might:
Yes, this error happens only when we run the tests.
Is this how we create the output stream for the test procedure?

https://github.com/deployphp/deployer/blob/master/test/recipe/Helper/RecipeTester.php#L50

ZeeCoder:
Seems so.
The `getFormatter` method needs stubbing.

ZhukV:
So, this issue have a next problems:

1. The ProgressBar should be configurable.

As example:
```php
upload('/tmp/temporary.files', '/var/www/domain.com/tmp/temporary.files', true);
```
The third parameter indicated for use ProgressBar.

2. How this work in parallel?

elfet:
What about parallel?

GrahamCampbell:
Looks like this needs rebasing.

elfet:
Please, rebase.

PR #304, opened 2015-05-28
Add support file .env in Laravel 5. Add test class for laravel recipe

oanhnn:
In this pull request:
- Add `.env` into `shared_files`, to support Laravel 5.
- Add test class for laravel recipe

@elfet , you can don't merge this pull request and implement it yourself, because this test will take a long time. Sorry!


test/recipe/LaravelTest.php
```diff
@@ -0,0 +1,40 @@
...
+class LaravelTest extends RecipeTester
+{
+    protected function loadRecipe()
+    {
+        require __DIR__ . '/../../recipe/laravel.php';
+    }
+    
+    public function testDeploy()
+    {
+        set('repository', 'git@github.com:laravel/laravel.git');
```
elfet: This is huge package. Please take a loot on how `CommonTest` implemented.

ZeeCoder:
So what's up with this PR?

oanhnn:
Laravel 5 using `.env` file to save environment variables for applicaion configures, like `app/config/parameters.yml` in Symfony 2.
In this pull request:
* Add .env into shared_files, to support Laravel 5.
* Add test class for laravel recipe (but test class is not good and i was remove it)

ZeeCoder:
Is it for specifically Laravel5?
Because if it is, then deploying a Laravel4 app after this update will create a useless ".env" shared file doesn't it?

As far as I understand the current "laravel" recipe is for laravel4.
If this is all it takes to support Laravel5 deployment, then it should be in the docs instead, like with the symfony recipe and the symfony 3 directory structure.

What are your thoughts on that?

elfet:
So do we need to merge it?

ZeeCoder:
It depends.
 - If it doesn't hurt laravel 4 apps -> merge,
 - If it does, but this is all that's needed for laravel 5 support -> write it in the docs instead of changing the recipe.
 - If it takes more than that to support laravel 5 apps -> we should consider creating a separate laravel5 recipe.

imho

elfet:
@ZeeCoder i think you can merge it :+1:

ZeeCoder:
I'll wait for @oanhnn's answer just to be sure, no need to rush. :)

oanhnn:
Dear @ZeeCoder 
1. Laravel 5 is current stable version of laravel framework. Deployer is better if it support both laravel 5 and laravel 4. Laravel 5 should be default.
2. With laravel 4 case, file `.env` is useless, but it isn't affect with your app.
3. With this hash, user don't need custom recipe, Both laravel 4 and 5 are supported.

ZeeCoder:
Okay, I'm convinced dear @oanhnn. :)

Issue #291, opened 2015-05-24
Deployment success icon seems off on Ubuntu


ZeeCoder:
![screenshot from 2015-05-25 01 18 15](https://cloud.githubusercontent.com/assets/3679506/7790132/f9bef79a-027b-11e5-8532-c7a462e0a49e.png)

Is this intentional?

I think that the icons should appear before the tasks description after a task finishes, and not in the place of the letter "x" in "Executing".

elfet:
Yes, it's a bug. On Mac OS x everything looks normal. Do you know why this happens on Ubuntu?

ZeeCoder:
I have no idea, sorry. :/

ZeeCoder:
I can confirm it works perfectly on Mac:

![screen shot 2015-05-28 at 13 54 54](https://cloud.githubusercontent.com/assets/3679506/7859442/408b25a0-0541-11e5-9663-e47c72c4c325.png)

elfet:
Really strange behavior.

Lynnworld:
Hi, I just debug about this problem , it is about ascii cursor control char. based wiki:https://en.wikipedia.org/wiki/ANSI_escape_code  I found 2 way to fix this on ubuntu(no mac for test).
'✔' will have 2 chars position,  so I have add "\033[3D"  (means move cursor left 3 char). or  just add "\r" to go back to  line start position.

```php
Deployer\Console\Output\OutputWatcher line 42
$this->output->write("\033[k\033[1A\033[3D➤\n", false, $type); 
or
$this->output->write("\033[k\033[1A\r➤\n", false, $type);

Deployer\Executor\Informer line 52
$this->output->write("\033[k\033[1A\033[3D<info>✔</info>\n");
or
$this->output->write("\033[k\033[1A\r<info>✔</info>\n");
```
any one can help to test on mac ?

elfet:
Wow, cool! Can you create PR so i can test it?

elfet:
Fixed.

ZeeCoder:
excellent :)

PR #312, opened 2015-05-31
fix issue #291 by add '\r' to output

Lynnworld:
About #291 , I tried to use '\r' to move cursor back to line head,and fix it on ubuntu. however it need test for mac.


elfet:
Works good!

GrahamCampbell:
What about windows?

elfet:
Can you test it there?

PR #327, opened 2015-06-10
Make sure cache dir is writable after cache warmup

ihsanudin:
Make sure cache dir is writable after cache warmup. This helpul if you use not web server user to execute deploy

### DIFF
--- a/recipe/symfony.php
+++ b/recipe/symfony.php
@@ -120,6 +120,7 @@ task('deploy', [
     'deploy:vendors',
     'deploy:assetic:dump',
     'deploy:cache:warmup',
+    'deploy:writable',
     'deploy:symlink',
     'cleanup',
 ])->desc('Deploy your project');


elfet:
I don't think this is good solution to have two same tasks:
```php
task('deploy', [
     'deploy:prepare',
     'deploy:release',
     'deploy:update_code',
     'deploy:create_cache_dir',
     'deploy:shared',
     'deploy:writable',  // HERE
     'deploy:assets',
     'deploy:vendors',
     'deploy:assetic:dump',
     'deploy:cache:warmup',
    'deploy:writable', // AND HERE
     'deploy:symlink',
     'cleanup',
 ])->desc('Deploy your project');
```
If you want you can do you own check with `after`:
```php
after('deploy:cache:warmup', 'check_wtitable_dirs');
```

ihsanudin:
in my deploy.php file, I did use the after function to handle this task but i think it would be better if it can be handled by default

elfet:
Yes, you are right, thats is done by `deploy:writable` task before this tasks. So we don't need to run it twice.

ihsanudin:
No, you need to do it. Because before and after cache warming, you need to make sure the directory is writable

ZeeCoder:
I think I get where this could be a problem.

Please correct me if you thought of something else @ihsanudin.

Imagine the following: you have the web-server user `www-data` and you use another: `deployer` for deployment.
Now imagine that for some reason only a basic `chmod` is usable on the server.
(So neither `chmod +a` nor `setfacl` is available.)

This means that for the first time, the following line will run: `run("$sudo chmod 777 $dirs");`
Then warmup happens, which creates files with the `deployer` user's permissions, which means that the `www-data` webserver-user won't have access to these files, unless `deploy:writable` runs once again.

Do I understand the situation correctly?

Although running the task a second time in this case wouldn't help, since `chmod` is not in recursive mode.

ihsanudin:
@ZeeCoder you are right,

i think the problem come from this function

```https://github.com/deployphp/deployer/blob/master/recipe/common.php#L175```

We need to change mode (chmod) recursively  and run it as both task (the top line of function)

**Comment by ZeeCoder** [comment]
Hmm, there was a discussion about this, which is closed now, have a look at this: https://github.com/deployphp/deployer/issues/323

@elfet is there any problems that could arise from using `chmod -R 777` instead of `chmod 777`?

This is the second issue regarding this exact problem, so I think we need to think about it.

**Comment by oanhnn** [comment]
We have a case:
```
dir_A
 |-- file_a1
 |-- flie_a2
 \-- file_a3
```
| directory/file | owner | permission |
|------------- | ----------- | ------------- |
| dirA | deploy_user | 777 |
| file_a1 | deploy_user | 755 |
| file_a2 | www_user | 755 |
| file_a3 | www_user | 700 |

when using command `chmod -R 777 dir_A` without `sudo`, you can have a problem.

**Comment by elfet** [comment]
Well i think we can change `chmod` to `chmod -R`. But note, it's better to use `setfacl`.

**Comment by ZeeCoder** [comment]
As @oanhnn pointed out, simple using `chmod -R` might still not be enough.
If we change that, and the developer doesn't allow sudo-ing, then silent errors could potentially occur. -> terrible developer experience.

And there's something else that still bugs me.
Suppose we change this, so that deployment from now on uses the `chmod -R` command.
Now suppose I have an application, where during deployment I:
 1) create a folder,
 2) generate some files in it, for which specific permissions are set up for whatever reason,
 3) make above mentioned folder "writable" by setting the `writable_dirs` global parameter.

Now every file's permissions would be overwritten silently.

I know it's an unlikely edge-case, but I'm still afraid of such side-effects.

**Comment by elfet** [comment]
This is really edge-case) If we try to solve all cases, we endup only solving them. 
Developers need to understand what they are doing. If some one does not have setfacl, runs command not from www user, and wants to create writable dirs without sudo, we has nothing to do this that.

**Comment by ZeeCoder** [comment]
Hmm I'm still opposed to this.
Why not just introduce another task which could be used in cases like this:

```php
<?php

// in common.php
set('writable_files', []);

task('deploy:writable:files', function () {
    $files = join(' ', get('writable_files'));
    if (empty($files)) {
        return;
    }

    run("$sudo chmod -R 777 $files");
});

// in the project's deployer.php
set('writable_files', [
    'cache/*',
    // ...
]);

after('deploy:cache:warmup', 'deploy:writable:files');
```

ihsanudin:
+1

ZeeCoder:
Opinions @oanhnn @elfet?

elfet:
Why not just add '-R' option to 'deploy:writable'?
 `chmod -R 777`?

ZeeCoder:
I'll summarise my reasoning:

 - The global parameter's name `writable_dirs` would be misleading, since it implies that only directories will be affected, while in reality it would affect everything and rewrite the permissions to exactly 777, which might not be what I want.
 - It could potentially cause side-effect we can't foresee. Developers can be using this task any number of ways, not just the way it was meant to be used. Examples could be: https://github.com/deployphp/deployer/pull/327#issuecomment-110641573 and https://github.com/deployphp/deployer/pull/327#issuecomment-110647200. And those are just the ones we came up with spontaneously.
 - Providing a separate task for this would be a more reusable solution imho, and it could be added easily to the docs too. It would be similar to the `database:migrate` solution, which recently got documented.

Admittedly, just writing in `-R` would a be much easier change, but the more I think about it, the less I like that solution.

Anyway, if you think my reasons are not good enough, we can just roll with the `-R` switch. :)

elfet:
Where are do you think to add this new task?

ZeeCoder:
common.php
In a way, this is an addition to the `deploy:writable` task, that's why I named it `deploy:writable:files`.

ihsanudin:
I think is better to use ``chmod 777 -R`` and add new task call ``deploy:writable:files`` as option.

It will be better instead add new task for this issue

ZeeCoder:
Well I rest my case, I summed up what I think already. :)
@elfet it's up to you what's it gonna be.

elfet:
Guys, i can understand the problem we are trying to solve? :smiley: If developers uses setfacl - everything ok (permission sets for both current user and http user). 
If user does not setfacl, when Deployer fallback to use `chmod 777`. May be just se `chown www-data` too?

elfet:
@ZeeCoder what about user who do not want to run this tasks? 
```php
run("$sudo chmod -R 777 $files");
```

ZeeCoder:
Running the `deploy:writable:files` would be totally optional, that's why it would be documented how to use it.
My example (https://github.com/deployphp/deployer/pull/327#issuecomment-110657598) is showing changes in two files: `common.php` and the dev's `deployer.php`.

Basically nothing would change in the current behaviour, but there would be a solution for cases like this, which would be highlighted in the documentation.

elfet:
@ZeeCoder well, your solution seem to be good. Can you create PR?

ZeeCoder:
Absolutely :+1:

ihsanudin:
The problem is resolved so i will close this PR

PR #328, opened 2015-06-10
server() support timeout

TheRatG:
Good day. Thanks for you library. 
Added support timeout command execution. In my case, some command running more than 10 second and deploy is failing.

### DIFF
--- a/src/Server/Configuration.php
+++ b/src/Server/Configuration.php
@@ -111,8 +111,9 @@ class Configuration
      * @param string $name
      * @param string $host
      * @param int $port
+     * @param int $timeout
      */
-    public function __construct($name, $host, $port = 22, $timeout = 60)
+    public function __construct($name, $host, $port = 22, $timeout = 360)
     {
         $this->setName($name);
         $this->setHost($host);
--- a/src/Server/Configuration.php
+++ b/src/Server/Configuration.php
@@ -101,18 +101,42 @@ class Configuration
     private $pemFile;
 
     /**
+     * @var int
+     */
+    private $timeout;
+
+    /**
      * Construct
      *
      * @param string $name
      * @param string $host
-     * @param int    $port
+     * @param int $port
      */
-    public function __construct($name, $host, $port = 22)
+    public function __construct($name, $host, $port = 22, $timeout = 60)
     {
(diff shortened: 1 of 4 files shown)


src/Server/Configuration.php
```diff
@@ -101,18 +101,42 @@ class Configuration
...
+    /**
      * Construct
      *
      * @param string $name
      * @param string $host
-     * @param int    $port
+     * @param int $port
      */
-    public function __construct($name, $host, $port = 22)
+    public function __construct($name, $host, $port = 22, $timeout = 60)
```
elfet: Maybe increase default timeout to 360 sec?

src/functions.php
```diff
@@ -57,15 +58,16 @@ function server($name, $host = null, $port = 22)
 
 /**
  * @param string $name
+ * @param int $timeout
  * @return Builder
  */
-function localServer($name)
+function localServer($name, $timeout = 60)
```
elfet: I do not think `localServer` need this option. We do not connecting to localhost.

src/functions.php
```diff
@@ -33,14 +33,15 @@
  * @param string $name
  * @param string|null $host
  * @param int $port
+ * @param int $timeout
  * @return Builder
  */
-function server($name, $host = null, $port = 22)
+function server($name, $host = null, $port = 22, $timeout = 60)
```
elfet: May do not use past this params to constructor? And just use this setter?
```php
server(...)
    ->connectionTimeout(60);
```

elfet:
And also allow to specify `0` as unlimited timeout.

TheRatG:
Thanks, I will fix it.

src/functions.php
```diff
@@ -65,7 +67,7 @@ function localServer($name)
 
     $env = new Environment();
     $server = new Local();
-    $config = new Configuration($name, 'localhost'); // Builder requires server configuration.
+    $config = new Configuration($name, 'localhost', 22); // Builder requires server configuration.
```
elfet: `22`? Magic number?

elfet:
ping @TheRatG

oanhnn:
This pull request is closed by better pull request #415

PR #334, opened 2015-06-11
Change to recursive chmod

ihsanudin:
Change to recursive chmod to avoid inconsistency permission in nested dirs


elfet:
Ok, thanks!

PR #335, opened 2015-06-12
Add Writable Files and Add Chown on Writable Dirs Task

ihsanudin:
Add Writable Files and Add Chown on Writable Dirs Task to support vps based server that do not have acl

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -11,6 +11,7 @@
 set('keep_releases', 3);
 set('shared_dirs', []);
 set('shared_files', []);
+set('writable_files', []);
 set('writable_dirs', []);
 set('writable_use_sudo', true); // Using sudo in writable commands?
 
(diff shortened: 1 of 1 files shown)


recipe/common.php
```diff
@@ -168,6 +169,17 @@
...
+/**
+ * Make writable files.
+ */
+task('deploy:writable:files', function () {
+    $files = join(' ', get('writable_files'));
+    if (empty($files)) {
+        return;
+    }
+
+    run("$sudo chmod -R 777 $files");
```
elfet: `$sudo` does not defined.

elfet:
@ihsanudin do you tests your PR?

recipe/common.php
ihsanudin: Sorry @elfet : i just copy that task from recent discussions

ihsanudin:
@elfet : I edit the code directly from github editor so i don't test anything. Just make no error from travis

elfet:
@deployphp/maintainer that do you think about this PR, guys?

oanhnn:
I am using `deployer` with Slim framework and CakePHP framework.
When project run, log files and cache file is created by `www-data`, but onwer of directory `logs`, `tmp` or `cache` is deploy-user. 
I was comment in #327 and i don't agree with this PR.
I think `deployer` will support flow of framework and user can custom original recipe for special project or them.

elfet:
Maybe separate `deploy:writable:files` task to custom recipe (3rd party)? And just merge `chmod -R`/`chown` changes?

ihsanudin:
@oanhnn : This is a common task not a specific framework purpose. I just improve the behavior to dealing with vps that have not and can not set an acl

elfet:
I think this PR need to be separated. @ihsanudin can you do it?

PR #332, opened 2015-06-11
Fix problems with use many tasks with equals name in group #331

ZhukV:

Issue #331, opened 2015-06-11
Could not execute same task two times or more


piotrekkr:
Hi,

Today I tried execute same task 'deploy:permissions:fix' two times like:

    task('deploy', [
        'deploy:prepare',
        // ...
        'deploy:permissions:fix',
        // ...
        'database:migrate',
        'deploy:permissions:fix',
        //...
    ])->desc('Deploy your project');

But `deploy:permissions:fix` was run only once. Is this intentional? If no, could you change your code in ` \Deployer\Console\TaskCommand::execute()` to:

    $tasks = [];
    foreach ($this->deployer->scenarios->get($this->getName())->getTasks() as $taskName) {
        $tasks[] = $this->deployer->tasks->get($taskName);
    }

Thank you

elfet:
Hmmm, look like a bug, but i not sure =0 
@deployphp/maintainer guys, can anybody check this? :smile:

ZhukV:
I will correct.

elfet:
@ZhukV thanks!

ZhukV:
So, fixed. BUT!!!

The task (in runtime) have a unique identifier for control many task with one name in group and execution context. The identifier create before run task for each task (`uniqueid`). And in Output we can see this identifier.

As example:

```bash
$ ./bin/dep foo production
✔ Executing task some #5579874ddfd9e
✔ Executing task some_2 #5579874de003e
✔ Executing task some #5579874de022c
```
The `5579874ddfd9e`, `5579874de003e` and `5579874de022c` - is a identifiers of task.

And, for control this identifiers in parallel executor, i adds storage `tasks_to_do_info` to `pure`. And, send the task info (id and name) (previously only send task name).

I tested on series executor and parallel executor. All ok.

elfet:
Fixed!

PR #336, opened 2015-06-12
Fix problems with use many tasks with equals name in group

elfet:
Based on PR #332
Bug from #331


elfet:
@deployphp/maintainer lets test this and discuss!

elfet:
Do we really need tasks ids?

ZhukV:
Yes, for correct control in informer system.

As example: In my deploy system i rewrite informer for run any logic after fail tasks. And, if i have a two task with name `some_foo`, i can control runs with this identifier.

elfet:
@ZhukV i think it's a little bit complicated. I'd fixed this myself without introduction tasks ids. 
Check this out:  9667b4c7

PR #338, opened 2015-06-12
[BUGFIX] : Use server date

lucasmichot:
There is an issue with the release date : it's based on the client date.
It should be based on the server date.

How to reproduce:
- set your script to keep 2 releases
- set your server on UTC
- deploy twice from CircleCI (UTC)
- deploy once from a Europe/Berlin machine (or any other timezone as long as you more than 1 hour offset with UTC)

Releases:
- 20150612111941 : CircleCI (1st deployment) at 11:19:41
- 20150612113811 : CircleCI (2nd deployment) at 11:38:11
- 20150612093922 : Deployment from my machine at 11:39:22

The symlink task will symlink the latest release (20150612093922).
The cleanup will then remove the oldest release (20150612093922 too).

Only outdated releases will remain on the server : 20150612111941 and 20150612113811.

Using the server date ensures this situation does not happen.

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -93,7 +93,7 @@ env('release_path', function () {
  * Release
  */
 task('deploy:release', function () {
-    $release = date('YmdHis');
+    $release = run('date +%Y%m%d%H%M%S');
 
     $releasePath = "{{deploy_path}}/releases/$release";
 
--- a/recipe/symfony.php
+++ b/recipe/symfony.php
@@ -58,7 +58,7 @@ task('deploy:assets', function () {
         return "{{release_path}}/$asset";
     }, get('assets')));
 
-    $time = date('Ymdhi.s');
+    $time = run('date +%Y%m%d%H%M.%S')
 
     run("find $assets -exec touch -t $time {} ';' &> /dev/null || true");
 })->desc('Normalize asset timestamps');


elfet:
Wow! Thanks you for PR!

oanhnn:
Dear @lucasmichot , @elfet 
I don't think this solution is best when deploy to multi server with different timezone and pa
I think we should set default timezone on top of `deploy.php` file.
```php
// file deploy.php
date_default_timezone_set('UTC');
```

johnny-bit:
@oanhnn is right. If basing release date on server timezone and deploying for example on multiple servers across the world You could end up with messy situation when trying to get "proper" release date. So IMO as @oanhnn said, best bet would be to set timezone to fixed UTC and document that we're using UTC for deploy date.

lucasmichot:
@oanhnn `date_default_timezone_set('UTC');` was my workaround before fixing this bug.
@johnny-bit: forcing `UTC` would be a good idea, and would be consistant - but do we want to force that @elfet ?

kenjis:
We don't have to force any timezone. It is okay users set timezone they want in `deploy.php`.

johnny-bit:
Well, that's OK actually. However after a760a9f release date is based on server time. I've noticed when developing my shared hosting workflow recipe, that this can cause minor issues: My system is CEST while my test server is UTC, so release dates on both differ by ~2h.

So... @elfet - how about reverting @lucasmichot and go the route @kenjis suggest, but document it better? Something like "Names of Release directories are based on date. We thus encourage to use UTC timezone if You're deplouing in multiple timezones by settfig timezone in Your `deploy.php` file" ?

piotrekkr:
@elfet I'm also voting for reverting a760a9f17bf0ad3b28b9c5dfe3a29982266b00bf , better docs and freedom to set timezone myself :+1: .

gordalina:
It's good to use UTC as default timezone, it's good that it is configurable. Although it should be added to the documentation.
Also, this might fail when deploying to multiple servers with different timezones, but does someone have this use case?

lucasmichot:
No @gordalina it wont fails with a common timezone

elfet:
Maybe separate this thing to local task?

PR #339, opened 2015-06-12
Fix version constraints

GrahamCampbell:


composer.json
```diff
@@ -22,10 +22,10 @@
         "elfet/pure": "~1.1",
         "symfony/finder": "~2.6",
         "symfony/console": "~2.6",
-        "symfony/yaml": "^2.6",
+        "symfony/yaml": "~2.6",
```
GrahamCampbell: NB, this behaves exactly the same. No idea why `^` was used?

elfet:
Thanks.

PR #342, opened 2015-06-16
Update StageStrategy.php

kevinkhill:
I found some small grammatical errors, so I thought I could help out with a patch. Fantastic tool by the way, very helpful.


elfet:
Thanks!

Issue #346, opened 2015-06-16
Faster cloning by borrowing objects from existing clones


elfet:
Using git:
* "git clone --reference=<over there>" learned the "--dissociate"
   option to go with it; it borrows objects from the reference object
   store while cloning only to reduce network traffic and then
   dissociates the resulting clone from the reference by performing
   local copies of borrowed objects.

johnny-bit:
Ooh!! I can do that! I did for my `local_common`! may I PR this? And can I use "git_cache" env for that?

elfet:
Yes, sure. But it's better to detect git version and if git support it - use it.

johnny-bit:
Hmmm... That *is* good idea... however there *are* some drawbacks to it... But OK, will add that to my `local_common` for check and then will pr `common` :)

johnny-bit:
PR #352 fixes this :)

johnny-bit:
So... #352 landed, then this issue can be closed, yes?

PR #352, opened 2015-06-16
Ref #346 - faster git cloning

johnny-bit:
This enables faster git cloning by using options `--reference` and `--dissociate`.

By default if git on server is newer or equal to 2.3, git cache is enabled. However if user sees any drawbacks to using git cache and would rather use shallow clones, he is allowed to do so by simply setting env variable `git_cache` to `false`.

This work is based on deployphp/recipes#12 for same behaviour on local machine.

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -20,17 +20,15 @@ set('writable_use_sudo', true); // Using sudo in writable commands?
 env('branch', ''); // Branch to deploy.
 env('env_vars', ''); // For Composer installation. Like SYMFONY_ENV=prod
 env('composer_options', 'install --no-dev --verbose --prefer-dist --optimize-autoloader --no-progress --no-interaction');
-env('git_cache', function(){ //whether to use git cache - fester cloning by borrowing objects from existing clones.
+env('git_cache', function(){ //whether to use git cache - faster cloning by borrowing objects from existing clones.
   $gitVersion = run('git version');
   $regs = [];
-  echo $gitVersion;
   if (preg_match('/((\d+\.?)+)/', $gitVersion, $regs)) {
     $version = $regs[1];
   } else {
     $version = "1.0.0";
   }
-  $compare = version_compare($version, '2.3', '>=');
-  return $compare;
+  return version_compare($version, '2.3', '>=');
 });
 
 /**
@@ -129,6 +127,8 @@ task('deploy:update_code', function () {
     $repository = get('repository');
     $branch = env('branch');
     $gitCache = env('git_cache');
+    $depth = $gitCache ? '' : '--depth 1';
+    
     if (input()->hasOption('tag')) {
         $tag = input()->getOption('tag');
     }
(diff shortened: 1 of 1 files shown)


recipe/common.php
```diff
@@ -20,6 +20,18 @@
 env('branch', ''); // Branch to deploy.
 env('env_vars', ''); // For Composer installation. Like SYMFONY_ENV=prod
 env('composer_options', 'install --no-dev --verbose --prefer-dist --optimize-autoloader --no-progress --no-interaction');
+env('git_cache', function(){ //whether to use git cache - fester cloning by borrowing objects from existing clones.
+  $gitVersion = run('git version');
+  $regs = [];
+  echo $gitVersion;
```
elfet: Maybe delete this? =)

recipe/common.php
```diff
@@ -20,6 +20,18 @@
 env('branch', ''); // Branch to deploy.
 env('env_vars', ''); // For Composer installation. Like SYMFONY_ENV=prod
 env('composer_options', 'install --no-dev --verbose --prefer-dist --optimize-autoloader --no-progress --no-interaction');
+env('git_cache', function(){ //whether to use git cache - fester cloning by borrowing objects from existing clones.
```
elfet: fester? =)

recipe/common.php
```diff
@@ -20,6 +20,18 @@
...
+  $gitVersion = run('git version');
+  $regs = [];
+  echo $gitVersion;
+  if (preg_match('/((\d+\.?)+)/', $gitVersion, $regs)) {
+    $version = $regs[1];
+  } else {
+    $version = "1.0.0";
+  }
+  $compare = version_compare($version, '2.3', '>=');
+  return $compare;
```
elfet: Why not `return version_compare($version, '2.3', '>=');`?

recipe/common.php
```diff
@@ -127,7 +140,18 @@
...
 
-    run("git clone $at --depth 1 --recursive -q $repository {{release_path}} 2>&1");
+    $releases = env('releases_list');
+    
+    if($gitCache && isset($releases[1])){
+      try {
+       run("git clone $at --recursive -q --reference {{deploy_path}}/releases/{$releases[1]} --dissociate $repository  {{release_path}} 2>&1");  
+      } catch (RuntimeException $exc) {
+        // If {{deploy_path}}/releases/{$releases[1]} has a failed git clone, is empty, shallow etc, git would throw error and give up. So we're forcing it to act without reference in this situation
+        run("git clone $at --recursive -q $repository {{release_path}} 2>&1");
```
elfet: Maybe its better to not duplicate code and run line 153?

recipe/common.php
```diff
@@ -127,7 +140,18 @@
...
+    
+    if($gitCache && isset($releases[1])){
+      try {
+       run("git clone $at --recursive -q --reference {{deploy_path}}/releases/{$releases[1]} --dissociate $repository  {{release_path}} 2>&1");  
+      } catch (RuntimeException $exc) {
+        // If {{deploy_path}}/releases/{$releases[1]} has a failed git clone, is empty, shallow etc, git would throw error and give up. So we're forcing it to act without reference in this situation
+        run("git clone $at --recursive -q $repository {{release_path}} 2>&1"); 
+      }
+    } else{
+      run("git clone $at --depth 1 --recursive -q $repository {{release_path}} 2>&1");
```
johnny-bit: It's not duplicate code: this line creates shallow clone. Shallow clone cannot be used as a reference point. Also clone with full history takes more time to do than shallow one.

So reasoning is this:
1. We have git supporting reference and dissociate, and we have possible reference
  - try clonig with reference
  - well, that didn't work out, do normal clone
2. Well we don't have git supporting/dont want to use it, so clone shallow.

point 2 fails under one condition: initial git cloning, without first reference, would be shallow clone. next cloning would fail to reference, so we get full clone and then subsequent cloning would be OK... I think I had this figured right at some point in local common, let me check.

PR #357, opened 2015-06-17
Allows to change timezone with UTC as a default

lucasmichot:
This PR aims to be a solution following various comments on https://github.com/deployphp/deployer/pull/338.

A revert of https://github.com/deployphp/deployer/commit/a760a9f17bf0ad3b28b9c5dfe3a29982266b00bf would reintroduce the bug previously explained.

The following PR provides a default timezone to UTC that can be overridden in the deployment script.
This provides thus a consistency between releases name on different servers, as the TZ on this server might all be different.

Also some documentation on http://deployer.org would be necessary.

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -84,10 +84,10 @@ task('deploy:prepare', function () {
         throw $e;
     }
 
-	// Set the deployment timezone
-	if (!date_default_timezone_set(env('timezone'))) {
-		date_default_timezone_set('UTC');
-	}
+    // Set the deployment timezone
+    if (!date_default_timezone_set(env('timezone'))) {
+        date_default_timezone_set('UTC');
+    }
 
     run('if [ ! -d {{deploy_path}} ]; then echo ""; fi');
 
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -17,6 +17,7 @@ set('writable_use_sudo', true); // Using sudo in writable commands?
 /**
  * Environment vars
  */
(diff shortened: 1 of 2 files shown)


gordalina:
:+1: But I think PR's like this should come with a PR for docs as well.

ZeeCoder:
Once again I agree @gordalina, there are so much changes it's hard to keep track.
Documentation should be as up-to-date as possible.

oanhnn:
I think using local time should be default and this PR should be merged. But we have to add docs for it.

lucasmichot:
> I think using local time should be default and this PR should be merged. But we have to add docs for it.

@oanhnn using the default timezone would lead to https://github.com/deployphp/deployer/pull/338#issue-87683073
It necessary has to be a fixed timezone that all deployment machines use

oanhnn:
@lucasmichot 
Yes. i understand it. I mean set default timezone on `deploy.php` file and using `date()` function from deployment PC (not deployment server).

johnny-bit:
@oanhnn - Isn't that exactly what this PR does? only default timezone is set via env, not via direct php function in deploy.php

however IMO timezone shouldn't be in env, rather it should be in settings - after all we want to have ONE timezone across all, not per-environmant timezone, right?

elfet:
I think this is solution for now. But we want to be this is common settings which now can not be implemented and will be implemented in Deployer v4.

elfet:
Merged.

PR #361, opened 2015-06-17
Add migrate task to Laravel recipe

lucasmichot:
Add missing migrate task to deployment

### DIFF
--- a/recipe/laravel.php
+++ b/recipe/laravel.php
@@ -17,6 +17,13 @@ set('shared_files', ['.env']);
 set('writable_dirs', ['storage', 'vendor']);
 
 /**
+ * Migrate database
+ */
+task('deploy:migrate', function () {
+    run("cd {{release_path}} && ./artisan migrate --force");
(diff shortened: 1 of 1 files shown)


recipe/laravel.php
```diff
@@ -17,6 +17,13 @@
 set('writable_dirs', ['storage', 'vendor']);
 
 /**
+ * Migrate database
+ */
+task('deploy:migrate', function () {
+    run("cd {{release_path}} && ./artisan migrate --force");
```
GrahamCampbell: :-1: for forcing

GrahamCampbell: Does laravel support running with "no interaction". If not, we need to do something about that. :)

lucasmichot: True

mbardelmeijer:
:+1:

recipe/laravel.php
elfet: Maybe not forcing?

PR #368, opened 2015-06-18
Avoid using FQN

lucasmichot:
As this is already the case where most FQN are imported, this PR skip the usage of FQN in the source

### DIFF
--- a/bin/dep
+++ b/bin/dep
@@ -6,6 +6,10 @@
  * file that was distributed with this source code.
  */
 
+use Deployer\Console\Application;
+use Symfony\Component\Console\Input\ArgvInput;
+use Symfony\Component\Console\Output\ConsoleOutput;
+
 $loaded = false;
 
 foreach ([__DIR__ . '/../../../autoload.php', __DIR__ . '/../vendor/autoload.php'] as $file) {
@@ -34,9 +38,9 @@ require_once 'src/functions.php';
 define('DEPLOYER_BIN', __FILE__);
 
 // Init Deployer
-$console = new \Deployer\Console\Application('Deployer', 'master');
-$input = new \Symfony\Component\Console\Input\ArgvInput();
-$output = new \Symfony\Component\Console\Output\ConsoleOutput();
+$console = new Application('Deployer', 'master');
+$input = new ArgvInput();
+$output = new ConsoleOutput();
 $deployer = new \Deployer\Deployer($console, $input, $output);
 
 // Require deploy.php script
--- a/bin/dep
+++ b/bin/dep
@@ -7,6 +7,7 @@
  */
 
 use Deployer\Console\Application;
+use Deployer\Deployer;
 use Symfony\Component\Console\Input\ArgvInput;
 use Symfony\Component\Console\Output\ConsoleOutput;
 
@@ -41,7 +42,7 @@ define('DEPLOYER_BIN', __FILE__);
 $console = new Application('Deployer', 'master');
 $input = new ArgvInput();
 $output = new ConsoleOutput();
-$deployer = new \Deployer\Deployer($console, $input, $output);
+$deployer = new Deployer($console, $input, $output);
 
 // Require deploy.php script
 $options = getopt('f::', ['file::']);
--- a/build
+++ b/build
@@ -1,5 +1,8 @@
 #!/usr/bin/env php
 <?php
+use Phar;
+use Symfony\Component\Finder\Finder;
+
 require __DIR__ . '/vendor/autoload.php';
 
 $opt = getopt('v::');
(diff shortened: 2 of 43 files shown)


gordalina:
:+1:

build
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env php
 <?php
+use Phar;
```
GrahamCampbell: missing newline

src/Deployer.php
```diff
@@ -134,14 +140,14 @@ public function getOutput()
     /**
      * @param string $name
      * @return mixed
-     * @throws \InvalidArgumentException
+     * @throws InvalidArgumentException
```
GrahamCampbell: :-1:

GrahamCampbell: FQCNs should always be used in phpdoc IMO.

build
lucasmichot: Is it PSR2 ?

GrahamCampbell: no, though the rest of the code base does that...

src/Deployer.php
gordalina: I disagree. FQCNs should be used if class is not part of file's namespace IMO.
So here it would be `\InvalidArgumentException`

lucasmichot: I am preparing a PR regarding the usage of FQN in docblocks

GrahamCampbell: @gordalina I don't understand why you'd want to make phpdoc hard to read? The whole point of phpdoc is for people to read, not for silly IDEs.

build
lucasmichot: Would make sense to add some more PHP-CS-Fixer fixers, but it's a bit on @elfet side

src/Deployer.php
gordalina: @GrahamCampbell i'm advocating the use of class names as you'd do if you were writing code. If you `use` a class or a namespace, then you should use that in the doc-block as well. 
Regarding your comment, i'm making it the same to read, you are making it hard to write.

elfet:
Again: why we need this?

lucasmichot:
It allows to have a more concise code.
Out of the function the developer has no clue of the arguments, so why the docblocks FQN are useful
But within the function it does not really makes sense to repeat many times the same namespace.

Also in the current code, sometimes the namespace is imported, sometimes it is not.

Importing FQN in code provide a common practice for all developer who wish to contribute to the project.
ATM, they don't have any clue what to do.

GrahamCampbell:
:+1:

bin/dep
```diff
@@ -34,9 +38,9 @@ require_once 'src/functions.php';
 define('DEPLOYER_BIN', __FILE__);
 
 // Init Deployer
-$console = new \Deployer\Console\Application('Deployer', 'master');
-$input = new \Symfony\Component\Console\Input\ArgvInput();
-$output = new \Symfony\Component\Console\Output\ConsoleOutput();
+$console = new Application('Deployer', 'master');
+$input = new ArgvInput();
+$output = new ConsoleOutput();
 $deployer = new \Deployer\Deployer($console, $input, $output);
```
gordalina: Why not use `Deployer\Deployer` here?

lucasmichot: I've skipped this one, my bad :+1:

lucasmichot:
BTW @elfet : https://github.com/deployphp/deployer/pull/368#discussion_r32720683
Some framework use lots of fixers https://github.com/laravel/framework/blob/5.1/.php_cs
Some less, it's really up to you

gordalina:
I would vote for a cs fixer, see why: https://github.com/deployphp/deployer/pull/370#issuecomment-113318748 (3rd paragraph)

lucasmichot:
Sure @gordalina and it's already implemented : https://github.com/deployphp/deployer/blob/master/.php_cs :-)

But shall we add more fixers to this .php_cs definition file ?

gordalina:
isn't there any CI that validates a PR with cs fixer? 
that would force correctness for all new code :smile:

GrahamCampbell:
> isn't there any CI that validates a PR with cs fixer? 

Yeh. https://styleci.io. :)

GrahamCampbell:
If you like the rules laravel uses, you can use the "laravel" preset on styleci. :)

GrahamCampbell:
https://github.com/StyleCI/Config/blob/master/src/Config.php

gordalina:
@GrahamCampbell does it notify the PR that its valid/invalid?
For example, this PR has no info from StyleCI

GrahamCampbell:
This PR has no info if it was sent after StyleCI was enabled. I think open/closeing this PR will trigger StyleCI to run. :)

GrahamCampbell:
> does it notify the PR that its valid/invalid?

Yes it does. It uses github's status API.

gordalina:
Awesome then!
Then ignore all my comments, as everything was setup before i've opened my mouth.

lucasmichot:
> This PR has no info if it was sent before StyleCI was enabled. I think open/closeing this PR will trigger StyleCI to run. :)

Can anyone @gordalina @elfet @ZeeCoder try to close and re-open this PR ?

GrahamCampbell:
@lucasmichot Can't you do it?

GrahamCampbell:
> I think open/closeing this PR will 

Looking at my code, the 3rd way of doing it is just to push to this PR. :)

GrahamCampbell:
A rebase and force push would do it for example.

lucasmichot:
Rebase or forced push

PR #370, opened 2015-06-18
Update FQN docblocks

lucasmichot:
And make them much more readable for users
Review appreciated

### DIFF
--- a/src/Console/Application.php
+++ b/src/Console/Application.php
@@ -18,10 +18,10 @@ class Application extends Console
     /**
      * Input definition for user specific arguments and options.
      *
-     * @var InputDefinition
+     * @var \Symfony\Component\Console\Input\InputDefinition
      */
     private $userDefinition;
-    
+
     /**
      * {@inheritdoc}
      */
@@ -68,14 +68,14 @@ class Application extends Console
     }
 
     /**
-     * @return InputDefinition
+     * @return \Symfony\Component\Console\Input\InputDefinition
      */
     public function getUserDefinition()
     {
         if (null === $this->userDefinition) {
             $this->userDefinition = new InputDefinition();
         }
-        
+
         return $this->userDefinition;
     }
 
--- a/src/Console/Output/OutputWatcher.php
+++ b/src/Console/Output/OutputWatcher.php
@@ -13,17 +13,17 @@ use Symfony\Component\Console\Output\OutputInterface;
 class OutputWatcher implements OutputInterface
 {
     /**
-     * @var OutputInterface
+     * @var \Symfony\Component\Console\Output\OutputInterface
      */
     private $output;
-    
+
     /**
      * @var bool
      */
     private $wasWritten = false;
 
     /**
-     * @param OutputInterface $output
+     * @param \Symfony\Component\Console\Output\OutputInterface $output
      */
     public function __construct(OutputInterface $output)
     {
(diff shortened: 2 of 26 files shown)


GrahamCampbell:
:+1:

elfet:
Why we need this? :smiley:

GrahamCampbell:
So the phpdoc is human readable. That's the whole point of phpdoc. :)

elfet:
This is more readable:
```php
     /**
      * Input definition for user specific arguments and options.
      *
      * @var InputDefinition
      */
```
when this:
```php
     /**
      * Input definition for user specific arguments and options.
      *
      * @var \Symfony\Component\Console\Input\InputDefinition
      */
```

GrahamCampbell:
Yeh, i think the long one is way better. You know exactly what it is. The first one could be anything.

elfet:
It could not be **anything**!  In you code you use `InputDefinition`, and not expect this to be anything? Same in phpdocs. 
PhpStorm fully supports this, by the way.

lucasmichot:
Hello @elfet ; yes for many developers it makes much more sense get the full FQN in docblocks.
Sure the `InputDefinition` example that you gives is true, but in many other case the developer might wonder which exact type of argument the function is using.

Deployer project would IMO greatly benefits from this kind of docblocks.

lucasmichot:
Also @elfet Deployer is growing very fast.
No one can really predict what it will be in 2 years, will it still use Symfony components, will it provide a system of plugins, will it use Interface dependency injection ?
You might have a class name and an interface having the same base name.
In this situation short FQN would become completely unreadable in docblocks.

GrahamCampbell:
> PhpStorm fully supports this, by the way.

As I said, phpdoc is not for IDEs. You could argue IDEs should run static code analysis and phpdoc isn't needed at all.

gordalina:
This isn't about IDEs (which may or may not use static code analysis especially because PHP is not strongly typed), it's about reading and writing code, it's for humans.

IMHO if you are `use`'ing a class or a namespace you should be able to use it on the doc block (without a FQN), if you do not `use` it the, you should definitely use the FQCN in the doc block.

Lets get a vote (or @elfet decides for all of us) to this and get this over with, there's no point in arguing, because humans tend to have different opinions, and that's ok. That's why PSR1&2 were done, to have a common way to write and style code so that we humans can interact with it better, faster and less annoyed.

elfet:
@gordalina lets do a little vote :+1:

gordalina:
Vote for:

1) Use FQCN in doc block if class isn't `use`d, or just class name if it is `use`d
2) Always use FQCN in doc block

I vote for 1.

GrahamCampbell:
I vote for 2. :)

ZeeCoder:
I did some google-ing before answering the question.

 - First of all, when I'm in doubt about standards / best practices, I always consult the Symfony github repo, where I found that the Symfony code [seems to use FQCNs](https://github.com/symfony/HttpFoundation/blob/master/Request.php#L203).
I think this weighs a lot here, since Deployer uses a lot of Symfony Components.
 - Another answer I found quite convincing: https://github.com/squizlabs/PHP_CodeSniffer/issues/325#issuecomment-63166037
 - Also: I do agree with @GrahamCampbell, [IDEs should not factor in the decision at all](https://github.com/deployphp/deployer/pull/370#issuecomment-113225127).
(I usually just use Sublime Text, Atom or Vim, for example.)
 - And I also agree with @lucasmichot here: https://github.com/deployphp/deployer/pull/370#issuecomment-113213742

Hence, due to the reasons above, I'm voting for **option 2**.

GrahamCampbell:
> I found that the Symfony code seems to use FQCNs

Laravel does too. :)

lucasmichot:
Well vote on number 2 as well :-)

PR #372, opened 2015-06-23
Change task order

ihsanudin:
Change task order to make sure the assets is already exist when assetic:dump task is running

### DIFF
--- a/recipe/symfony.php
+++ b/recipe/symfony.php
@@ -118,9 +118,9 @@ task('deploy', [
     'deploy:writable',
     'deploy:assets',
     'deploy:vendors',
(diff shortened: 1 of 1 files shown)


ZhukV:
The commands in Symfony:

assets:install
assetic:dump
cache:warmup
can be processed with error end returning not zero code.
If this commands return not zero code, then we should stop deploy process. And now (with this commit) we starting new release without assets (if assets not installed).

ZhukV:
The command `assets:install` has a `target` parameter for copy or link assets to directory. And the command `assetic:dump` has a `write_to` parameter for generate assets to directory.
So, best use this parameters for this command.

PR #379, opened 2015-06-25
Use mirrored git repository for faster results

czettnersandor:
The problem: big repositories takes too long to copy, even with the --reference. This solution will not copy, fetch any git objects, but uses only one mirror, updates it and then creates an archive to the release path.
    
The original idea is coming from Capistrano:
https://github.com/capistrano/capistrano/blob/ad61d9288a2567faa5ae37e7fa74b1c1947e592d/lib/capistrano/git.rb

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -131,7 +131,6 @@ task('deploy:update_code', function () {
         $at = "$branch";
     }
 
-    $releases = env('releases_list');
     if (run("if [ ! -f {{deploy_path}}/repo/HEAD ]; then echo \"1\"; fi")->toString()) {
         run("git clone --mirror $repository {{deploy_path}}/repo");
     } else {
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -21,14 +21,14 @@ env('branch', ''); // Branch to deploy.
 env('env_vars', ''); // For Composer installation. Like SYMFONY_ENV=prod
 env('composer_options', 'install --no-dev --verbose --prefer-dist --optimize-autoloader --no-progress --no-interaction');
 env('git_cache', function () { //whether to use git cache - faster cloning by borrowing objects from existing clones.
-  $gitVersion = run('git version');
-  $regs = [];
-  if (preg_match('/((\d+\.?)+)/', $gitVersion, $regs)) {
-      $version = $regs[1];
-  } else {
-      $version = "1.0.0";
-  }
-  return version_compare($version, '2.3', '>=');
+    $gitVersion = run('git version');
+    $regs = [];
+    if (preg_match('/((\d+\.?)+)/', $gitVersion, $regs)) {
+        $version = $regs[1];
+    } else {
+        $version = "1.0.0";
+    }
+    return version_compare($version, '2.3', '>=');
 });
 
 /**
@@ -82,7 +82,7 @@ task('deploy:prepare', function () {
 
         throw $e;
     }
-    
+
     run('if [ ! -d {{deploy_path}} ]; then echo ""; fi');
 
     // Create releases dir.
@@ -128,7 +128,7 @@ task('deploy:update_code', function () {
     $branch = env('branch');
     $gitCache = env('git_cache');
     $depth = $gitCache ? '' : '--depth 1';
-    
+
     if (input()->hasOption('tag')) {
         $tag = input()->getOption('tag');
     }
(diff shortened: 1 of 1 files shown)


oanhnn:
:+1:  Good ideal!
Please create new PR for magento recipe.

czettnersandor:
Done.

elfet:
@deployphp/maintainer what you thinks about this PR? Can anybody test it?

elfet:
One issue: https://scrutinizer-ci.com/g/deployphp/deployer/inspections/e0dc283a-1aac-48e5-b007-68b349f9bdb1/issues/files/recipe/common.php?status=new&orderField=path&order=asc

czettnersandor:
Sorry about the unused variable, it was fixed on the next commit which I removed accidentally when took out the Magento recipe

czettnersandor:
Testing would be much welcomed as it's a core part of Deployer. I tested it with a few of our repositories, but maybe other users have different configurations. I'm trying to switch from Capistrano and a slow deployment with lots of io is a no go for us. This one reduced the original deployment time of a big repository from 10 minutes to 23s.

On the big repository, we have an 1G sql file, updated a few times since the initial commit, which slowed down the clone a lot. With a mirror, we don't even have to get any .git objects.

johnny-bit:
Imo it's great thing to have, but implementation is too invasive. In capistrano strategy is configurable. When I created git cache pr, the whole thing was configurable with `use-if-possible` by default. To some this might be problematic depending on repository configuration, plus You'd have to actually keep additional dir on deployment target (and I got clients bitching about additional dirs and unneded versions since forever). And that dir would have to be there etc etc...

My position: make it configurable, just like I did with git cache. That way anybody would be able to choose strategy, be it shallow clone, reference or mirrored repo.

czettnersandor:
It would be possible to make it configurable. However, it's more thing to maintain. Also, the previous behaviour didn't hide the .git directory, so I discovered I can access /.git/config and /.git/index from the web so in theory I could download every git objects if I'm patient enough. With directory listing enabled, it would be much easier.

In my case, the change uses much less disk space because the releases does not have the .git folder (5 folders less with the default keep-releases). It's a design change and I can't see it could practically live together with the old behaviour, but it's the maintainer's task to decide. Worst case, I have to use a private fork.

johnny-bit:
You don't *have* to use fork if it's so hard to maintain. If maintainer decides against it, then there's always deployphp/recipes repo in which You can add Your own recipes and in which recipe that changes 2 tasks can live happily :smiling_imp:

With `.git` present - I guess it depends on what person wants. If using shallow clones, it would actually do no harm to simply cleanup that dir post clone. In case of git cache with reference - it's simpy needed (aand securing access from outside to that is excercise to the admin/depops). With Your way - it's simply gone :)

But no need to listen to me - I am all for freedom of choice at expense of simplicity and maintainability. :joy_cat:

czettnersandor:
Ok if this does not get to the next version, I shall move it to deployphp/recipes, much better to not have to maintain a private fork. I'm also open for improvements.

kenjis:
> In capistrano strategy is configurable. 

I hope I can configure strategy. 

> It's a design change and I can't see it could practically live together with the old behaviour,

Yes, it can't live together. So users select one.

Speaking of `.git`, I think the directory structure is not secure. It seems you put too much files in Web document root.

I mean, for example,

~~~
project/        <-- not here
project/.git/
project/src/
project/public/ <-- this should be doc root.
project/public/index.php
~~~

czettnersandor:
I didn't expect so much resistance (well, 2 so far) so could you guys tell me why you need git objects in each release? Is it the case you are modifying files on the production server and committing to git from there?

Some of my repositories are 1GB big, this means 5GB additional disk space when I keep only 5 releases. Also, using bare mirror in this PR is speeding up the deployment significantly without any side effects.

Moving everything to a public folder except the files needed for development is not a solution, but perhaps I'm just too convenient after a few years RoR development. IDE's are not behaving like this by default, composer requires it's files and the vendor folder in the document root, and then there's the problem how the framework can support this. You have to add far more files anyway.

johnny-bit:
I'm NOT oposing this, quite to the contrary. I just say that some CAN and
WILL object. And those that will object are users... And if You leave them
no choice they WILL leave. That's what We're trying to avoid here.

I advise that You do same thing as I did: make it configurable. Then those
like You will enable it and those who need it other way will disable it and
use shalow clondes/referencee/local copy/whatever.

2015-06-27 12:04 GMT+02:00 Sándor Czettner <notifications@github.com>:

> I didn't expect so much resistance (well, 2 so far) so could you guys tell
> me why you need git objects in each release? Is it the case you are
> modifying files on the production server and committing to git from there?
>
> Some of my repositories are 1GB big, this means 5GB additional disk space
> when I keep only 5 releases. Also, using bare mirror in this PR is speeding
> up the deployment significantly without any side effects.
>
> Moving everything to a public folder except the files needed for
> development is not a solution, but perhaps I'm just too convenient after a
> few years RoR development. IDE's are not behaving like this by default,
> composer requires it's files and the vendor folder in the document root,
> and then there's the problem how the framework can support this. You have
> to add far more files anyway.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/deployphp/deployer/pull/379#issuecomment-116000760>.
>



-- 
Pozdrawiam,
Hubert Kowalski

kenjis:
> could you guys tell me why you need git objects in each release? 

Just I like it. I can see git log and status on production server.
But if I had 1GB git repository, I would reconsider using git repository.

So I also do not oppose it, I may use it. But Capistrano has some strategy, so why does Deployer do also?

I think it is better Deployer has functionality to have some strategies for a task.

> Is it the case you are modifying files on the production server and committing to git from there?

Of course not.

elfet:
Already implemented with git clone references.

PR #380, opened 2015-06-26
Magento recipe

czettnersandor:
Added Magento recipe.


elfet:
Cool! Can you add docs?

PR #384, opened 2015-07-02
Add --force-color (or -c) option to force output coloring

piotrekkr:
Usefull when redirecting Deployer output inside shell script. Since output is redirected, posix_isatty() returns false and coloring is disabled by default. This option forces deployer to color output.

### DIFF
--- a/bin/dep
+++ b/bin/dep
@@ -39,8 +39,13 @@ $input = new \Symfony\Component\Console\Input\ArgvInput();
 $output = new \Symfony\Component\Console\Output\ConsoleOutput();
 $deployer = new \Deployer\Deployer($console, $input, $output);
 
-// Require deploy.php script
-$options = getopt('f::', ['file::']);
+$options = getopt('f::c', ['file::', 'force-color']);
+
+if (isset($options['c']) || isset($options['force-color'])) {
+    // force colored output
+    $deployer->getOutput()->setDecorated(true);
+}
+
 $userSpecifiedFile = null;
 
 if (isset($options['f'])) {
--- a/bin/dep
+++ b/bin/dep
@@ -39,13 +39,8 @@ $input = new \Symfony\Component\Console\Input\ArgvInput();
 $output = new \Symfony\Component\Console\Output\ConsoleOutput();
 $deployer = new \Deployer\Deployer($console, $input, $output);
 
-$options = getopt('f::c', ['file::', 'force-color']);
-
-if (isset($options['c']) || isset($options['force-color'])) {
-    // force colored output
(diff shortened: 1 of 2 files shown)


elfet:
Cool! But why you need it?

piotrekkr:
I'm using bash script when deploying new versions. It works like that:

1. check existance of backup and log directories
1. create log file and log everything that is on screen 
2. start timer
3. use Deployer to deploy code on servers
4. show deploy time

I like to see colored output when I use Deployer but when I redirect Deployer output like:

```
php vendor/bin/dep deploy prod | tee "$log_path"
```
or 
```
exec > >( tee /dev/tty | sed $'s/\E[^m]*m//g' > "$log_path" ) 2>&1
```
colors are lost because `posix_isatty()` return `false`.

kenjis:
I also do:
~~~
php deployer.phar deploy -vvv production | tee $log
~~~

But I can see colored output. What's the difference?
I use Ubuntu and LXTerminal.

piotrekkr:
@kenjis Sorry I forgot to add stderr redirection, try this:
```
php vendor/bin/dep 2>&1 | tee test.log
```

I mostly use:
```
exec > >( tee /dev/tty | sed $'s/\E[^m]*m//g' > "$log_path" ) 2>&1
```
because i don't want to to write `command 2>&1 | tee -a $log` to every command. I just redirect `stderror` to `stdout` and use `tee` to log everything at the begining of bash script.

bin/dep
```diff
@@ -39,8 +39,13 @@ $input = new \Symfony\Component\Console\Input\ArgvInput();
 $output = new \Symfony\Component\Console\Output\ConsoleOutput();
 $deployer = new \Deployer\Deployer($console, $input, $output);
 
-// Require deploy.php script
-$options = getopt('f::', ['file::']);
```
kenjis: We need above line before `if (isset($options['f'])) {`, don't we?

kenjis:
@piotrekkr I got no color output. Agree with `--force-color` option.

elfet:
@piotrekkr what about using `option` func instead of `getopt`?

piotrekkr:
@elfet Do you mean to only check `--force-color` or `-c` inside `Deployer::run()` ?

piotrekkr:
@elfet Ok my entire PR is going to trash :) I just looked at `\Symfony\Component\Console\Application` and there is option:

```
--ansi            Force ANSI output
```
Which is doing same thing as my --force-color. You can reject my PR. Thx.

elfet:
Cool! :)

kenjis:
Good! I got colored output.

Issue #206, opened 2015-03-02
Copy files between deployment


nicodmf:
As capistrano as this is a very common pratice to copy files between two deployments.

A common use case is with composer and the vendor dir : it can be simply copied to gain time.

elfet:
Cool thing. Will be implemented 100%.

tomzx:
This can't be done if dependencies have changed. You'd either have to monitor `composer.json/lock` between deployment and decide whether or not it has changed => needs to run `composer install` or blindly assume dependencies won't *ever* change (don't do that).

Furthermore, what kind of use case do you have that makes composer takes *minutes* to install? Are you using `composer update` by any chances? Composer uses a cache mechanism which will make reinstalling the same libraries/dependencies extremely fast since it will be similar to a folder copy, so I don't see the point of that use case.

Let me know if I'm missing something...

elfet:
> Composer uses a cache mechanism which will make reinstalling the same libraries/dependencies extremely fast since it will be similar to a folder copy, so I don't see the point of that use case.
WOW! Cool think! Need more to discuss.

nicodmf:
@tomzx if you copy the entire directory, composer check the states of dependancies. The composer install command is always executed to ensure that a change is taken in charge. Try this work!!!

        rm -rf vendor
        time composer install 
        real    0m54.926s

with only one dev dependancy

and the case is the same with bower : 

       real    0m38.031s

so one minute gain on just two directory and the use case could be make by another local ip datatbases or many other personnal stuff.

nicodmf:
i have retake the test after you affraid me and it's sure if the composer.lock change all dependencies are updated.

i don't know why i don't trust me in this case...

the log : 

composer install
Loading composer repositories with package information
Installing dependencies (including require-dev) from lock file
  - Removing doctrine/doctrine-bundle (v1.3.0)
  - Installing doctrine/doctrine-bundle (v1.4.0)
    Downloading: 100%

  - Removing michelf/php-markdown (1.4.1)
  - Installing michelf/php-markdown (1.5.0)
    Downloading: 100%

  - Removing sensio/distribution-bundle (v3.0.16)
  - Installing sensio/distribution-bundle (v3.0.18)
    Downloading: 100%

Generating autoload files
Updating the "app/config/parameters.yml" file
Clearing the cache for the dev environment with debug true

oanhnn:
@nicodmf I think you should use symbolic link (shared_dirs and shared_files) for releases.
```php
[...]

PR #385, opened 2015-07-05
WIP/POC - copy directories

nicodmf:
see #206 

If you agree this change i update doc, tests and rebase

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -166,10 +166,10 @@ task('deploy:copy_dirs', function () {
 
     foreach ($dirs as $dir) {
         //Delete directory if exists
-        run("rm -rf {{release_path}}/$dir");
+        run("if [ -d $(echo {{release_path}}/$dir) ]; then rm -rf {{release_path}}/$dir; fi");
 
-        //Copy files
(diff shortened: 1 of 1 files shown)


johnny-bit:
I'm okay with this, but have You tried `rsync:warmup` from `deployphp/recipes/rsync.php`? I guess it does everything Your `deploy:copy_dirs` does, ony You can leverage all of rsync glory :wink:

nicodmf:
Rsync is a good solution for many case. 

Copy (cp) is installed on all system, so i think there solutions are complements one from other. Rsync if you install it and use on the same serveur or not and copy for simpliest cases.

nicodmf:
@ZeeCoder Have you a point of view ?

elfet:
@deployphp/maintainer this is really requested feature. We need to think how is better to bring it.

nicodmf:
@AlexStansfield if you think this is usefull and writed as you wait, you can add a comment here

AlexStansfield:
Just added PR #406 for copying vendors between releases

elfet:
I think cool feature. @nicodmf can you add docs for this?

PR #388, opened 2015-07-07
add support forward agent feature when using ext-ssh2

oanhnn:
In this PR:
* add support forward agent feature when using `ext-ssh2`


oanhnn:
Ping @elfet @ZeeCoder

elfet:
Does `ext-ssh2` support it?

oanhnn:
I think it supported.
http://php.net/manual/en/function.ssh2-auth-agent.php

And I was try test with some my projects. It run successful. :+1:

PR #399, opened 2015-07-24
Add vendor to share dir

Kleinast:
vendor folder should be a shared folder. no need to download all package each time.

### DIFF
--- a/recipe/symfony.php
+++ b/recipe/symfony.php
@@ -12,7 +12,7 @@ require_once __DIR__ . '/common.php';
  */
 
 // Symfony shared dirs
-set('shared_dirs', ['app/logs']);
+set('shared_dirs', ['app/logs', 'vendor']);
 
 // Symfony shared files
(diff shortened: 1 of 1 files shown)


johnny-bit:
I concur. Rarely, but surely that may simply not work. Composer caches packages anyway, so it's fast anyway. Problem arises when current release has old dependencies and You are releasing new stuf. now reverting THAT would be a huge pain in the ass.

ZeeCoder:
I'm sorry but that's a terrible idea. It would make rolling back almost non-achievable. If you must, you can add  `vendor` to the shared dirs yourself.

AlexStansfield:
I agree with @ZeeCoder, this is bad practise and it would destroy the ability to rollback easily. If you want to add it to shared you can but really you shouldn't.

I added a feature in 2.x that would copy the vendors folder from the current release before ```composer install``` so only changed libraries would need to be fetched.

It looks like this never made it across to 3.x.

PR #406, opened 2015-07-28
Allow Copy vendors from previous release to latest release

AlexStansfield:
Using ```set('composer_copy_vendors', true);``` the ```deploy:vendors``` task will check for existing vendors directory in previous release and copy to new release before running ```composer install```.

See #399 and #206

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -8,6 +8,7 @@
 /**
  * Common parameters.
  */
+set('composer_copy_vendors', false);
 set('keep_releases', 3);
 set('shared_dirs', []);
 set('shared_files', []);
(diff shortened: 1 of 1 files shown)


elfet:
Cool feature! Can you add some tests?

AlexStansfield:
Sure thing, will need to figure out a good way of doing that.

AlexStansfield:
Sorry for the delay, life got in the way. Hope to have this finished soon

AlexStansfield:
So I like the idea of this setting because firstly it existed in 2.x and secondly as deployer specifically supports composer it seems to make sense to be able to have this option as a setting.

Having said that I like the idea of an array of directories that will get copied. I think we can have the best of both worlds where you have the array of directories and the setting in this PR just adds the vendor directory to that array.

AlexStansfield:
@elfet & @gquemener any thoughts on that?

gquemener:
I think this is a great solution @AlexStansfield !

Moreover, I would add another 'ask_before_copy' option that displays confirmation message before copying directories (default to true).

AlexStansfield:
@elfet just wondered if we could get your thoughts before I press ahead with this

elfet:
Composer it selfs cache vendors in `~/.composer` dir. So this is pretty fats evan this out this.

Issue #408, opened 2015-07-28
Run command on different server


davidkudera:
Hi,

please, is it possible to somehow run command on different server? I have for example task `database:sync` which I run like this

```
$ dep database:sync test
```

I have two servers defined:
* `production`
* `test`

Now I would like to connect to `production`, create database dump, download it and then upload it to the test server.

If it is not possible (maybe yet :smile:), is there at least some workaround please?

Thank you

davidkudera:
Ou.. I just found #355 so I'll try the suggested solution. Sorry if this is a duplicate

davidkudera:
So I tried `onlyOn` method, but that will only run task if given server name (from command) is same

elfet:
You can set dumping by onlyOn on prod server.

oanhnn:
@davidkudera 
Can you try?
```php
task('database:dump', function(){
    // TODO: dump and download
})->onlyOn('production');

task('database:restore', function(){
    // TODO: upload and restore
})->onlyOn('test');

task('database:sync', [
     'database:dump',
     'database:restore',
]);

server('test', '192.168.1.2', 22)
    ->user('your_user')
    ->identityFile()
    ->stage(['dev'])
    ->env('deploy_path', '/var/www/apps/your_app')
    ->env('branch', 'master');

server('production', '192.168.1.3', 22)
    ->user('your_user')
    ->identityFile()
    ->stage(['dev'])
    ->env('deploy_path', '/var/www/apps/your_app')
    ->env('branch', 'master');
```
and run:
```
dep database:sync dev -vvv
```
Goodluck!

davidkudera:
@oanhnn that's what I tried :wink: 

```php
task('database:dump', function(){
	write('DUMPING...');
})->onlyOn(['production']);

task('database:restore', function(){
	write('RESTORING...');
});

task('database:sync', [
	'database:dump',
	'database:restore',
]);
```

**output:**

```
✔ Executing task database:dump
➤ Executing task database:restore
RESTORING...✔ Ok
```

oanhnn:
@davidkudera 
With your case, please try my script.
Your script missing `->onlyOn('test')` and settings server stage `dev`

davidkudera:
@oanhnn I didn't noticed your stage setup, sorry...

So with that setup:

```php
localServer('production')
	->stage(['dev', 'david_work', 'test']);

localServer('test')
	->stage(['dev', 'test']);

localServer('david_work')
[...]

Issue #355, opened 2015-06-16
Task chain that requires 2+ remote servers


johnny-bit:
I have to write weird piece. Usually that would be 3 separate commands, but it would be great to be able to do in one:

1. Connect to server 1, issue command to create file on server 1
2. Dowload file locally (or upload to server2 directly if possible, then skip to point 4)
3. Connect & upload file downloaded to server 2
4. Issue command on server 2 to accept file

Reason for this: when developing long-running wordpress site (for example) it's great to have production database on staging, and vice versa - have changes applied on staging work on production witout lengtly introducing them.

Any suggestions for that?

elfet:
Yes, try to use `onlyOn`. :wink: 
```php
task(...)->onlyOn('server 1');
task(...)->onlyOn('server 2');
task(...)->onlyOn('server 1');
```

oanhnn:
@johnny-bit 
Are you haven't any other questions about @elfet 's suggestion?
I am closing thing issue but feel free to reopen if you think there needs to be more discussion.

PR #410, opened 2015-07-28
Task: syntactic suger for onlyOn method

davidkudera:
Just little bit of syntactic sugar. What do you think?

**String at onlyOn method**
[That](https://github.com/deployphp/deployer/issues/408#issuecomment-125478786) example from #408 didn't work and [here](https://github.com/deployphp/deployer/issues/355#issuecomment-112614911) it was suggested to use strings too. But the method internally works with arrays, so it should either accept both arrays and strings or I think that there should be an `array` type hint.

### DIFF
--- a/src/Task/Task.php
+++ b/src/Task/Task.php
@@ -126,12 +126,12 @@ class Task
     }
 
     /**
-     * @param array $servers
+     * @param array|string $servers
      * @return $this
      */
-    public function onlyOn($servers)
+    public function onlyOn($servers = [])
     {
-        $this->onlyOn = array_flip($servers);
+        $this->onlyOn = array_flip(is_array($servers) ? $servers : func_get_args());
         return $this;
     }
 
--- a/test/src/Task/TaskTest.php
+++ b/test/src/Task/TaskTest.php
@@ -37,6 +37,13 @@ class TaskTest extends \PHPUnit_Framework_TestCase
 
         $task->onlyOn([]);
         $this->assertTrue($task->runOnServer('server'));
+
+        $task->onlyOn('server');
+        $this->assertEquals(['server' => 0], $task->getOnlyOn());
+        $this->assertTrue($task->runOnServer('server'));
+
+        $task->onlyOn();
+        $this->assertTrue($task->runOnServer('server'));
         
         $task->setPrivate();
         $this->assertTrue($task->isPrivate());


src/Task/Task.php
```diff
@@ -126,12 +126,17 @@ public function isOnce()
...
      * @return $this
      */
-    public function onlyOn($servers)
+    public function onlyOn($servers = null)
     {
-        $this->onlyOn = array_flip($servers);
+        if ($servers === null) {
+            $this->onlyOn = [];
+        } else {
+            $this->onlyOn = array_flip(is_array($servers) ? $servers : [$servers]);
```
davidkudera: I was also thinking about `is_array($servers) ? $servers : func_get_args()`

oanhnn:
I think this method should internally works with arrays.
We should not complicate it.

davidkudera:
I agree about not complicating things, so what do you think about adding `array` type hint? Because now I think that many people (myself, you and owner of repo including) actually try strings, but it will [throw](http://3v4l.org/pONFB) completly unrelated error.

oanhnn:
Yes, it is added `array` type hint by phpdoc
```
    /**
     * @param array $servers
     * @return $this
     */
```

davidkudera:
Sorry I didn't mean phpdoc, but method type hint.

```php
public function onlyOn(array $servers) {}
```

davidkudera:
Error [http://3v4l.org/FLQhN](http://3v4l.org/FLQhN) is much better than [http://3v4l.org/gGq4o](http://3v4l.org/gGq4o).. At least I think :wink:

oanhnn:
Yes, i agree, we will use method type hint.
Can you create PR?

davidkudera:
Sure thing. Should I also add default value? `array $servers = []`

elfet:
I think it's better to allow strings. 
I vote to merge!
@deployphp/maintainer what do you think?

src/Task/Task.php
rodnover55: This code may be easier

```PHP
    public function onlyOn($servers = []) {
        $this->onlyOn = array_flip(is_array($servers) ? $servers : [$servers]);
    }
```

davidkudera: Yep, I was actually suggesting that bellow and was waiting for more responses. Thank you sir anyway

davidkudera:
I simplified `onlyOn` method like @rodnover55 said

oanhnn:
@elfet @davidkudera 
I think it is easy using for users if param  always be array.
Document will be clearer and run faster!
```php 
->onlyOn('dev');
// or
->onlyOn(['dev']);
```

elfet:
I think it can prevent some developers from errors:  `->onlyOn('dev');` 
Even it will complicate it a little bit. 
@davidkudera you not agree? I'm thinking to merge it.

davidkudera:
@elfet I agree with @oanhnn that things should not be too complicated, but than I noticed that both of you used `->onlyOn('dev')` :smile: Thats the true reason I created this PR. 

The only thing that remains is if I should keep it like this:

```php
array_flip(is_array($servers) ? $servers : [$servers]);
```

or change it to

```php
array_flip(is_array($servers) ? $servers : func_get_args());
```

People than can use `->onlyOn('dev', 'test', 'production')`, but thats maybe too much of magic

davidkudera:
Hi, any improvement on this? Did you decide if you want this feature or if I should change it somehow?

davidkudera:
rebased to master

elfet:
I vote for `array_flip(is_array($servers) ? $servers : func_get_args());` if also get good docs for this. Will be awesome!

davidkudera:
@elfet should I wait for some other votes or do you want me to change it? I can write docs too

elfet:
@davidkudera i think do it :) If we will have great docs - everything will be ok.

elfet:
@davidkudera can you set `array_flip(is_array($servers) ? $servers : func_get_args());`?
I want to merge it)

davidkudera:
Sorry, I forgot, so thanks for reminding me.. Btw done :wink:

elfet:
Cool! Thanks!

Please add some docs on this feature!

PR #411, opened 2015-07-28
Local process: add option to configure timeout with process_timeout

davidkudera:
Unfortunately there is no way to change process timeout for commands running on local server. Method `runLocally` has that option.

### DIFF
--- a/src/Server/Local.php
+++ b/src/Server/Local.php
@@ -26,10 +26,12 @@ class Local implements ServerInterface
      */
     public function run($command)
     {
+        $timeout = env()->get('process_timeout', self::TIMEOUT);
+
         $process = new Process($command);
         $process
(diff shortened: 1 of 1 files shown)


oanhnn:
I think all servers (local or remote) should support timeout link pull request #328 
And not using env var.

elfet:
Agree with @oanhnn.

davidkudera:
Yep, you are of course right. Sorry I didn't noticed that pull request :wink:

elfet:
But that PR is hang out. May be you can fork this PR, fix suggestion and create your own PR baser on that? @davidkudera?

davidkudera:
@elfet I'm thinking about it too, so I'll probably do that, because I really need that

elfet:
Cool!

PR #413, opened 2015-07-28
Task: add runAlwaysOn option

davidkudera:
Adds new option `runAlwaysOn` to task, based on #408. 

**Configuration:**

```php
localServer('production');
localServer('test1');
localServer('test2');

task('database:sync', function() {
    write('syncing database from '. env('server.name'));
})->runAlwaysOn('production');
```

**Test:**

```
$ dep test production
output: syncing database from production

$ dep test test1
output: syncing database from production

$ dep test test2
output: syncing database from production
```

Now it doesn't matter on which server you want to run some task, it will just always run on the defined one. That can be easily used for example for syncing database from production to other servers (independently). 

**Todo:**

* Is `runAlwaysOn` good name?
* Is it possible to somehow test at which server current task runs? `env('server.name')` can't be used
* I don't know how to test `ParallelExecutor`. At least not yet

### DIFF
--- a/src/Executor/ParallelExecutor.php
+++ b/src/Executor/ParallelExecutor.php
@@ -285,6 +285,10 @@ class ParallelExecutor implements ExecutorInterface
                     $this->tasksToDo = [];
 
                     foreach ($this->servers as $serverName => $server) {
+                        if (($runnerServerName = $task->getRunAlwaysOn()) && isset($this->servers[$runnerServerName])) {
+                            $serverName = $runnerServerName;
+                        }
+
                         if ($task->runOnServer($serverName)) {
                             $this->informer->onServer($serverName);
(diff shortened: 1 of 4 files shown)


davidkudera:
I'll be able to test it at real application tomorow at work

elfet:
This is really disputable feature. We need really discuss this before merging. (and test parallel execution)
@deployphp/maintainer do we need a vote on this feature?

davidkudera:
rebased to master

elfet:
@davidkudera this is cool feature, but i think its really edge case. 
I'm working on plugin system for Deployer v4 with will be able to add your functionality as plugin.

davidkudera:
Ok, I'll use my fork for now. Btw plugins would be really really great :wink:

PR #415, opened 2015-07-29
Add support for timeout option

davidkudera:
Based on #328 and #411.

Unfortunately it seems that php-ssh doesn't support timeout - Herzult/php-ssh#46

### DIFF
--- a/src/Server/Remote/PhpSecLib.php
+++ b/src/Server/Remote/PhpSecLib.php
@@ -16,6 +16,8 @@ use RuntimeException;
 
 class PhpSecLib implements ServerInterface
 {
+    const TIMEOUT = 10;
+
     /**
      * @var Configuration
      */
@@ -103,10 +105,15 @@ class PhpSecLib implements ServerInterface
     /**
      * {@inheritdoc}
      */
-    public function run($command)
+    public function run($command, $timeout = null)
     {
+        if ($timeout === null) {
+            $timeout = self::TIMEOUT;
+        }
+
         $this->checkConnection();
 
+        $this->sftp->setTimeout($timeout);
         $result = $this->sftp->exec($command);
 
         if ($this->sftp->getExitStatus() !== 0) {
--- a/src/Server/Remote/SshExtension.php
+++ b/src/Server/Remote/SshExtension.php
@@ -109,8 +109,12 @@ class SshExtension implements ServerInterface
     /**
      * {@inheritdoc}
      */
-    public function run($command)
+    public function run($command, $timeout = null)
     {
+        if ($timeout !== null) {
+            throw new \RuntimeException('Timeout is not supported in ssh connection.');
+        }
+
         $this->checkConnection();
 
         return $this->session->getExec()->run($command);
--- a/src/Server/Local.php
+++ b/src/Server/Local.php
@@ -24,12 +24,19 @@ class Local implements ServerInterface
     /**
      * {@inheritdoc}
      */
-    public function run($command)
+    public function run($command, $timeout = null)
     {
         $process = new Process($command);
+
+        if ($timeout === null) {
+            $timeout = self::TIMEOUT;
+        } elseif ($timeout === 0 || $timeout === false) {
+            $timeout = null;
+        }
+
         $process
-            ->setTimeout(self::TIMEOUT)
-            ->setIdleTimeout(self::TIMEOUT)
+            ->setTimeout($timeout)
+            ->setIdleTimeout($timeout)
             ->mustRun();
 
         return $process->getOutput();
--- a/src/Server/ServerInterface.php
+++ b/src/Server/ServerInterface.php
@@ -17,9 +17,10 @@ interface ServerInterface
     /**
      * Run shell command on remote server.
      * @param string $command
+     * @param int $timeout
      * @return string Output of command.
      */
-    public function run($command);
+    public function run($command, $timeout = null);
 
     /**
      * Upload file to remote server.
--- a/src/functions.php
+++ b/src/functions.php
@@ -277,9 +277,10 @@ function workingPath()
  * Run command on server.
  *
  * @param string $command
- * @return Result
+ * @param int $timeout
+ * @return \Deployer\Type\Result
  */
-function run($command)
+function run($command, $timeout = null)
 {
     $server = Context::get()->getServer();
     $command = env()->parse($command);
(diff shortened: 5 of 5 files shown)


elfet:
Cool! Need to code review and test it. 
@deployphp/maintainer please, help!

oanhnn:
Cool!
I think it is better if support:
```
localServer($name, $timeout = 300);
// and
server($name, $host, $port = 22, $timeout =300);
```

davidkudera:
@oanhnn Well, I made it like this because @elfet said that it would be better (https://github.com/deployphp/deployer/pull/328#discussion_r32095173). If I understand correctly

oanhnn:
@elfet What do you think?
@davidkudera Can you write docs and create a PR to https://github.com/deployphp/docs

davidkudera:
@oanhnn sure, I'll do that right after you'll decide which version is better (method or argument)

oanhnn:
@elfet Do you decide which version is better (method or argument).
```
server(...)
    ->timeout(300)
```
or 
```
server('dev', 192.168.1.2, 22, 300)
```
or both :smile:

elfet:
I dont think what is good to setup this server wide. Its better to specify timeout on run:
```php
run(‘…’, 360);
```

davidkudera:
@elfet yes that is only logical since `runLocally` also has it. Actually I was thinking about it too, but I didn't want to do that in my first PR because it worked just for local server and it would mean to change implementation of `ServerInterface`. Also I made just the changes mentioned in the second PR. 

But what if we'll change `ServerInterface::run()` but keep global server configuration too? If user provide second argument at `run` method, we'll use that, othervise we'll use value from server configuration. And for people using ssh we can throw some `NotImplementedException` for example.

What do you think?

elfet:
I think global config here is bad decision.

davidkudera:
Ok, so should I change it? @elfet @oanhnn

elfet:
I think yes :)

davidkudera:
@elfet ok, I'll force push these changes. Unfortunately right now I don't have time, but I should be back home maybe at sunday

davidkudera:
@elfet @oanhnn what do you think now guys?

src/Server/Remote/PhpSecLib.php
```diff
@@ -102,10 +104,15 @@ public function checkConnection()
...
+    public function run($command, $timeout = null)
     {
+        if ($timeout === null) {
+            $timeout = self::TIMEOUT;
+        }
+
         $this->checkConnection();
 
+        $this->sftp->setTimeout($timeout);
         $result = $this->sftp->exec($command);
```
torinaki: How about to add here check that operation wasn't reached timeout like that
        
        if ($this->sftp->isTimeout()) {
            throw new \RuntimeException('Operation timeout');
        }
        
// cc @elfet , @oanhnn

davidkudera:
ping @elfet @oanhnn

davidkudera:
rebased to master

src/Server/Remote/PhpSecLib.php
elfet: What is `isTimeout()`?

elfet:
```
Timeout is not supported in ssh connection.
```
WE really need to figure out how to implement this in ssh2 ext. Otherwise timeout with be really strange functionality.

src/Server/Remote/PhpSecLib.php
torinaki: @elfet it is related to my latest temp.fixed issue with timeout(see PR #429 )
Currenlty if timeout will be reached deployer will silently stop task without any notification and will continue with the next one. This exception will at least help to understand what was happen. 

This flag will be set to true if timeout happen:
https://github.com/phpseclib/phpseclib/blob/7da4f2e6307d11c5984906dafe3c53d4f7c157a1/phpseclib/Net/SSH2.php#L2823

Here is oficial documentation about timeout: http://phpseclib.sourceforge.net/ssh/examples.html#timeout

torinaki: BTW `$this->sftp->exec($command)` will not throw any exception about timeout

davidkudera:
Yeah, but do we know if ssh2 is still developed? Last commit was 5 months ago..

torinaki:
@elfet , @davidkudera SSH wrapper library use stream blocking. It's easy to refactor code to imlement not blocking implementation with callback:
https://github.com/Herzult/php-ssh/blob/d569d15483a5f9218700313cd21c50090aa93351/src/Ssh/Exec.php#L26

But I'm still wonder about use cases of this timeout. @davidkudera can you please provide some examples of use cases? Specialy for me. I'm just curious.

torinaki:
Or another idea is to wrap command with some shell timeout function.
Windows: http://stackoverflow.com/questions/12394338/windows-batch-analogue-for-timeout-command
Linux: http://man7.org/linux/man-pages/man1/timeout.1.html
Without any magic :)

davidkudera:
@torinaki sure, for example we were developing new version of our website for more than half of a year (without touching 1st version). My deploy process failed almost immediatelly - to be more specific at database migrations. On my PC with linux it runs for about 15 minutes and I wasn't even brave enough to try it on Windows :grinning:

torinaki:
@davidkudera for what exact task you suppose to use `timeout`? Database migrations?

davidkudera:
Yes, for now only for database migrations, but maybe there will be some other tasks too.

src/Server/Remote/SshExtension.php
```diff
@@ -109,8 +109,12 @@ public function checkConnection()
     /**
      * {@inheritdoc}
      */
-    public function run($command)
+    public function run($command, $timeout = null)
     {
+        if ($timeout !== null) {
+            throw new \RuntimeException('Timeout is not supported in ssh connection.');
```
elfet: So sad( But if some one try to run it with ext-ssh?

davidkudera: It should just throw this exception, or what do you mean? Maybe the exception should have better message?

davidkudera:
I just noticed that this PR is still open. Is there anything I can do now to make it merged?

The problem with SSH is unfortunate but I think that the solution proposed in this PR is still much better than the current state of deployer.

elfet:
Ok, i'll test this PR to merge. Need some time.

ceesvanegmond:
Still open, @elfet When does it get merged?

torinaki:
@elfet from my point of view it is important that following comment should be addressed before merge:
https://github.com/deployphp/deployer/pull/415#discussion_r37126085
otherwise deployer will silently stop task in case when it wil reaches timeout.

TheRatG:
+1

PR #417, opened 2015-07-29
Added ability to enable ssh config via serverList yaml functionality.

kmaid:
Corrected comments
Refactored default setting into builder class to match other defaults location


elfet:
Cool!

PR #429, opened 2015-08-15
Temporary fixes problem with command execution timeout

torinaki:
It temporary fixes problem that described in phpseclib/phpseclib#775.
For example `composer install` may be stoped on the middle of progress if it takes more than 10 sec.

### DIFF
--- a/src/Server/Remote/PhpSecLib.php
+++ b/src/Server/Remote/PhpSecLib.php
@@ -46,7 +46,7 @@ class PhpSecLib implements ServerInterface
     public function connect()
     {
         $serverConfig = $this->getConfiguration();
-        $this->sftp = new SFTP($serverConfig->getHost(), $serverConfig->getPort());
+        $this->sftp = new SFTP($serverConfig->getHost(), $serverConfig->getPort(), PHP_INT_MAX);
 
         switch ($serverConfig->getAuthenticationMethod()) {
             case Configuration::AUTH_BY_PASSWORD:


elfet:
Cool! Thanks!

torinaki:
@elfet @oanhnn it will be greate to prepare new release with this fix, because this may affect many users.

elfet:
Ok! @deployphp/maintainer can anybody make new release?

oanhnn:
I created a new release. @torinaki @AlexStansfield , Please test again!

AlexStansfield:
Sorry I get:
```
PHP Notice:  Cannot connect to stage.ourserver.com:22. Error 60. Operation timed out in phar:///Users/alex/Downloads/deployer (1).phar/vendor/phpseclib/phpseclib/phpseclib/Net/SSH2.php on line 1011
```

When it tries to connect.

torinaki:
@AlexStansfield , @elfet , @oanhnn I was able to reproduce problem. Sorry guys it was my bad. Found that I was testing with patched `phpseclib` with few fixed issues by me.
Adding `3600` as default timeout is proper fix in this case. I have prepared PR #435 with fix and prepared functional tests that cover timeout case. Please check PR, because I have to change Travis configurations to achieve possibility to implement these tests. In case if you will not agree to change Travis changes I will remove tests from PR.

torinaki:
The problem was as I said before was that `fsockopen` do not create socket with huge timeout, but `3600` is ok for it.

PR #435, opened 2015-08-20
Implemented tests and fixed default timeout value to not soo huge

torinaki:
It fixes wrong huge default value for timeout that was introduced in #429
- [x] be sure that Travis tests works fine

IMPORTANT! To prepare ssh tests, I have to use `sudo: required` that means that all tests will run on old Travis infrastructure not [container-base](http://docs.travis-ci.com/user/workers/container-based-infrastructure/)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 /.idea/
 /nbproject/
 *.phar
+/.vagrant
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: php
 
+sudo: required
+
 php:
   - 5.4
   - 5.5
@@ -13,10 +15,10 @@ matrix:
     - php: hhvm
 
 install:
+  - eval `ssh-agent -s`
+  - test/travis/setup-secure-shell.sh
   - composer self-update
   - composer install --no-interaction --prefer-source
 
 script:
   - phpunit
-  
-sudo: false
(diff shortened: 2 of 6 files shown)


oanhnn:
Sorry @torinaki ,
This is an important issue, it make error for version 3.0.7 and 3.0.8.
I want merge this PR with out add tests (all tests will be reviewed by @elfet or other maintainer).
Please split this PR to two PR:
* fixed default timeout value
* adding tests
Thank you!

elfet:
Thanks for fix! :beers: 
I think @oanhnn is right, but now we need to fix it as soon as possible. 
So i merge this PR.
@torinaki is it possible to implement this tests, but on new travis dockers?

torinaki:
@elfet , @oanhnn here is PR https://github.com/deployphp/deployer/pull/438 . Manually tested both cases(w/o bug, with bug)

torinaki:
@elfet about tests implementation on new travis docker. I was searching on github for already existing examples, but unfortunatly no luck. My current implementation was based on `phpseclib` implementation. 
Currenlty i have few ideas where we should drip. One of them the most real is:
Try to install SSH server on travis docker. It should listen on 127.0.0.1(it is [possible](http://docs.travis-ci.com/user/ci-environment/#Networking)). For example we can try to install binary of openssh server: https://help.ubuntu.com/lts/serverguide/openssh-server.html .

PR #437, opened 2015-08-21
Only clear controller when environment is prod

ihsanudin:
Only clear controller when environment is prod


elfet:
What about other envs?

ihsanudin:
@elfet: in other envs we need to use app_*, profiler toolbar and many thing that we don't have in production env. So we need to make sure the front controller for other envs is existing

PR #438, opened 2015-08-21
Fixed incorrect huge default timeout value

torinaki:


torinaki:
Changes already in master

Issue #356, opened 2015-06-17
Namespace for functions


elfet:
Should Deployer use namespace for functions in `functions.php`?
To protect from overriding when installed via Composer with Laravel.
In recipe can be used in two ways:
```php
use function Deployer/task;
```
or
```php
namespace Deployer;
```
@deployphp/maintainer what do you think?

kenjis:
`use function` can be used on PHP 5.6 or lator.

gordalina:
If you use `use function` you'd have to import all functions (not only `task`). So i'd go for `namespace Deployer`.

Also, if this is to be implemented, i'd suggest in 4.0, per [semver](http://semver.org/) compliance. Lets not break API in minor updates.

ZeeCoder:
I already wrote a couple of times how I'm a bit troubled by the "global" way Deployer sometimes handles things, so I'd back any solution.
On another note: I agree with @gordalina here, the idea is good, but it's hard to see whether it could be achieved in a non-BC-breaking fashion.

gordalina:
@ZeeCoder I don't think it's achievable in a non-BC-breaking fashion, and I'm ok with that, **if and only if**, there's a proper versioning scheme (such as semver) and documentation for it.

ZeeCoder:
My thoughts exactly

elfet:
So you think go for v4.0?

ZeeCoder:
For certain features, like this one, yes imho.

elfet:
Well, lets release 4.0! Also i haму 100% awesome feature for this release. :fork_and_knife:  :fork_and_knife:   :fork_and_knife:   :fork_and_knife:

gordalina:
Is there any other way to fix deployment of laravel apps?

felipepodesta:
+1

mbardelmeijer:
:+1:

marcosh:
The global nature of deployer functions clashes also with `hamcrest-php`.

I get the following error:

    PHP Fatal error:  Cannot redeclare set() 
    (previously declared in vendor/hamcrest/hamcrest-php/hamcrest/Hamcrest.php:455)
    in vendor/deployer/deployer/src/functions.php on line 417

Definitely +1 for using a namespace in `functions.php`

diimpp:
Same here

JHGitty:
Need this as fast as possible for Symfony2 / Symfony3.

JHGitty:
I think we need #471 to complete this issue.

Issue #316, opened 2015-06-01
Laravel override get() function


Zae:
I'm getting this error when trying to run deployer.

I installed deployer ^3.0 through composer in a laravel 5 project

running PHP 5.6.8 or PHP 7.0.0-dev (did not test any other versions)

elfet:
What version of Laravel?

Zae:
Version v5.0.32 according to my composer.lock file.

elfet:
Show full `deploy.php ` file here, please.

Zae:
~~~~php
<?php

require 'recipe/laravel.php';

server('prod', 'XXXX', 22)
	->user('ezra')
	->identityFile()
	->forwardAgent() // You can use identity key, ssh config, or username/password to auth on the server.
	->stage('production')
	->env('deploy_path', '/home/ezra/apps/'); // Define the base path to deploy your project to.

// Specify the repository from which to download your project's code.
// The server needs to have git installed for this to work.
// If you're not using a forward agent, then the server has to be able to clone
// your project from this repository.
set('repository', 'XXXXXX');
~~~~

elfet:
Do your require your `autoload.php` file? You shoud not to do this.

Zae:
I do not, I installed deployer with composer into my vendors directory and called deployer using

~~~~
./vendor/bin/dep deploy production
~~~~

I guess that composer is automatically loading laravel's helper file before your helper file and because they both define a global get() function they clash.

elfet:
Yeah, seems like that. Can you use phar?

Zae:
I can and that works, but wanted to let you know of the bug anyway.
Op 1 jun. 2015 15:13 schreef "Anton Medvedev" <notifications@github.com>:

andreascreten:
Are there any plans to make Deployer L5 and Lumen compatible?

kenjis:
Don't worry. It is on going. See #356 

And you can deploy now if you use `deployer.phar`.

elfet:
@andreascreten you can implement Lumen recipe by yourself.

felipepodesta:
+1

ceesvanegmond:
We do need the deployer namespace. I really want to use composer, since it is an dependency of my project, but when using Laravel 5.1, I have to use the .phar file, which is annoying me.

This means that all Laravel users cannot use the composer dependency, but have to use the phar file.

I hope this is going to be fixed soon.

dimsav:
+1

[...]

PR #446, opened 2015-08-26
#316 Implemented namespaces in application.

ceesvanegmond:
See #356. I think namespacing the helper functions will remove alot of problems. Especially those who's using the Laravel/Lumen framework.

### DIFF
--- a/recipe/codeigniter.php
+++ b/recipe/codeigniter.php
@@ -5,18 +5,20 @@
  * file that was distributed with this source code.
  */
 
+use Deployer\Functions;
+
 require_once __DIR__ . '/common.php';
 
 // CodeIgniter shared dirs
-set('shared_dirs', ['application/cache', 'application/logs']);
+Functions\set('shared_dirs', ['application/cache', 'application/logs']);
 
 // CodeIgniter writable dirs
-set('writable_dirs', ['application/cache', 'application/logs']);
+Functions\set('writable_dirs', ['application/cache', 'application/logs']);
 
 /**
  * Main task
  */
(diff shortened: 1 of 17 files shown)


recipe/codeigniter.php
```diff
@@ -5,18 +5,20 @@
  * file that was distributed with this source code.
  */
 
+use Deployer\Functions;
```
elfet: Why not this?
```php
namespace Deployer\Functions;
```

src/functions.php
```diff
@@ -1,4 +1,7 @@
 <?php
+
+namespace Deployer\Functions;
```
elfet: Must be under file comment.

elfet:
Will add this to v4.

Baldinof:
Hi, why not putting `functions.php` under `Deployer` namespace, instead of `Deployer\Functions`.

So we'll just have to put `namespace Deployer;`  in `deploy.php` and recipes files, which is sexier than `namespace Deployer\Functions;`

It could be another cool namespace :smile:

JHGitty:
I think everything should be under ``Deployer`` namespace directly.

elfet:
Agree.

JHGitty:
ping @ceesvanegmond
Can you fix it or should we create a new PR?

elfet:
PR?

JHGitty:
PR = Pull Request.

Baldinof:
I think this PR could be closed as 5eadb57 implements namespace for functions.

elfet:
Right!

PR #447, opened 2015-08-30
Added support for setting http user for chmod

daFish:
This feature makes it possible to specify the user for `chmod` & `setfacl`. In my use case the PHP-FPM pool is using an dedicated user but Apache is served by www-data. Using `deploy:writable` all writable dirs would use the Apache user rather the pool user which causes permission issues when trying to set the file mode and owner.

You can now use the directive `set('http_user', 'username')` to set the user owning this files & directories.


oanhnn:
Thankyou!

elfet:
@daFish can you add docs, please?

daFish:
@elfet PR opened under deployphp/docs/pull/40

PR #453, opened 2015-09-07
Make clearing the controller files an option and be able to define them

jverdeyen:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -

This can be defined per stage/server when using fe `app_staging.php`.

### DIFF
--- a/recipe/symfony.php
+++ b/recipe/symfony.php
@@ -31,6 +31,10 @@ env('env', 'prod');
 set('bin_dir', 'app');
 set('var_dir', 'app');
 
+// Clear controllers
+env('clear_controllers', true);
+// Clear controllers path, relative to release_path
+env('clear_controllers_path', ['web/app_*.php', 'web/config.php']);
 
 /**
  * Create cache dir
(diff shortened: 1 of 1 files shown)


oanhnn:
@elfet @jverdeyen 
What about think a task, it using parameter `clear_paths` to delete files or directories after deploy ?
I think it is useful in many cases.
```php
set('clear_paths', []);
set('clear_use_sudo', false);

task('deploy:clean', function() {
     $paths = get('clear_paths');
     $sudo  = get('clear_use_sudo') ? 'sudo' : '';
     foreach($paths as $pash) {
          run("$sudo rm -rf {{release_path}}/$path");
     }
});
```

jverdeyen:
That's indeed doing the same but more generic.
I've been using capistrano and capifony, but I'm switching on to deployer.

:+1: @oanhnn

elfet:
PR #455

PR #392, opened 2015-07-12
Update Laravel Recipe

KennedyTedesco:
The current stable version of Laravel (5.1) doesn't use the `vendor` folder anymore (to write things). Also added the `deploy:writable` task.

PR #456, opened 2015-09-09
Updated shared dirs in laravel 5 recipe

dimsav:
This fixes a bug when laravel tries to get the full path of these directories but returns false because they don't exist.

This is caused in some cases like the [view config](https://github.com/laravel/laravel/blob/master/config/view.php#L31),  because `realpath()` is used.

If we don't create each shared folder during the deployment, realpath returns false because the dir doesn't exist and thus laravel tries to write to the root `/` directory.

- **Deployer version:** master branch
- **PHP version:** all php versions
- **Deployment target(s) OS:** hmm, all

### DIFF
--- a/recipe/laravel.php
+++ b/recipe/laravel.php
@@ -13,7 +13,7 @@ set('shared_dirs', [
     'storage/framework/cache',
     'storage/framework/sessions',
     'storage/framework/views',
-    'storage/framework/logs',
+    'storage/logs',
 ]);
 
 // Laravel 5 shared file
--- a/recipe/laravel.php
+++ b/recipe/laravel.php
@@ -8,7 +8,13 @@
 require_once __DIR__ . '/common.php';
 
 // Laravel shared dirs
-set('shared_dirs', ['storage']);
+set('shared_dirs', [
+    'storage/app',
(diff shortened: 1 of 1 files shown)


elfet:
Hmm, does work with standard Laravel?

oanhnn:
I think this fix is correct with Laravel 5.1 (LTS version).
Are you test for all laravel version? @dimsav 
Ref #392

dimsav:
Nope I tested this only for laravel 5.1.

oanhnn:
@dimsav 
I think this recipe should support laravel 5.1.x (long term support version) and we should add a note to support other version. Can you create a pull request to `deployphp/docs`?

dimsav:
After doing a check, this change enables support for both laravel 5.0 and 5.1 versions.

I could help filling the documentation but this is another totally different task :)

elfet:
Great!!! I think merge? But with docs it will be 200% more cooler!

dexbarrett:
Any ETA for this? Would be really helpful. Thanks :D

dimsav:
> Great!!! I think merge? But with docs it will be 200% more cooler!

The docs is in another git repo.

PR #461, opened 2015-09-21
Add support to deploy a sub-directory only of a git repo

tmarly:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -

Hi,

I don't know if many projects have the same concern as mine: usually, the app (the symfony app, drupal app, ...) are not at the top level of my repo. For example:
 - repo
   - myapp
   - build_tools
   - vagrant_conf
   - ...

So this pull request may interest people. You have then to add to deploy.php:
```php
set('repository_subdir', 'my/appdir');
```

Another solution is of course to create our own recipe with this code fragment (for example "deploy:git_subdir"), but you can not call it with:
```php
  after('deploy:update_code', 'deploy:git_subdir'); 
```
because some framework recipes (for ex symfony) have already registered tasks after deploy:update_code. Instead (for symfony at least) you can use:
```php
before('deploy:clear_controllers', 'deploy:git_subdir');
```

++

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -15,6 +15,7 @@ set('copy_dirs', []);
 set('writable_dirs', []);
 set('writable_use_sudo', true); // Using sudo in writable commands?
 set('http_user', null);
+set('repository_subdir', '.');
 
 /**
  * Environment vars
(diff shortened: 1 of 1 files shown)


elfet:
Please resolve conflicts.

PR #467, opened 2015-10-13
Adding new variable "cleanup_use_sudo"

Claymm:
We faced this in the past where we needed to run the cleanup command as sudo as all our releases are owned by a different user than the one which we use for deployment.


elfet:
I don't this what it is save. What do you think?

Claymm:
Sorry but what do you mean by "I don't this what it is save."?

elfet:
Do you ask password on sudo?

Claymm:
Nope.

elfet:
So every body can delete files from sudo as root?

Issue #419, opened 2015-07-31
Remove assetic from Symfony recipe?


garak:
Symfony recipe currently forces to use assetic. I think this a bit rude, since assetic is absolutely not mandatory for a Symfony project.
It would be nice to remove from main task, maybe adding a note in documentation suggesting to possibly enable 'deploy:assetic:dump' using `after` function.
A possible alternate solution is setting a boolean configuration variable, so you can disable/enable assetic in Symfony (this is how capifony does)

elfet:
I think we need to investigate how to do it best.
ping @deployphp/maintainer

elfet:
I think we need to investigate how to do it best.
ping @deployphp/maintainer

ZeeCoder:
To be honest I build by assets by gulp on every deployment on the server-side, so I don't use assetic either.
The simplest solution could be to separate the assetic part into a separate recipe, but this would of course would be a BC break.

Another solution that's a non-breaking change could be achieved using global variables, so that if something like: "sf_skip_assetic" was set to true, then the task would do nothing.
It's not the nicest solution, since the task itself would still run, pretending to build assets, but in reality it would do nothing.

elfet:
symfony.php has it’s own setting list. May be add it there?

ZeeCoder:
Sure

ihsanudin:
+1

hason:
:+1:

JHGitty:
Assetic has been removed from Symfony Standard Edition for Symfony >= 2.8 **as default**.
I think the recipe should only contain the basics of Symfony against the newest Symfony Standard Edition.

dvzzz:
Symfony 2.8.0

```
  [RuntimeException]                                                  
                                                                      
    [Symfony\Component\Console\Exception\CommandNotFoundException]    
    There are no commands defined in the "assetic" namespace.         
    Did you mean this?                                                
        assets
```

garak:
Maybe you closed it too fast? There is no mention about the new option added in #499 in docs.

elfet:
Agree. Need to update docs.

elfet:
Lets turn off assetic by default.

PR #499, opened 2015-12-18
Flag for assetic:dump task #419

piotrantosik:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #419

Default is set to true - current configuration does not need to be changed.
If you do not use AsseticBundle just switch flag to false:

```php
set('dump_assets', true);
```


oanhnn:
Thank you!

PR #500, opened 2015-12-18
Remove php 7 from failures

piotrantosik:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -
Php 7 is stable so it should be removed from failures


oanhnn:
Thank you!

PR #504, opened 2015-12-23
Make more correct RegExp in RemoteServerTest

torinaki:
Problem with tests has been appeared because Travis has changed IPv6 to IPv4 (see
"IPv6 no longer present" section in https://blog.travis-ci.com/2015-11-27-moving-to-a-more-elastic-future)


oanhnn:
Thank you!

PR #511, opened 2016-01-05
Added variable for composer command path

zupolgec:
Now we can specify a path for the composer command, avoiding the download

Q | A
------------ | -------------
Bug fix? | no
New feature? | yes
BC breaks? | no
Deprecations? | no
Fixed tickets | -

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -22,6 +22,7 @@ set('http_user', null);
 env('timezone', 'UTC');
 env('branch', ''); // Branch to deploy.
 env('env_vars', ''); // For Composer installation. Like SYMFONY_ENV=prod
+env('composer_command', 'composer'); // Path to composer
 env('composer_options', 'install --no-dev --verbose --prefer-dist --optimize-autoloader --no-progress --no-interaction');
 env('git_cache', function () { //whether to use git cache - faster cloning by borrowing objects from existing clones.
     $gitVersion = run('git version');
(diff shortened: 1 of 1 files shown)


oanhnn:
why are we need this feature?

zupolgec:
Because sometimes we have composer on the server, but cannot be reached with a simple `composer`. On a server I use for deploy the command is `composer.phar`. With this setting we can avoid downloading composer

oanhnn:
Yes. Can you add a PR to [document](https://github.com/deployphp/docs) about this feature ?

zupolgec:
Docs added!

elfet:
But you can always add it. In .bashrc:
```
alias composer='/bin/composer'
```

zupolgec:
Ok, seems fair :+1:

zupolgec:
@elfet I checked and I already had that line in my .bashrc
The problem lies in the commandExist function. The `hash` command doesn't "follow" aliases. Maybe you should change that to `command`.
http://stackoverflow.com/questions/592620/check-if-a-program-exists-from-a-bash-script

lbarulski:
We keep `composer.phar` in project repository to keep its version under control - there were some situations when phar file on getcomposer was broken or installation based on the same `composer.lock` gave us different results on different `composer.phar` versions (dev vs prod)...

:+1: for feature

oanhnn:
Thank you!

elfet:
Awesome!

mbrodala:
Another usecase: sometimes there are various PHP versions available for CLI and you want to ensure the correct one is used. With this change you can simply prepend it, thus :+1:

PR #515, opened 2016-01-07
Add Symfony3 recipe

### LINKED ISSUE #514: symfony 3 recipe support
Seems like the symfony recipe doesn't support the symfony 3 directory structure yet. 
* console command moved to bin folder
* cache & logs moved to var folder. 

I guess the recipe just needs to be tweaked some more to accommodate these

pierredup:
Fixes #514

### DIFF
--- a/recipe/symfony3.php
+++ b/recipe/symfony3.php
@@ -1,0 +1,21 @@
+<?php
+/* (c) Anton Medvedev <anton@elfet.ru>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+require_once __DIR__ . '/symfony.php';
+
+/**
+ * Symfony 3 Configuration
+ */
+
+// Symfony shared dirs
+set('shared_dirs', ['var/logs']);
+
+// Symfony writable dirs
+set('writable_dirs', ['var/cache', 'var/logs']);
+
+set('bin_dir', 'bin');
+set('var_dir', 'var');


oanhnn:
Please eidt message like

Q | A
------------ | -------------
Bug fix? | no
New feature? | yes
BC breaks? | no
Deprecations? | no
Fixed tickets | -

elfet:
It merge, we will have two recipe:
* symfony.php
* symfony3.php
Maybe it's better to add configuration for Symfony 3 in `symfony.php` recipe?
```php
set('symfony_version', 3);
```
Like this?

pierredup:
The only issue I see with this is that it will clutter the recipe with code like this

```
if (get('symfony_version') === 3) {
    set('var_dir', 'var');
} else {
    set('var_dir', 'app');
}
```

But a config option might be a better option, since it can then be used in other areas as well (E.G a user can set custom tasks based on the Symfony version)

If you are happy with this approach, then I will update the PR

elfet:
I think we need to discus it before. Support Symfony 3 very important. What @deployphp/maintainer thinks? @oanhnn ?

oanhnn:
We have 2 ways:
1. Only a recipe for framework and using `set('symfony_version', 3.0)` or `set('laravel_version', 5.1)` to switcth configure options. Default version variable is LTS version or last stable version. With this way, recipe have to cover more cases.
2. Using framework+version to save recipe (Eg: `symfony_2`, `symfony_3`, `laravel_5.1`). With this way, recipe is simple but number of recipe is more (Eg: `laravel_4`, `laravel_5.0`, `laravel_5.1`, ...)

I think we can use configure resolver by version of framework.

elfet:
What to do if framework very different? Like symfony 1 and 2? I think better it's to have both options) Configs for same, separate recipe for diff.

oanhnn:
>What to do if framework very different? Like symfony 1 and 2?

Two version are very different but i think deployment are same, only different about: shared_dir, shared_file, wriable_dirs and migration command.
Note: old version of frameworks (not support composer) are very different. Should we support them?

elfet:
Very old we do not support for now %)

I think lets merge `symfony3.php`, as no more code needed to be written to support it. :) But docs really needed.

elfet:
Now we have it! Can anybody update docs?

pierredup:
Thanks @elfet. I have updated the docs here deployphp/docs#48

PR #516, opened 2016-01-07
Adding useful tasks to put application in maintenance mode

zupolgec:
Q | A
------------ | -------------
Bug fix? | no
New feature? | yes
BC breaks? | no
Deprecations? | no
Fixed tickets | -

Laravel script `artisan` has a lots of useful commands. Among them, there are the `down` and `up` commands.

Maybe we should create a complete wrapper to run artisan commands from deployer (like migration and such).


oanhnn:
Thank you!

PR #528, opened 2016-01-14
Fix environment variable replacement in "download" function

x3ro:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | none, afaik

I've also modified the test for `upload` so that it verifies environment variable replacement.


elfet:
Cool!

Issue #527, opened 2016-01-14
Laravel recipe shouldn't have shared file .env


SanderSander:
In the laravel recipe the `.env` file is defined as shared file on any environment and therefor is always there while it shouldn't! Dotenv loads the environment variables from `.env` into $_ENV, $_SERVER but i guess the most of the time the .env is just a empty file and hopefully not used on production environments

    // Laravel 5 shared file
    set('shared_files', ['.env']);

Also from the documentation Usage notes of dotenv  
https://github.com/vlucas/phpdotenv#usage-notes

`phpdotenv is made for development environments, and generally should not be used in production. In production, the actual environment variables should be set so that there is no overhead of loading the .env file on each request. This can be achieved via an automated deployment process with tools like Vagrant, chef, or Puppet, or can be set manually with cloud hosts like Pagodabox and Heroku.`

elfet:
Can you make PR for it?

SanderSander:
Of course, will do it somewhere today

SanderSander:
Ok PR is there. 

I noticed that there was a `test/recipe/LaravelTest.php` but couldn't find it anywhere in the source on the master branch, so I assume it was removed.

https://github.com/deployphp/deployer/commit/6b69e8f3e43a4ba67e17e07205527c656a63b612

pelletiermaxime:
I disagree with this. I think in most environments (except prod) it's helpful to automatically have a shared .env. In production you should just not create the .env file, so it won't be shared.

SanderSander:
I rather explicit define on which environments I want to use dotenv, than doing it implicit. 

For testing you can use phpunit.xml to set the environment variables

oanhnn:
In a lot of cases, you can't modified environments (sharing host). `.env` is solution.
If you don't want use it, you can using an empty `.env` file.

SanderSander:
You're right on the shared hosting part, I was actually a bit surprised to see .env files on the servers because I only use it for development environments all private configuration on staging/production is done in the environment itself, and I'm not really used anymore to shared hosting to be honest. 

It's simply solved by adding `set('shared_files', []);` to the deployment script. 

The only reason I don't want the .env file is because I don't want others to think they can use that .env file to fix some mis configuration on staging or whatever it's cleaner to don't have it when not used.

badconker:
How you can do ``php artisan migrate`` without .env file ?

badconker:
[...]

PR #530, opened 2016-01-15
#527 Remove shared file .env on laravel recipe

SanderSander:
#527 Remove shared file .env on laravel recipe

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes?
| Deprecations? | no
| Fixed tickets | #527

It will break deployments that use .env for configuration, but this is really discouraged by dotenv

### DIFF
--- a/recipe/laravel.php
+++ b/recipe/laravel.php
@@ -17,9 +17,6 @@ set('shared_dirs', [
     'storage/logs',
 ]);
 
-// Laravel 5 shared file
-set('shared_files', ['.env']);
-
 // Laravel writable dirs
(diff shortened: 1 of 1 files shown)


elfet:
@oanhnn you add `.env.`. Is it does not needed?

oanhnn:
`.env` is configure file in laravel 5.0+. It like `parameters.yml` in symfony 2.
I think it should be shared between deployed versions.

oanhnn:
In a lot of cases, you can't modified environments (sharing host). .env is solution.
If you don't want use it, you can using an empty .env file.

elfet:
So leave `.env`.

PR #535, opened 2016-01-22
Added way to execute tasks only for certain stages

michaelmoussa:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | N/A

I have a use case where my `servers.yml` file is not a static list of servers. I have more than one production webserver, and the server information is pulled down from AWS and written to `servers.yml` each time I do a deployment. The number of production servers could increase or decrease based on needs, so I don't want to have to update the deployment script every time that happens.

This means I can't just do:

```php
task('task_name', function () {
})
    ->onlyOn('production1', 'production2', 'etc...');
```

However, I know that I always want to run that task on Production, regardless of how many Production servers I have. That is the purpose of this pull request. Instead, I can do:

```php
task('task_name', function () {
})
    ->onlyFor('production');
```

Then, no matter how many production servers I have, or what they are named, this task will only execute for production.

Some things to note:
1. If a task has `->onlyFor(...)` set, then it will _not run_ if the current server environment does not have a stage set.
2. You can specify multiple stages using `->onlyFor('production', 'staging')` for tasks that can run in more than one stage, but not necessarily all of them.

### DIFF
--- a/src/Executor/ParallelExecutor.php
+++ b/src/Executor/ParallelExecutor.php
@@ -9,6 +9,7 @@ namespace Deployer\Executor;
 
 use Deployer\Console\Output\OutputWatcher;
 use Deployer\Console\Output\VerbosityString;
+use Deployer\Server\Environment;
 use Deployer\Task\Context;
 use Pure\Server;
 use Pure\Storage\ArrayStorage;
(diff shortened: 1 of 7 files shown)


elfet:
Hi! This is good feature. 
But maybe it's better to rename it in more telling and unambiguous, like `onlyForStage`?

jacmoe:
`onlyForStage` is a good idea :+1:

michaelmoussa:
Good idea @elfet. I've made the change and updated the PR. Any chance of a release soon with this change? We try to use only tagged libraries / tools when possible. :)

Thanks!

elfet:
Done! @oanhnn can you care a new release, please?

PR #536, opened 2016-01-23
[master] Ignore composer.lock

lucasmichot:
Ignore composer.lock


jacmoe:
The lock is in the repository because Deployer is an application and should be locked to specific versions of it's dependencies.

jacmoe:
Composer.lock was added on February 2015 - check the history:
https://github.com/deployphp/deployer/commits/master/composer.lock

lucasmichot:
This could apply to full project alongside with packages.json, bower.json and son on
In a package (even a binary one) like deployer, the constraints should be stricter and defined in `composer.json`

lucasmichot:
If a dependencies package is buggy, we stay locked on it...

jacmoe:
OK, if that is the intention, then your PR does not update composer.json accordingly :)

lucasmichot:
?
> If a dependencies package is buggy, we stay locked on it...

lucasmichot:
And I suppose unit tests are written on purpose, are they not?

jacmoe:
> ?

People are building Deployer *at home* - it is the third option in the *installing Deployer* doc. How does removing *composer.lock* without modifying *composer.json* help?

lucasmichot:
See phpunit / php-cs-fixer
- https://github.com/sebastianbergmann/phpunit/blob/master/composer.json  
- https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.12/composer.json (even less strict version) 

For both of these projects, people can build them and they do it very fine without any composer.lock

jacmoe:
Those are libraries..

Composer has a composer.lock in their repository..

Anyway, let's wait for the Deployer project maintainers :)

elfet:
Why composer.lock  disturbs you?

lucasmichot:
It does not

PR #537, opened 2016-01-26
Update composer it self before updating vendors

ihsanudin:
Update composer it self before updating vendors

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -299,6 +299,7 @@ task('deploy:writable', function () {
 task('deploy:vendors', function () {
     if (commandExist('composer')) {
         $composer = 'composer';
+        run("$composer self-update");
     } else {
         run("cd {{release_path}} && curl -sS https://getcomposer.org/installer | php");
         $composer = 'php composer.phar';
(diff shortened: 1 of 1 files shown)


elfet:
Are you sure to do it automatically? Maybe add some options?

ihsanudin:
Yes, I think automatic updates whenever deploy is better. But config/options is not bad

ihsanudin:
@elfet : this PR is same as download new composer.phar but it is downloading update

oanhnn:
@ihsanudin In a lot of cases, users installed `composer` global with root user ( using sudo ), this PR will be error. I think the update `composer` is customized by individual.

ihsanudin:
@oanhnn : you are right, ok may be better to check if use sudo, try to update `composer` automatically

elfet:
`sudo` isn’t solution.

oanhnn:
@ihsanudin Using `sudo` to update `composer` and make writable dirs are two problem different. You can't using same config. I think **the updating composer task have do manually**.
@elfet I think this PR should be closed.

elfet:
Agree.

Issue #404, opened 2015-07-27
More secure permission granting on writable dirs


AlexStansfield:
At the moment I find two issues with the current implementation of setting permissions on writable directories:
- Can't set a particular method to use
- Automatically fall back to ```chmod 777``` if ```chmod +a``` and ```setfacl``` fail

The first I think would be pretty useful to have (it was in 2.x) and the second could lead to people giving read/write permissions to all users without realising.

I propose this change:

```
    $httpUser = run("ps aux | grep -E '[a]pache|[h]ttpd|[_]www|[w]ww-data|[n]ginx' | grep -v root | head -1 | cut -d\  -f1")->toString();

    cd('{{release_path}}');

    $funcPermChmod = function($httpUser, $dirs, $sudo) {
        if (!empty($httpUser)) {
            run("$sudo chmod +a \"$httpUser allow delete,write,append,file_inherit,directory_inherit\" $dirs");
        }

        run("$sudo chmod +a \"`whoami` allow delete,write,append,file_inherit,directory_inherit\" $dirs");
    };

    $funcPermAcl = function($httpUser, $dirs, $sudo) {
        if (!empty($httpUser)) {
            run("$sudo setfacl -R -m u:\"$httpUser\":rwX -m u:`whoami`:rwX $dirs");
            run("$sudo setfacl -dR -m u:\"$httpUser\":rwX -m u:`whoami`:rwX $dirs");
        }
    };

    $funcPermChmodBad = function($dirs, $sudo) {
        run("$sudo chmod 777 $dirs");
    };

    // Use chosen permission method if it exists
    if (has('permission_method')) {
        switch (get('permission_method')) {
            case 'chmod':
                $funcPermChmod($httpUser, $dirs, $sudo);
                break;
            case 'chmod_777':
                $funcPermChmodBad($dirs, $sudo);
                break;
            case 'setfacl':
                $funcPermAcl($httpUser, $dirs, $sudo);
                break;
            default:
                $formatter = \Deployer\Deployer::get()->getHelper('formatter');

                $errorMessage = [
                    "Unable to setup correct permissions for writable dirs.                  ",
                    "Chosen permission_method needs to be chmod, chmod_777 or setfacl        "
                ];
                write($formatter->formatBlock($errorMessage, 'error', true));
                break;
        }
    } else {
        // No chosen permission method, try chmod then acl
        if (strpos(run("chmod 2>&1; true"), '+a') !== false) {
[...]

PR #540, opened 2016-01-28
Add writable_setgid_group option to enable sticky gid writable dirs

nosnickid:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | n/a

The new option 'writable_setgid_group' allows deployments that use a common group between web server user and project user, using the existing deploy:writable task.

It's similar to https://github.com/deployphp/deployer/issues/404#issuecomment-126986068, but it also works with default recipes.

Test included.


oanhnn:
What's different?
```
$ setfacl -R -m u:"www-data":rwX -m u:`whoami`:rwX tmp
$ setfacl -dR -m u:"www-data":rwX -m u:`whoami`:rwX tmp
```
Vs
```
$ setfacl -R -m u:"www-data":rwX g:deployers:rwX u:`whoami`:rwX tmp
$ setfacl -dR -m u:"www-data":rwX  g:deployers:rwX u:`whoami`:rwX tmp
```

We need document and more test cases.
```
$hasfacl = run("getfacl -p $dir | grep \"^user:$httpUser:.*w\" | wc -l")->toString();
```

elfet:
May be add fallback to this:
```php
run("chgrp {{common_group}} $dirs");
run("chmod 775 $dirs");
```

nosnickid:
I do intend to get back to this update and get the docs updated, been busy with other things.

0775 would be marginally better but my 2775 is better still as an option, once I get around to documenting it.

elfet:
Cool!

nosnickid:
I've been looking at this.  I've taken the code in the first comment for #404 and added my extra setgid method.

I'd like to write a test that confirms each mode runs the expected command.  I don't think that I can write a reliable test that tries to examine the outcome of each of the commands because they depend on things such as a group the user is in, or an OSX command in a Linux environment.

I don't see a simple way to set an expected flag. I've passed ['verbosity' => 999] options to `$this- >exec` in the test and captured the output, so I could scan the output for Run: lines.  Eugh.

Any better suggestions?

elfet:
Documentation will be very helpful.

nosnickid:
I've made a start on the docs:

https://github.com/nosnickid/docs/blob/master/recipes.md#deploywritable_dirs

elfet:
Hi,

Since in this PR no progress, i'm closing it.

PR #541, opened 2016-01-28
Adding artisan migration helpers

filpgame:
Adding Laravel's `artisan migration` helpers

Q | A
------------ | -------------
Bug fix? | no
New feature? | yes
BC breaks? | no
Deprecations? | no
Fixed tickets | n/a

### DIFF
--- a/recipe/laravel.php
+++ b/recipe/laravel.php
@@ -41,6 +41,18 @@ after('deploy', 'success');
 /**
  * Helper tasks
  */
+task('deploy:migrate', function () {
+    cd('{{release_path}}');
+    $output = run('php artisan migrate --force');
+    writeln('<info>' . $output . '</info>');
+})->desc('Execute artisan migrate');
+
+task('deploy:migrate_rollback', function () {
+    cd('{{release_path}}');
+    $output = run('php artisan migrate:rollback --force');
+    writeln('<info>' . $output . '</info>');
+})->desc('Execute artisan migrate:rollback');
+
 task('deploy:up', function () {
     $output = run('php {{deploy_path}}/current/artisan up');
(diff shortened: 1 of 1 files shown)


oanhnn:
Please edit comment and prepend table 

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | n/a

Migartion should be executed without option `--force`

elfet:
Agree, no `--force`.

filpgame:
But in `production` environment, the migration command will be ignored without `--force`

elfet:
Is `--force` used only for asks for a confirmation? May be there is an not-interactive mode?

philbates35:
Pretty sure all `--force` does is to automatically proceed when it asks for confirmation. @flipgame is right, without it the command would be cancelled.

barryvdh:
Agreed, you need `--force`.

Perhaps also add `php artisan db:seed` and a `php artisan migrate:status` to check before running migrate?

Not sure if it's possible to do the confirm on the Deployer side, to be safe?

elfet:
It's possible, but what happens if you run `php artisan migrate --force` if there no migrations?

philbates35:
It just says "nothing to migrate". Force is purely just to get past the "you're in production, are you you sure you want to do this?" prompt.

elfet:
Do we need `deploy:migrate` in `deploy` task?

barryvdh:
> Do we need deploy:migrate in deploy task?
I wouldn't do that by default.

oanhnn:
I think this problem will be resolved in #622 
Thank you @filpgame about your contribution.

PR #556, opened 2016-02-14
Compatibility with Symfony3

vpx:


oanhnn:
You should using file deployer.phar and shouldn't require deployer in composer.json.

elfet:
@oanhnn a lot of people install it via composer. I think it's good to have both ways.

composer.json
```diff
@@ -23,12 +23,12 @@
         "bin/dep"
     ],
     "require": {
-        "php": ">=5.4.0",
+        "php": ">=5.5.9",
```
elfet: 5.4.0 still supported.

composer.json
```diff
@@ -23,12 +23,12 @@
...
         "elfet/pure": "~1.1",
```
elfet: pure need to support it too.

cedricziel:
This pull request is obsolete since i cherry-picked the relevant commit which was merged. Thank you @vpx !

PR #559, opened 2016-02-16
fix #558: check releases_list to contain only releases

### LINKED ISSUE #558: releases_list items are not checked to be a release
If any other files/directories appear in deploy_path/releases/ dir, deployer thinks it is a release without any check. It leads to problems with keep_releases: instead of releases deployer keeps other entries.

Live example:

```sh
deploy/releases$ ls -1
20160216103038
20160216105529
20160216105529.bak
20160216110130
runtime
```

`runtime` is a garbage from slightly misconfigured yii. `20160216105529.bak` is a manually created backup.

`dep cleanup` in this sutuation (assuming that keep_releases is 3) would remove last 2 releases, leaving only latest one along with `runtime` and `20160216105529.bak`:

```sh
20160216105529.bak
20160216110130
runtime
```

Thus current behaviour may lead to unwanted removal of releases. It can even remove all releases including current.

cronfy:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #558

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -324,18 +324,19 @@ task('deploy:symlink', function () {
  * Return list of releases on server.
  */
 env('releases_list', function () {
-    $list = run('ls {{deploy_path}}/releases')->toArray();
+    // find will list only dirs in releases/
+    $list = run('find {{deploy_path}}/releases -maxdepth 1 -mindepth 1 -type d')->toArray();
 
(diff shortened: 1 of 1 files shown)


elfet:
:+1: Merged!

Issue #575, opened 2016-03-07
Symfony-cache is not writable if deployer runs on another user than the httpd 


nemoinho:
I installed and configured deployer on a clean new installation and use different users for deployment and httpd.

So the directories are owned by my deployment-user. Due to the recipe of symfony and symfony3 the cache should be writable for users of the same group but after the cache warmup the prod-directory looks like this (php_deployer is my deployment-user www-data is also in group php_deployer):

    drwxrwxrwx 3 php_deployer php_deployer 4096 Mar  8 00:11 .
    drwxr-xr-x 4 php_deployer php_deployer 4096 Mar  8 00:11 ..
    drwxr-xr-x 6 php_deployer php_deployer 4096 Mar  8 00:11 prod

But to run symfony it has to be 

    drwxrwxrwx 3 php_deployer php_deployer 4096 Mar  8 00:11 .
    drwxr-xr-x 4 php_deployer php_deployer 4096 Mar  8 00:11 ..
    drwxrwxr-x 6 php_deployer php_deployer 4096 Mar  8 00:11 prod

I created a workaround for this that looks like this:
    
    /**
     * Make cache writable again after warmup
     */
    task('deploy:make_cache_writable', function () {
        run("chmod -R g+w {{cache_dir}}");
    })->desc('Make cache writable again');
    
    after('deploy:cache:warmup', 'deploy:make_cache_writable');

oanhnn:
:+1:  
This is great way to resolve this problem. But deploy user not always has group www-data.
I think it is a special case, you should write a blog or Wiki. We will insert link in document of Deployer.
What are you think? @elfet

elfet:
Yes! I really want to create better docs, i think this is 50% success of open source libs.
I think about section with hot-to or faq.

nemoinho:
I'm totally fine with that, but I have one appeal left. If you do so you should remove the lines [49-51](https://github.com/deployphp/deployer/blob/master/recipe/symfony.php#L49-L51) from the symfony-recipe because they are just confusing and have not really an effect.

oanhnn:
Thank @nemoinho , I will check it.

nemoinho:
After some research I found another possible reason for this behaviour. The deploy-task of the symfony-recipe has the wrong order in my opinion:
    
    /**
     * Main task
     */
    task('deploy', [
        'deploy:prepare',
        'deploy:release',
        'deploy:update_code',
        'deploy:create_cache_dir',
        'deploy:shared',
[...]

PR #576, opened 2016-03-08
Fix cache access-rights

nemoinho:
Bugfix for the mentioned issue #575 

Now the access-rights will be set after the cache-warmup and for more flexibility it is transferred into a task.


elfet:
May be better to run 'deploy:writable' again?

nemoinho:
Yep that's better and much more general, so I'll close this request and you can insert it or I create another one.

PR #577, opened 2016-03-08
Add comments to describe access-right settings

nemoinho:
Add some comments to make clear why the checks of the access-rights are
defined in such a complex structure.


elfet:
Cool! I also thinking about get rid of 777. May be replace it with something not so horrible?

Issue #555, opened 2016-02-11
write() weird output on short strings


cronfy:
When write()/writeln() is used inside a task to print short string (1 or 2 chars), it gives weird output. Consider this `deploy.php`:

### example 

```php
<?php

localServer('test');

task('test', function() {
        writeln("b");
});
```

`dep test` would output:

```
$ dep test
➤ Executing task test
b[k
✔ Ok
```

Note these symbols after ```b```.

### example with use case

Use case is when you want to print some message in a task, separated from other output with newlines:

```php
task('test', function() {
        writeln("\n\n\nHello\n\n\n");
});
```

It prints some characters at the beginning:

```
$ dep test
➤ Executing task test
[k


Hello



✔ Ok
```

### info

```
$ dep -V
Deployer version master
```

I tried it in 2 environments:

1. FreeBSD 8.4, PHP 5.4.16
2. Linux Mint 17, PHP 5.5.9-1ubuntu4.14

elfet:
Very strange bug. Try to run with -vvv option.

cronfy:
with -vvv it's ok:

```
$ dep test -vvv
➤ Executing task test
⤷ on [test]
b
⤶ done on [test]
✔ Ok
```

PR #580, opened 2016-03-11
Fix bug #555 and clean output

nemoinho:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #555

The bug #555 was caused by using a wrong escape-sequence for the
terminal.  Instead of using \e[K it has used \e[k which actually has no
meaning and only bloats some noise on the terminal (except on OSX where
the noise is catched and hidden from the user). To see this behaviour
you need a real linux client for example a VM and you have to use it's
own terminal – not the OSX-terminal via ssh or something like that. Than
type the following example to see the behavior:

    echo -e "aaaaaaaaa\r\033[k"
    echo -e "bbbbbbbbb\r\033[K"

Furthermore the cursor is now set to the beginning of the line first to
really clean the line and not just something starting from a random
position.

### DIFF
--- a/src/Console/Output/OutputWatcher.php
+++ b/src/Console/Output/OutputWatcher.php
@@ -35,14 +35,6 @@ class OutputWatcher implements OutputInterface
      */
     public function write($messages, $newline = false, $type = self::OUTPUT_NORMAL)
     {
-        // Next code prints arrow on task line if some output was inside task.
-        // This is ugly hack, and this part should be refactored later, but now i go segmentation fault.
-        static $isFirstTime = true;
-        if (!$this->wasWritten && !$isFirstTime) {
-            $this->output->write("\033[k\033[1A\r➤\n", false, $type);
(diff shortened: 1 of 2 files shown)


elfet:
Cool, is output is same?

nemoinho:
Output is the same, but I made a bad mistake and pushed the wrong pranch as fix, this change is fixes also a bug but the actual bug-fix is done by my other pull-request https://github.com/deployphp/deployer/pull/581 therefore I will close the other pull-request, cherry-pick the other into this one and than we have two bugs fixed, removed this ugly workaround (hack?) and have a more consistent output.

elfet:
Ok, pls inform me then you done.

nemoinho:
I'm done

elfet:
Cool!

PR #581, opened 2016-03-11
Clean ugly hack in output of an executing task

nemoinho:
Removed ugly hack in OutputWatcher. No the arrow will be printed by beginning of the task and be overwritten with an check mark. I think this is ok, because it only speed up the old implementation with lees pain.

Old:
1. set space or arrow, depending on verbositiy
2. perform task
3. set arrow or check mark, depending of written lines

New:
1. set arrow
2. perform task
3. set check mark if no new lines were written

The result is actually the same.

### DIFF
--- a/src/Console/Output/OutputWatcher.php
+++ b/src/Console/Output/OutputWatcher.php
@@ -35,14 +35,6 @@ class OutputWatcher implements OutputInterface
      */
     public function write($messages, $newline = false, $type = self::OUTPUT_NORMAL)
     {
-        // Next code prints arrow on task line if some output was inside task.
-        // This is ugly hack, and this part should be refactored later, but now i go segmentation fault.
-        static $isFirstTime = true;
-        if (!$this->wasWritten && !$isFirstTime) {
-            $this->output->write("\033[k\033[1A\r➤\n", false, $type);
-        }
-        $isFirstTime = false;
-
         $this->wasWritten = true;
         $this->output->write($messages, $newline, $type);
     }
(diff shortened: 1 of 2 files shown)


elfet:
Now it will be printing arrow instead of space?

nemoinho:
I adopted this change to the other pull-request https://github.com/deployphp/deployer/pull/580
But yes it will now print this arrow at the beginning of the line, because here is the output of the normal output of the old implementation:

      Executing task A
    \033[k\033[1A^M➤
    A✔ Ok
      Executing task B
    \033[k\033[1A^M➤
    \033[k\033[1A^M✔

It prints `Executing …` than it moves the cursor up to the beginning of the line and overwrites the first char (the space) with an arrow and that happens once before the next output will be written.
At the end it will do the same trick if no other output was written.
The same happens in the verbose output with one exception, it already prints an arrow as first character:

    ➤ Executing task A
    \033[k\033[1A^M➤
    ⤷ on [localhost]
    A⤶ done on [localhost]
    ✔ Ok
    ➤ Executing task B
    \033[k\033[1A^M➤
    ⤷ on [localhost]
    ⤶ done on [localhost]
    ✔ Ok

And this also happens in some error-cases because of the ugly workaround.

So my change will always print an arrow at the beginning because it will do so anyway and removes the ugly extra line though.
So the output looks like this in the normal output:

    ➤ Executing task A
    A✔ Ok
    ➤ Executing task B
    \033[k\033[1A^M✔

And like this for the verbose one:

    ➤ Executing task A
    ⤷ on [localhost]
    A⤶ done on [localhost]
    ✔ Ok
    ➤ Executing task B
    ⤷ on [localhost]
    ⤶ done on [localhost]
    ✔ Ok

I think that is much cleaner, actually in the verbose version shouldn't stay any cursor movements!
I used this recipe to test this:

    <?php
    task('A', function(){write('A');});
    task('B', function(){});
    
    task('AB', ['A', 'B']);

PR #584, opened 2016-03-11
Added constant DEPLOYER for other autoload scripts to check

oisvidi:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

Using this to deploy Joomla. Have a few autoloading scripts for Joomla only checking if JEXEC is defined. This allows my autoloads to return if DEPLOYER is defined.

### DIFF
--- a/bin/dep
+++ b/bin/dep
@@ -6,6 +6,8 @@
  * file that was distributed with this source code.
  */
 
+define("DEPLOYER", 1);
+
 $loaded = false;
 
 foreach ([__DIR__ . '/../../../autoload.php', __DIR__ . '/../vendor/autoload.php'] as $file) {
--- a/bin/dep
+++ b/bin/dep
@@ -6,7 +6,7 @@
  * file that was distributed with this source code.
  */
 
-define("DEPLOYER", 1);
+define('DEPLOYER', true);
 
 $loaded = false;
 


elfet:
That is this about? What you trying to check?

oisvidi:
@elfet Im not suggesting that this library use this constant internally. It is for projects and other libraries where deployer is included as require-dev and has a lot of autoloading libraries. 

My Joomla config library (which sets up a DIC) has a check for a few constants or die since it requires the Joomla Application and Joomla classes/functions for various stuff. I can now add this line at the start of the config library: if (defined("DEPLOYER")) return;

elfet:
You mean if you call it as `dep` some autoloaded staff will run?

oisvidi:
@elfet Yes. The vendor/bin/dep loads composer autoloader, which autoloads some library files. With this constant other autoloading libraries can (if they want to) know deployer is the application running.

elfet:
Ok, but please fix some code styles.

bin/dep
```diff
@@ -6,6 +6,8 @@
  * file that was distributed with this source code.
  */
 
+define("DEPLOYER", 1);
```
elfet: `define('DEPLOYER', true);`

oanhnn:
@oisvidi Why not use `DEPLOYER_BIN` ?

oisvidi:
@oanhnn That can work if we move the declaration of DEPLOYER_BIN above the foreach loading autoloaders.

oanhnn:
@oisvidi If using constant `DEPLOYER` always is TRUE, it mean deployer alway loaded success?
I think deployer only is loaded success when constant `DEPLOYER_BIN` is defined.

oisvidi:
@oanhnn it just means DEPLOYER is the application running. The value is irrelevant for that purpose.

elfet:
I think it's ok to have a set on constants `DEPLOYEК, DEPLOYER_*`.

oanhnn:
> it just means DEPLOYER is the application running. The value is irrelevant for that purpose.

If not loaded `autoload.php`, DEPLOYER = true but DEPLOYER_BIN is undefined and application not running. So that, what is you want check?

elfet:
I think some legacy code may load at this point:
```php
require $file;
```

oanhnn:
> I think some legacy code may load at this point:
> ```
> require $file;
> ```

Is your mean?
```
if (!defined('DEPLOYER')) {
    require '/path/to/vendor/bin/dep';
}
```
If that, why not using `require_one` ?

elfet:
I think its when composer load files with PSR-0.

oisvidi:
Composer has an autoload index which loads the specified files at all times. 

deployer is in the project require-dev so it's not installed on the server but it is installed on my machine (and others when they'll work on this project again).

PR #585, opened 2016-03-11
Allows people to use other release names like tag without replacing t…

oisvidi:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

Allow people to name release folders (to tag) without replacing task deploy:release .

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -34,6 +34,7 @@ env('git_cache', function () { //whether to use git cache - faster cloning by bo
     }
     return version_compare($version, '2.3', '>=');
 });
+env('release_name', date('YmdHis')); // name of folder in releases
 
 /**
  * Default arguments and options.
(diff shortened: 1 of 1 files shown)


elfet:
This is very useful feature, but with this place a lot of problems (like different timezones between developers), to it's must be refactored at all.

oisvidi:
@elfet This commit simply moved the $release = date("YmdHis"); out of the task to an env() key. 

If your developers have different timezones this change allows you to make a new env value which uses timezone. 

Or like me to use the git tag as release.

elfet:
Ok, i think it's ok. What do you think @oanhnn ?

oanhnn:
:+1:  I agree merge this feature. It is OK.

Issue #526, opened 2016-01-14
Add task duration to the output


whyte624:
It would be great to see how long task was running.
I suppose \Deployer\Executor\Informer::endTask can output some more data.

elfet:
Cool idea for -vvv option!

whyte624:
What is the reason of closing this proposal?

oanhnn:
I think comment of elfet is answer for you.

whyte624:
Is it implemented already? It seems like no.

oanhnn:
Yes, It is implemented

whyte624:
> Deployer version 3.0.10

And I run deployer with -vvv
Here is the output partially

```
➤ Executing task database:schema:update
⤷ on [prod]
Run: cd /www/releases/20160217075037 && php ./bin/console doctrine:schema:update --env=prod --force
> No Metadata Classes to process.
> 
⤶ done on [prod]
✔ Ok
➤ Executing task deploy:symlink
⤷ on [prod]
Run: cd /www && ln -sfn /www/releases/20160217075037 current
Run: cd /www && rm release
⤶ done on [prod]
✔ Ok
```

I see no information on the task duration. Am I doing something wrong?

oanhnn:
Uhm, Do you mean like `gulp`, output show duration of task?
```
$ gulp
[09:10:06] Using gulpfile /var/projects/oanhnn/slim-skeleton/gulpfile.js
[09:10:06] Starting 'build'...
[09:10:06] Starting 'clean'...
[09:10:07] Finished 'clean' after 813 ms
[09:10:07] Starting 'jshint'...
[09:10:10] Finished 'jshint' after 2.44 s
[09:10:10] Starting 'create:dir'...
[09:10:10] Finished 'create:dir' after 553 μs
[09:10:10] Starting 'build:sass'...
[09:10:13] Starting 'build:css'...
[09:10:13] Starting 'build:js'...
[09:10:14] Finished 'build:js' after 819 ms
[09:10:14] Finished 'build:css' after 824 ms
[09:10:14] Finished 'build:sass' after 4.51 s
[09:10:14] Finished 'build' after 7.77 s
[09:10:14] Starting 'default'...
[09:10:14] Finished 'default' after 4.58 μs
```

whyte624:
Yes, something similar. Looks good for me. :+1:

asake2007:
[...]

PR #589, opened 2016-03-11
Task duration

whyte624:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #526

### DIFF
--- a/src/Executor/Informer.php
+++ b/src/Executor/Informer.php
@@ -62,7 +62,7 @@ class Informer
             $seconds = floor($millis / 1000);
             $millis = $millis - $seconds * 1000;
             $taskTime = ($seconds > 0 ? "{$seconds}s " : "")."{$millis}ms";
-            $this->output->writeln("<info>✔</info> Ok after {$taskTime}");
+            $this->output->writeln("<info>✔</info> Ok [$taskTime]");
         }
     }
 
--- a/src/Executor/Informer.php
+++ b/src/Executor/Informer.php
@@ -18,6 +18,11 @@ class Informer
     private $output;
 
     /**
+     * @var int
+     */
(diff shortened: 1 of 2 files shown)


src/Executor/Informer.php
```diff
@@ -51,7 +57,12 @@ public function endTask()
         if ($this->output->getVerbosity() == OutputInterface::VERBOSITY_NORMAL && !$this->output->getWasWritten()) {
             $this->output->write("\033[k\033[1A\r<info>✔</info>\n");
         } else {
-            $this->output->writeln("<info>✔</info> Ok");
+            $endTime = round(microtime(true) * 1000);
+            $millis = $endTime - $this->startTime;
+            $seconds = floor($millis / 1000);
+            $millis = $millis - $seconds * 1000;
+            $taskTime = ($seconds > 0 ? "{$seconds}s " : "")."{$millis}ms";
+            $this->output->writeln("<info>✔</info> Ok after {$taskTime}");
```
elfet: May be add some format? Like gray colors: `<info>✔</info> Ok <gray>[$taskTime]</gray>`

whyte624: Is there such tag as <gray> in output? When I tested it looked fine. Anyway, if you insist, I can write something like `<fg=gray>[$taskTime]</>`.

elfet: Ok!

elfet:
And please rebase.

src/Executor/Informer.php
whyte624: Unfortunately there is no gray color for output. So, I've added just square brackets.

PR #592, opened 2016-03-11
Run deploy:writable again after cache warmup

nemoinho:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #575 

As described in ticket #575 the cache is not writable after the deployment of symfony if the deployment-user is not the httpd-user. Therefore we would make the cache-directory writable again after the
cache-warmup.  This is necessary because symfony may not do that and the deploy user should know who is able to write into the cache.  You can finally define this by using `set('writable_dirs', […])` and
`set('http_user', '…')`.

### DIFF
--- a/recipe/symfony.php
+++ b/recipe/symfony.php
@@ -126,6 +126,7 @@ task('deploy', [
     'deploy:vendors',
     'deploy:assetic:dump',
     'deploy:cache:warmup',
+    'deploy:writable',
     'deploy:symlink',
     'cleanup',
 ])->desc('Deploy your project');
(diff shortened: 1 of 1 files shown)


elfet:
I think ok. @oanhnn? @deployphp/maintainer ?

sidz:
I think you don't need to duplicate deploy:writable task. We only needed deploy:writable after 'deploy:cache:warmup'.

I tested and everything is ok.

elfet:
@nemoinho it's looks like @sidz is right. Can you test it and update PR?

nemoinho:
You are totally right, so I changed the task as suggested.

Issue #495, opened 2015-12-08
Symfony 3.0: dependencies are incompatible


dvzzz:
Composer log:

```
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - deployer/deployer v3.0.5 requires phpseclib/phpseclib dev-master#2c96af214bf1b5e29b707249108504b4e0041a21 -> no matching package found.
    - deployer/deployer v3.0.4 requires phpseclib/phpseclib dev-master#2c96af214bf1b5e29b707249108504b4e0041a21 -> no matching package found.
    - deployer/deployer v3.0.3 requires phpseclib/phpseclib dev-master#2c96af214bf1b5e29b707249108504b4e0041a21 -> no matching package found.
    - deployer/deployer v3.0.2 requires phpseclib/phpseclib dev-master#2c96af214bf1b5e29b707249108504b4e0041a21 -> no matching package found.
    - deployer/deployer v3.0.1 requires phpseclib/phpseclib dev-master#2c96af214bf1b5e29b707249108504b4e0041a21 -> no matching package found.
    - deployer/deployer v3.0.0 requires phpseclib/phpseclib dev-master#2c96af214bf1b5e29b707249108504b4e0041a21 -> no matching package found.
    - Conclusion: remove symfony/symfony v3.0.0
    - deployer/deployer v3.0.10 requires symfony/finder ~2.6 -> satisfiable by symfony/finder[v2.6.0, v2.6.1, v2.6.10, v2.6.11, v2.6.12, v2.6.2, v2.6.3, v2.6.4, v2.6.5, v2.6.6, v2.6.7, v2.6.8, v2.6.9, v2.7.0, v2.7.1, v2.7.2, v2.7.3, v2.7.4, v2.7.5, v2.7.6, v2.7.7, v2.8.0].
    - deployer/deployer v3.0.6 requires symfony/finder ~2.6 -> satisfiable by symfony/finder[v2.6.0, v2.6.1, v2.6.10, v2.6.11, v2.6.12, v2.6.2, v2.6.3, v2.6.4, v2.6.5, v2.6.6, v2.6.7, v2.6.8, v2.6.9, v2.7.0, v2.7.1, v2.7.2, v2.7.3, v2.7.4, v2.7.5, v2.7.6, v2.7.7, v2.8.0].
[...]

PR #593, opened 2016-03-13
Symfony 3 compatibility

cedricziel:
This introduces a change to the travis build that instructs builds with PHP 5.4 to use the lower bound dependencies.

| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #495 

Fixes: #495 
Replaces: #556

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ language: php
 sudo: required
 
 php:
-  - 5.4
   - 5.5
   - 5.6
   - 7.0
@@ -12,16 +11,19 @@ php:
 matrix:
   allow_failures:
     - php: hhvm
+  include:
+    - php: 5.4
+      env: COMPOSER_FLAGS="--prefer-lowest"
 
 install:
   - eval `ssh-agent -s`
   - test/travis/setup-secure-shell.sh
   - composer self-update
-  - composer install --no-interaction --prefer-source
+  - composer update --no-interaction --prefer-source $COMPOSER_FLAGS
 
 script:
   - phpunit
-  
+
 notifications:
   webhooks:
     urls:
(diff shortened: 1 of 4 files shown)


cedricziel:
I added a 3rd commit that *should* trigger scrutinizer to use PHP 5.5 which is still reasonable since it just does the static analysis and the coverage - and travis performs the builds for the different PHP versions.

.travis.yml
```diff
@@ -12,16 +11,19 @@ php:
...
     - php: hhvm
+  include:
+    - php: 5.4
+      env: COMPOSER_FLAGS="--prefer-lowest"
 
 install:
   - eval `ssh-agent -s`
   - test/travis/setup-secure-shell.sh
   - composer self-update
-  - composer install --no-interaction --prefer-source
```
elfet: Why update?

cedricziel: The commit message specifies: The new Symfony Components v3 dependencies are written to the lock-file and thus would be installed on the travis PHP 5.4 build as well. Since they depend on PHP 5.5, the composer install command would fail.

Since PHP 5.4 needs to be kept - the locked composer dependencies need to be downgraded on travis build, which needs the added composer flags.

elfet: I see. Thinks come complicated. I think to drop support for 5.4.

cedricziel: I'd drop it since it's EOL anyways.

So. - Since the builds pass and the removal of php 5.4 support is not subject to the pull request, do you think we can get this in?

elfet: I think yes %)

PR #594, opened 2016-03-14
Adds check for existing /current directory to deploy:prepare

vicdelfant:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

Re-submission of #583, but instead of enforcing the removal of this directory we'll now throw an exception.

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -103,6 +103,13 @@ task('deploy:prepare', function () {
 
     run('if [ ! -d {{deploy_path}} ]; then mkdir -p {{deploy_path}}; fi');
 
+    // Check for existing /current directory (not symlink)
+    $result = run('if [ ! -L {{deploy_path}}/current ] && [ -d {{deploy_path}}/current ]; then echo 1; fi')->toString();
+    if ($result)
+    {
+        throw new RuntimeException('There already is a directory (not symlink) named "current" in the deploy_path. Remove this directory so it can be replaced with a symlink for atomic deployments.');
+    }
+
     // Create releases dir.
     run("cd {{deploy_path}} && if [ ! -d releases ]; then mkdir releases; fi");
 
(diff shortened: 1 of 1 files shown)


recipe/common.php
```diff
@@ -103,6 +103,12 @@
 
     run('if [ ! -d {{deploy_path}} ]; then mkdir -p {{deploy_path}}; fi');
 
+    // Check for existing /current directory (not symlink)
+    $result = run('if [ ! -L {{deploy_path}}/current ] && [ -d {{deploy_path}}/current ]; then echo 1; fi')->toString();
```
elfet: It's better to use `toBool()`:
```php
run('if [...]; then echo true; fi')->toBool();
```

recipe/common.php
```diff
@@ -103,6 +103,12 @@
...
+    if ($result) {
+        throw new RuntimeException('There already is a directory (not symlink) named "current" in the deploy_path. Remove this directory so it can be replaced with a symlink for atomic deployments.');
```
elfet: Please, insert real deploy path in error message.

recipe/common.php
```diff
@@ -103,6 +103,12 @@
 
     run('if [ ! -d {{deploy_path}} ]; then mkdir -p {{deploy_path}}; fi');
 
+    // Check for existing /current directory (not symlink)
+    $result = run('if [ ! -L {{deploy_path}}/current ] && [ -d {{deploy_path}}/current ]; then echo true; fi')->toBool();
+    if ($result) {
+        throw new RuntimeException('There already is a directory (not symlink) named "current" in ' . env('deploy_path') . '. Remove this directory so it can be replaced with a symlink for atomic deployments.');
```
mbrodala: Did you mean `\RuntimeException` instead? ;-)

elfet: now ton in namespace. May be problem for v4.

vicdelfant: Fixed in 7ecee77. Maybe there should be another PR to fix the other occurrences of RuntimeException without an explicit namespace ;)

PR #600, opened 2016-03-21
Fixed SSH upload method bug

Whyounes:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? |No
| Fixed tickets | N/A

The `Ssh\Sftp\send` method return `false` on failure. The `!$this->session->getSftp()->send($local, $remote)` check will throw and exception if the file is empty (0 bytes sent).


elfet:
Thanks

Issue #607, opened 2016-04-03
ability to change GIT executable path


mikk150:
| Q                 | A
| ----------------- | ---
| Issue Type        | Feature Request
| Deployer Version  | N/A
| Local Machine OS  | N/A
| Remote Machine OS | N/A

It would be nice if we could change GIT executable path OR modify $PATH variable (latter would be even better)

oanhnn:
Please follow and discuss in #586

oanhnn:
This will implement in v4.x.

PR #606, opened 2016-04-01
Add initial support for custom php and php-cli binary.

gperdomor:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #586 

This add initial support for configure a custom php for each server.

Some servers, need use php-cli to install vendors. This PR use a new env var (<code>composer_require_cli</code>) to handle that too

Issue #586, opened 2016-03-11
Configurable bin file


elfet:
| Q                 | A
| ----------------- | ---
| Issue Type        | Feature Request
| Deployer Version  | 4.x
| Local Machine OS  | N/A
| Remote Machine OS | N/A
#538 
Allow user to specify there to look for php bin.

```php
env('php_bin', function () {
    return run('which php');
});
```

oisvidi:
Support for reading composers bin-dir instead of writing it multiple places? https://getcomposer.org/doc/06-config.md#bin-dir

nemoinho:
He meant PHP itself. I really like the idea, because than I can use deployer as well to deploy legacy code which may needs special php-versions.

gperdomor:
@elfet i submit a PR to handle this issue :D

elfet:
I see, need more time for testing.

oanhnn:
@elfet I think we allow setting up all bin path like:
```
env('bin.php', function() {
   return run('which php');
});
env('bin.git', '/usr/local/git1.7/bin/git');
env('bin.composer', null); // for using local composer
```

Because in some cases, users using vagrant, they can mount software sources (like php, git, apache, ..) and build custom.

elfet:
What about creating new mechanizm for bins? Like PATH or something?

elfet:
`run('php ...');` Maybe some replace('php', '/...')?

oanhnn:
What are you think a bout multi version php. Fix PATH is not good solution.

oanhnn:
> run('php ...'); Maybe some replace('php', '/...')?
Users can custom like `env('bin.git', '/usr/local/git1.7/bin/git');`

gperdomor:
I don't like the idea of modify the $PATH env. But i like to have a map of bin paths, with default values and configurable by server. For example, i have a <code>deploy.yml</code> file with the deploy configuration, with the map of bin executable shoul be something like this:

mmmmm having <code>env(bin.executable, path_to_executable)</code> sounds great... and can be configured by server. In my case i have a <code>deploy.yml</code> with this content:
<pre><code>timezone: America/Caracas

git:
  repository: url_to_repo.git
  branch: master

servers:
  prod:
    host: domain.com
    stage: production
    deploy_path: 'path/for/deploy'
    port: 22
    ssh_user: user
    password: password
    bin: # Alternative names: bins, exec, executables, paths?
        php: '/usr/local/bin/php'
[...]

Issue #548, opened 2016-02-02
Split downloading of composer.phar and installing project dependencies in deploy:vendors task


arogachev:
I suggest to split `deploy:vendors` task to two separate tasks:

- 1st will just download `composer.phar` file.
- 2nd will install project Composer packages.

The benefit of this is we can execute custom tasks after 1st one is completed, for example: installation / update of global packages.

Example with Yii2 framework - we need to install [Composer Asset Plugin](https://github.com/francoispluchino/composer-asset-plugin) *before* installing project packages:

```
composer global require "fxp/composer-asset-plugin:~1.1.1"
```

This is only needed on first installation, then we only need to update this plugin.

Skipping installation of this plugin leads to errors and application will not be working. Currently both Yii2 recipes ([basic](https://github.com/deployphp/deployer/blob/master/recipe/yii2-app-basic.php) and [advanced](https://github.com/deployphp/deployer/blob/master/recipe/yii2-app-advanced.php)) lack this, that means users have to install it separately on their own.

This is not limited to just Yii2 of course.

Writing custom task is not appropriate because there is no way to inject custom task now and we need duplicate code from `deploy:vendors`.

Related - https://github.com/deployphp/deployer/issues/518.

nadar:
:+1:

elfet:
I think this is cool thing to have. May be in next release 4.x.

arogachev:
@elfet So what's the reason of closing in this case?

elfet:
Because we found better solution: custom bin. Will be in next release.

arogachev:
OK, this sounds more clear.

PR #608, opened 2016-04-04
Custom bin

elfet:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | Yes `composer_command`
| Fixed tickets | #548 #607 #586 #606

This is my version of custom bins. Here an example:

```php
task('default', function () {
    run('{{bin/php}} -v');
    run('{{bin/git}} --version');
    run('{{bin/composer}} --version');
});
```
What do you think, guys?

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -15,7 +15,6 @@ set('copy_dirs', []);
 set('writable_dirs', []);
 set('writable_use_sudo', true); // Using sudo in writable commands?
 set('http_user', null);
-set('composer_command', 'composer'); // Path to composer
 set('clear_paths', []);         // Relative path from deploy_path
 set('clear_use_sudo', true);    // Using sudo in clean commands?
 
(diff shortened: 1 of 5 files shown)


nemoinho:
All in all it looks good to me but would you really download composer if it's missing? I think that should only rise an exception and stop deployer. And maybe deployer should test for a few common locations of composer before doing that. I really don't like the idea, that a application loads extra software without warning or at least an info.

elfet:
For now composer automatically downloaded too. If we remove this, may broke BC.

elfet:
Also it's downloaded via `https`, so risks a minimal. I think this behavior is good for developer experience, and we should keep it. You can always install composer manually to prevent downloading it.

nemoinho:
Yep, unfortunately, but maybe we can keep this topic  on the agenda for a next major release.

elfet:
:+1:

recipe/common.php
```diff
@@ -39,6 +38,30 @@
...
+});
+
+env('bin/composer', function () {
+    if (commandExist('composer')) {
+        $composer = run('which composer')->toString();
+    }
+
+    if(empty($composer)) {
+        run("cd {{release_path}} && curl -sS https://getcomposer.org/installer | php");
+        $composer = 'php composer.phar';
```
oanhnn: Please replace with `{{bin/php}} composer.phar`

elfet: :+1:

oanhnn:
Users can void download `composer.phar` by a task upload `composer.phar` to shared directory and:
```
env('bin/composer', '{{bin/php}} {{deploy_path}}/shared/composer.phar');
```
:smile:

elfet:
@deployphp/maintainer merge? :)

oanhnn:
I think we will merge to both 3.1.x and 4.x

oanhnn:
@elfet I merged branch `custom_bin` and `master` to [`oanhnn/4.x`](https://github.com/oanhnn/deployer/tree/4.x). Now you can merge :smile:

elfet:
I think merge master to 4.x too :)

oanhnn:
Yeah, I merged :smile:

rvanlaak:
@antonmedv would it be easy to change the php bin per host?

Our use-case: we would like to start experimenting with PHP7.4 without affecting the other applications on that host.
The vhost was changed to let Apache serve PHP7.4, but deployment fails because CLI is on PHP7.3.

Issue #619, opened 2016-04-11
Laravel Recipe Improvements


barryvdh:
| Q | A |
| --- | --- |
| Issue Type | Feature Request |

I see a few issues have been opened for Laravel. To keep it central, imho it needs the following features:
- [x] Storage/writable folder improvements (#572 #622 + below)
- [x] Public storage folder symlink to shared storage (see quote below)
- [ ] Clear cache after deployments (eg. cached views/html parts; `php artisan cache:clear`
- [ ] Clear old cache and cache new config; both for speed and security (`php artisan config:cache`) (#527)
- [x] Update shared folders _before_ vendors. This because of `php artisan optimize` being dependant on environment, so needs shared `.env` file.
- [ ] Add optional helpers tasks for `php artisan migrate` and `php artisan db:seed`, using the `--force` flag for non-interactive mode (but should perhaps confirm on the deployer side. And perhaps `php artisan migrate:status` for info. (#541 #622)

About Laravel folders, should we update the recipe to follow the 'guidelines' set by Taylor Otwell http://taylorotwell.com/spark-storage/

> When using deployment systems like Envoyer, an entirely fresh copy of the repository is installed with each deploy and the storage directory is attached to a permanent storage directory using a symbolic link. If you store your user uploads in public, they will all go missing on your subsequent deployment. **The solution is to store the uploads in your persistent storage directory and then create a symbolic link from your public directory to your storage directory during your deployment process.**
> 
> Before yesterday, there was no conventional place to store these user uploads other than “anywhere you want within the storage directory”. After reviewing a few libraries in other languages, it became apparent that there would be great benefit to designating a standard location within the storage directory for files intended to be publicly accessible. So, I created the storage/app/public directory in the Laravel repository so that it will be present on all future installs of Laravel 5.2.

So the default would be this
- Shared: `.env` + entire `storage` folder
- Writable:  `storage` + `bootstrap/cache` (I would recommend sharing the entire storage folder)
- Create symlink `public/storage` to `storage/app/public` (`run("ln -s {{deploy_path}}/storage/app/public {{deploy_path}}/public/storage ");` ?)

oanhnn:
@barryvdh Can you send link about 

> Create symlink public/storage to storage/app/public (run("ln -s {{deploy_path}}/storage/app/public {{deploy_path}}/public/storage "); ?)

in Laravel document. I read http://taylorotwell.com/spark-storage/ but i support it only it is included Laravel document.

[...]

Issue #572, opened 2016-03-06
Laravel application and shared folders


dexbarrett:
I'm trying out this library and everything seemed going well until I noticed some problems with the folders the Laravel framework needs to be writable. If I only set those folders as writable and not shared the permissions are set correctly, but if I set those folders as both writable and shared they won't get the writable permissions. I don't use the Laravel recipe but a custom file based on the majority of that recipe; I just added the **writable_dirs** task and some other things my app needs.

Am I missing something or why could this be happening? I could leave those folders as only writable but not shared, but that wouldn't be useful because shared folders are the main point of an atomic deployment.

Hope I explained myself. Thank you.

philbates35:
I noticed this too. I found that the writable permission was being set correctly on the `storage` directory in each release (it appears in each release because `storage` isn't a shared directory, it's the subdirectories in the `storage` directory that are shared), but the writable permission wasn't being set on the subdirectories that were shared (`storage/app` etc).

elfet:
Hi guys. Need to check if deploy:writeable task can make shared folders writable. May be this feature is missing.
@oanhnn can you check?

oanhnn:
@dexbarrett @philbates35 
Please show a part of your `deploy.php` file. It may be help me to reproduce this issue.

dexbarrett:
@oanhnn Sure. This is my deploy.php file. I'm testing it locally first, before testing on a VPS. 

    require_once 'recipe/common.php';

    //shared dirs
    set('shared_dirs', [
        'storage/app',
        'storage/framework/cache',
        'storage/framework/sessions',
        'storage/framework/views',
        'storage/logs',
        'public/images/book-covers',
        'public/images/book-photos/photos',
        'public/images/book-photos/thumbnails'
    ]);


    //writable dirs
    set('writable_dirs', [
    	'bootstrap/cache',
    	'storage',
    	'public/images/book-covers',
        'public/images/book-photos/photos',
        'public/images/book-photos/thumbnails'
    ]);

    //tasks

    task('migration', function() {
      run('php {{release_path}}/artisan migrate --force');
    })->desc('Artisan migrations');

    task('env_file', function() {
    	upload('/home/localuser/.librosjbenv', '{{release_path}}/.env');
    })->desc('Copy environment file');

    task('reload:apache', function() {
    	run('service apache2 reload');
    })->desc('Reload Apache Server');


    task('deploy', [
        'deploy:prepare',
        'deploy:release',
        'deploy:update_code',
[...]

PR #622, opened 2016-04-12
Bugfix laravel recipe

oanhnn:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #527 #572 

In this PR:
- Replace `php` by `{{bin/php}}`
- ~~Add helper tasks about migration and option `--force` can set by `env()`~~
- Resolved issues about shared directories and `.env` file
- Resolved issues about writable `storage` directory. Tested with `setfacl` on CentOS
- Changed order task deploy:shared and deploy:vendors in laravel recipe (See #619)

### DIFF
--- a/recipe/laravel.php
+++ b/recipe/laravel.php
@@ -20,12 +20,12 @@ set('writable_dirs', ['bootstrap/cache', 'storage']);
 /**
  * Helper tasks
  */
-task('deploy:up', function () {
+task('artisan:up', function () {
     $output = run('{{bin/php}} {{deploy_path}}/current/artisan up');
     writeln('<info>'.$output.'</info>');
 })->desc('Disable maintenance mode');
 
-task('deploy:down', function () {
+task('artisan:down', function () {
     $output = run('{{bin/php}} {{deploy_path}}/current/artisan down');
     writeln('<error>'.$output.'</error>');
 })->desc('Enable maintenance mode');
@@ -37,8 +37,8 @@ task('deploy', [
     'deploy:prepare',
     'deploy:release',
     'deploy:update_code',
-    'deploy:vendors',
     'deploy:shared',
+    'deploy:vendors',
     'deploy:writable',
     'deploy:symlink',
     'cleanup',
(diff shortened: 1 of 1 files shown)


elfet:
Does it will be working without  `--force` at all?

barryvdh:
Migrate is useless without `--force` on production, it will never run.

What about adding options to the `deploy:migrate --force` command (if that's possible), or asking confirmation before running?

oanhnn:
WIth this PR, users can use `--force` option or not by `env()`

```
sever('prod', 'xx.xx.xx.xx', 22)
  ->env('allow_force_migrate', false);

server('dev', 'xx.xx.xx.xx', 22)
   ->env('allow_force_migrate', true);
```

Default `env('allow_force_migrate', false);`   
And task `deploy:migrate` is not included `deploy` task.

barryvdh:
I understand, but without `--force`, it's just useless.

barryvdh:
I've listed some more ideas BTW, if you feel like adding more stuff ;)
https://github.com/deployphp/deployer/issues/619

oanhnn:
@barryvdh 
Option `--force` is very dangerous. Users have to control it. We wont add it as default.

oanhnn:
@elfet Please review and merge and close issue refs

elfet:
If `--force` is stand only for *non-interactive*, i think it's ok to force all the time (Better DX).

barryvdh:
I understand, but setting the force value in your config doesn't make it less dangerous (because it's set in the deploy file, so you still run it accidentally)
Why not do `askConfirmation('Are you sure you want to migrate')` with an option to always force?

elfet:
`askConfirmation` it's always bad pattern in automation tools :(

barryvdh:
Didn't you put that in the functions.php? :P And it's also in drupal7 recipe. But fair enough.
Imho the `--force` flag is similar to --non-interactive. I opened and issue here to fix that: https://github.com/laravel/framework/pull/13125

oanhnn:
@elfet @barryvdh 
I split this PR to:
- __Bugfix__ This PR only fix bug of laravel recipe
- __Feature__ Other PR will add helper task

PR #625, opened 2016-04-13
Add helper tasks for laravel recipe

oanhnn:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #541 #619 

Added helper tasks:
- `artisan:migrate` => `artisan migrate --force`
- `artisan:migrate:rollback` => `artisan migrate:rollback --force`
- `artisan:migrate:status` => `artisan migrate:status`
- `artisan:db:seed` => `artisan db:seed --force`
- `artisan:cache:clear` => `artisan cache:clear`
- `artisan:config:cache` => `artisan config:cache`
- `deploy:public_disk` => Create symlink `public/storage` to `storage/app/public`

### DIFF
--- a/recipe/laravel.php
+++ b/recipe/laravel.php
@@ -6,7 +6,8 @@
  */
 
 require_once __DIR__ . '/common.php';
-// This recipe support Laravel 5.1+, with orther version, please see document https://github.com/deployphp/docs
+// This recipe support Laravel 5.1+,
+// with orther version, please see document https://github.com/deployphp/docs
 
 // Laravel shared dirs
 set('shared_dirs', ['storage']);
@@ -59,6 +60,11 @@ task('artisan:config:cache', function () {
     $output = run('{{bin/php}} {{deploy_path}}/current/artisan config:cache');
     writeln('<info>' . $output . '</info>');
 })->desc('Execute artisan config:cache');
+
+task('artisan:route:cache', function () {
+    $output = run('{{bin/php}} {{deploy_path}}/current/artisan route:cache');
+    writeln('<info>' . $output . '</info>');
+})->desc('Execute artisan route:cache');
 
 /**
  * Task deploy:public_disk support the public disk.
--- a/recipe/laravel.php
+++ b/recipe/laravel.php
@@ -7,7 +7,7 @@
 
 require_once __DIR__ . '/common.php';
 // This recipe support Laravel 5.1+,
-// with orther version, please see document https://github.com/deployphp/docs
+// with other version, please see document https://github.com/deployphp/docs
 
 // Laravel shared dirs
 set('shared_dirs', ['storage']);
--- a/recipe/laravel.php
+++ b/recipe/laravel.php
@@ -23,6 +23,38 @@ set('shared_files', ['.env']);
 // Laravel writable dirs
 set('writable_dirs', ['bootstrap/cache', 'storage']);
 
+task('deploy:migrate', function () {
+    $output = run('{{bin/php}} {{deploy_path}}/current/artisan migrate --force');
+    writeln('<info>' . $output . '</info>');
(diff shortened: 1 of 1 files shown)


barryvdh:
:+1: 

I'm not familiar with the code, but perhaps it would be a viable option to map all artisan commands automatically? eg. `dep artisan command arg1 arg2 --option` -> `{{bin/php}} {{deploy_path}}/current/artisan command arg1 arg2 --option`. This would reduce boilerplate and people can run artisan commands on the external server easy (eg. `artisan migrate --prepend`, `artisan db:seed --class=UserSeeder` etc.)
The same could apply for Drupal (drush), Magento (magerun), Wordpress (wp-cli), Symfony etc.

It would also make the argument about --force invalid, as you can do just `dep artisan migrate --force`.

adriandmitroca:
I don't think that using `--force` by default is a good idea. This way you could overwrite important data by mistake.

Secondly, what do you think about running `php artisan migrate` on every deployment? Application should not be deployed with outdated database scheme because it could case unnecessary exceptions.

barryvdh:
1. Adding `--force` just skips the warning on production. It doesn't overwrite anything else. Running it without --force just doesn't work on production. Not ever.
2. Either you run it without --force and it doesn't have any effect, or you run it with --force and you have the potential to run unwanted migrations. So *that* is dangerous imho..

adriandmitroca:
@oanhnn 

You have also missed `route:cache` command - could be triggered just after `config:cache`.

oanhnn:
@adriandmitroca 
I was try `route:cache` command but i see it has a problem.
```
 [LogicException] 
Unable to prepare route [/] for serialization. Uses Closure.
```
May be `route:cache` not support Closure. I think it will throw error when deploy.

oanhnn:
@barryvdh 
I can make `dep artisan <stage> <artisan:sub command> <options>`, but options can use for deployer or artisan, that is not clear and may be have many error. I think everyone should using custom task by themself, that is better.

barryvdh:
Looks good to me.

Not sure if this belongs here, but perhaps it's helpful to add a note about reloading PHP FPM, especially because a lot of Laravel sites are running on Forge, which doesn't refresh the symlinks automatically..

Forge recipe (which you can run on your servers in the Forge panel) to allow php7.0-fm to be run without sudo password:
`echo "forge ALL=NOPASSWD: /usr/sbin/service php7.0-fpm reload" > /etc/sudoers.d/php-fpm`

Command to run after the release symlink:
`sudo /usr/sbin/service php7.0-fpm reload`

Although this is general (and similar for PHP5),

oanhnn:
@elfet Please review and merge it (by button "Squash and merge")

garygreen:
Is there a nice way to 'extend' the existing laravel recipe? I would like to add route:cache as I think it should always be there (I don't use closure routes)...also I want auto migrate. Maybe those can be options somehow?

oanhnn:
@garygreen 
You can make `deploy.php`
```
<?php
# deploy/php

require 'recipe/common.php';

task('artisan:route:cache', function () {
    $output = run('{{bin/php}} {{deploy_path}}/current/artisan route:cache');
    writeln('<info>' . $output . '</info>');
})->desc('Execute artisan route:cache');

task('deploy', [
    'deploy:prepare',
    'deploy:release',
    'deploy:update_code',
    'deploy:shared',
    'deploy:vendors',
    'deploy:writable',
    'deploy:symlink',
    'cleanup',
    'artisan:cache:clear',
   'artisan:migrate',
    'artisan:config:cache',
    'artisan:config:cache',
])->desc('Deploy your project');
after('deploy', 'success');

// Your custom task and servers

```

PR #639, opened 2016-04-25
Enabling 2-factor with private key + password

jdavidbakr:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | 638

The underlying SFTP library allows for multiple authentication methods, so I've added a new setting AUTH_BY_IDENTITY_FILE_AND_PASSWORD.  I would imagine a better way to handle this would be to queue up each authentication method in an array and pass them as the arguments for the sftp->login() call, but I felt that I had too much potential to create breakage - so instead I created an additional method.

To use:

``` php
server('staging', 'example.com', 22)
    ->user('username')
    ->identityFileAndPassword() // Arguments are the 3 identityFile() arguments + the password() argument.  It will assume those defaults if nothing is passed.
    ->stage('staging')
    ->env('branch', 'staging')
    ->env('deploy_path', '/var/www/html/staging');
```

### DIFF
--- a/src/Server/Builder.php
+++ b/src/Server/Builder.php
@@ -154,6 +154,25 @@ class Builder
     }
 
     /**
+     * Authenticate with public key + password (2-factor)
+     *
+     * @param string $publicKeyFile
+     * @param string $privateKeyFile
+     * @param string $passPhrase
+     * @param string $password
+     *
+     * @return Builder
+     */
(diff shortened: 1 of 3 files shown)


oanhnn:
:+1: It is very cool. We need some unit test.

elfet:
Cool, can yoу add some docs too?
And what about php-ssh2 ext?

jdavidbakr:
I just submitted a pull request for the docs, but am unclear on how to configure the unit tests.

PR #642, opened 2016-05-01
Fix assets installation in Symfony 3

peschee:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

The `assetic:dump` command does not exist in Symfony 3. The `assetic` namespace has been replaced by `assets`. The new command to install assets is `assets:install`. 

This PR overrides the `deploy:assetic:dump` command so it works in Symfony 3. Ideally, the `deploy:assetic:dump` command would be renamed to something like `deploy:assets:install` in the main deploy task. However, this would mean overriding the entire list of tasks.

### DIFF
--- a/recipe/symfony3.php
+++ b/recipe/symfony3.php
@@ -10,6 +10,20 @@ require_once __DIR__ . '/symfony.php';
 /**
  * Symfony 3 Configuration
  */
+ 
+/**
+ * Dump all assets to the filesystem
+ * 
+ * This overrides the Symfony 2 assetic:dump command
+ * in favor of the new assets:install command.
+ */
+task('deploy:assetic:dump', function () {
+    if (!get('dump_assets')) {
+        return;
+    }
+
+    run('{{bin/php}} {{release_path}}/' . trim(get('bin_dir'), '/') . '/console assets:install --env={{env}} --no-debug {{release_path}}/web');
+})->desc('Dump assets');
 
 // Symfony shared dirs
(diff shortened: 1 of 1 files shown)


elfet:
Cool!

cmodijk:
It does work >if< you install the assetic bundle. It's not default for symfony 3 any more but it's still something you can do and the`assetic:dump` works. The `assetic` is not `replaced` by `assets ` and the `assets:install` you replaced it with is also available in symfony 2 but this is run by default using [`post-install`](https://github.com/symfony/symfony-standard/blob/master/composer.json#L35) and `post-update` from the `composer install` command. 

What we should do is set the `set('dump_assets', false);` inside the symfony3.php to disable this by default.

cmodijk:
But i have an other problem with the override, If i try the following the `deploy:dump-real` task is never run. When i remove the override it works. I'm not sure why this happens with tasks which are overwrite.

```
task('deploy:dump-real', function () {
    run('{{bin/php}} {{release_path}}/' . trim(get('bin_dir'), '/') . '/console assetic:dump --env={{env}} --no-debug');
});
after('deploy:assetic:dump', 'deploy:dump-real');
```

oanhnn:
@cmodijk Thank you for the report, i created an issue, #720 
About your problem, i think you should check running flow of tasks

Issue #649, opened 2016-05-12
Timezone inconsistency might delete whole release folder being deployed


tchapi:
| Q                 | A
| ----------------- | ---
| Issue Type        | Bug
| Deployer Version  | 3.2.1
| Local Machine OS  | Mac OS X El Capitan
| Remote Machine OS | Ubuntu 16.04 PHP 7

### Description 
When deploying to a remote server, the name of the release folder is determined with the `date()` function of the local machine.

If another user deploys to the same server with a machine that is not configured to use the same timezone, or any timezone at all, the folder name timestamps will not be consistent to *rsort*, which will trigger a wrong cleanup task and potentially delete the current release (being deployed).

The problem apparently lies in https://github.com/deployphp/deployer/blob/master/recipe/common.php#L38

### Steps to reproduce
Take two machines. One machine on which `php.ini` is the standard release file, and another where you added this line in `php.ini`:

    date.timezone = UTC

Deploy from the "standard release" machine, then deploy from the other : the current release folder will be deleted each time you try to release (see log below).

### Content of `deploy.php`
```php
<?php
require 'recipe/symfony.php';
server('[redacted', '[redacted', 22)
    ->user('deploy')
    ->identityFile('[redacted]')
    ->stage('production')
    ->env('deploy_path', '/var/www/[redacted]'); // Define the base path to deploy your project to.
set('repository', '[redacted]');
```

### Output log
```
✔ Ok [808ms]
➤ Executing task deploy:symlink
↳ on [redacted]
> cd /var/www/[redacted] && ln -sfn /var/www/[redacted]/releases/20160512184857 current
> cd /var/www/[redacted] && rm release
• done on [redacted]
✔ Ok [231ms]
➤ Executing task cleanup
↳ on [redacted]
> rm -rf /var/www/[redacted]/releases/20160512184857
```

elfet:
Yes, it's a bug. I'm thinking to using UTC for releases.

oanhnn:
@tchapi I fixed this bug with #656 , please confirm

tchapi:
Ok thanks to you @coox and @oanhnn ! I will confirm asap

peschee:
@tchapi Any updates?

tchapi:
Sorry for the late response — to me it looks good !

Issue #641, opened 2016-04-29
Release names are hard to read.


ju1ius:
| Q | A |
| --- | --- |
| Issue Type | Feature Request |
| Deployer Version | 3.2 |
| Local Machine OS | N/A |
| Remote Machine OS | N/A |
### Description

Hi,
I find the current way of naming release hard to read, especially if you have several of them during the same day/hour.

Couldn't it be a bit more human friendly in future versions ?
For example:

``` php
// recipe/common.php
/**
 * Release
 */
task('deploy:release', function () {
    $releaseDir = env()->parse("{{deploy_path}}/releases");

    $guard = 42;
    do {
        $time = microtime();
        list($usec, $sec) = explode(' ', $time);
        $date = \DateTime::createFromFormat('U.u', $sec . '.' . ($usec * 1e6));
        $releaseName = $date->format('Y-m-d_H:i:s.u');
        $releasePath = $releaseDir . '/' . $releaseName;
    } while (is_dir($releasePath) && --$guard);

    env('release_name', $releaseName);

    run("mkdir $releasePath");

    run("cd {{deploy_path}} && if [ -h release ]; then rm release; fi");

    run("ln -s $releasePath {{deploy_path}}/release");
})->desc('Prepare release');
```

This way you have releases that look like:

```
+ releases
|-- 2016-04-25_12:05:00.011813
|-- 2016-04-25_18:01:02.614783
|-- 2016-04-26_12:00:03.314854
|-- 2016-04-26_18:03:00.214896
|-- 2016-04-27_12:00:12.114936
|-- 2016-04-27_18:02:33.014975
|-- 2016-04-28_12:04:45.015015
|-- 2016-04-28_18:00:07.215054
|-- 2016-04-29_12:01:18.115094
\-- 2016-04-29_18:00:00.415149
```

Which is quite more readable than

```
+ releases
|-- 20160425120500
|-- 20160425180102
|-- 20160426120003
|-- 20160426180300
|-- 20160427120012
|-- 20160427180233
|-- 20160428120445
|-- 20160428180007
|-- 20160429120118
\-- 20160429180000
```

Don't you think ? :wink:

oanhnn:
I think it not hard. 
I vote using 20160427180233

elfet:
Why not 2016-04-25_12:05:00?

oanhnn:
Easy read but not good for computer :smile:

ju1ius:
> Why not 2016-04-25_12:05:00?

[...]

PR #650, opened 2016-05-12
Format release name according to timezone.

coox:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | _Potentially_
| Deprecations? | No
| Fixed tickets | [649](https://github.com/deployphp/deployer/issues/649)

The `timezone` environment variable was not used when generating release folder name, which could cause the build to break.

This *can* however be considered a breaking change, because users from a timezone that is ahead of UTC, who move towards consistently using UTC instead of their own timezone, _and_ who already deployed enough releases in the last few hours, might see their release folder instantly removed on deploy.

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -35,7 +35,9 @@ env('git_cache', function () { //whether to use git cache - faster cloning by bo
     }
     return version_compare($version, '2.3', '>=');
 });
-env('release_name', date('YmdHis')); // name of folder in releases
+env('release_name', (new DateTime())
+    ->setTimeZone(new DateTimeZone(get('timezone')))
+    ->format('YmdHis')); // name of folder in releases
 
 /**
  * Custom bins.
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -17,11 +17,11 @@ set('writable_use_sudo', true); // Using sudo in writable commands?
 set('http_user', null);
 set('clear_paths', []);         // Relative path from deploy_path
 set('clear_use_sudo', true);    // Using sudo in clean commands?
+set('timezone', 'UTC'); // Timezone used in release folder name
 
 /**
  * Environment vars
  */
-env('timezone', 'UTC');
 env('branch', ''); // Branch to deploy.
 env('env_vars', ''); // For Composer installation. Like SYMFONY_ENV=prod
 env('composer_options', 'install --no-dev --verbose --prefer-dist --optimize-autoloader --no-progress --no-interaction');
@@ -36,7 +36,7 @@ env('git_cache', function () { //whether to use git cache - faster cloning by bo
     return version_compare($version, '2.3', '>=');
 });
 env('release_name', (new DateTime())
-    ->setTimeZone(new DateTimeZone(env('timezone')))
+    ->setTimeZone(new DateTimeZone(get('timezone')))
     ->format('YmdHis')); // name of folder in releases
 
 /**
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -17,11 +17,11 @@ set('writable_use_sudo', true); // Using sudo in writable commands?
 set('http_user', null);
 set('clear_paths', []);         // Relative path from deploy_path
 set('clear_use_sudo', true);    // Using sudo in clean commands?
-set('timezone', 'UTC'); // Timezone used in release folder name
 
 /**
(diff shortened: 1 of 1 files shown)


tchapi:
👍

coox:
Sorry about the first two check-breaking commits. 🙇 

As mentioned above, this change can be considered breaking, in at least two ways:
- users located in a timezone ahead of UTC and who deploy enough times (`keep_releases` times between their timezone and UTC) will potentially delete the folder being release during deploy (just like described in #649, in reverse). To circumvent this problem, it could be a possibility to postfix release names with a [`Z`](https://en.wikipedia.org/wiki/Coordinated_Universal_Time#Time_zones), and treat them differently in sort (this would require chiming in the discussion in #641).
- since `timezone` becomes a global (not environment) variable, it might break deploys of users who use `env('timezone')` in their custom tasks. Not sure how to prevent this, though…

recipe/common.php
```diff
@@ -119,11 +121,6 @@
         throw $e;
     }
 
-    // Set the deployment timezone
-    if (!date_default_timezone_set(env('timezone'))) {
-        date_default_timezone_set('UTC');
-    }
-
```
coox: I believe that this chunk is no longer useful since cd6b297.

oanhnn:
Thank you for your PR.
But i think why not breck BC?
```
env('release_name', function () {
    // Set the deployment timezone
    if (!date_default_timezone_set(env('timezone'))) {
        date_default_timezone_set('UTC');
    }
    return date('YmdHis');
}); // name of folder in releases
```

oanhnn:
Please see #338 #357 #585

recipe/common.php
```diff
@@ -35,7 +35,9 @@
     }
     return version_compare($version, '2.3', '>=');
 });
-env('release_name', date('YmdHis')); // name of folder in releases
+env('release_name', (new DateTime())
```
elfet: Better use a function here.

elfet:
About possible breaking changes. Every think will be okay. Only order of releases may be broken. But we can add some sort of check: is new release name a last one?
```php
is_last(env('releases'), $releaseName)
```

coox:
Thank you @oanhnn.

In the light of the additional context described in the issues that you mentioned, I believe that indeed, your proposed solution of moving the `date_default_timezone_set` bit inside the `release_name` function makes the most sense.

coox:
Also, good idea @elfet. Indeed there could still be a side effect as described in #649. Let me explore this…

coox:
Actually, a check on whether the release currently being deployed is indeed the last one would fix the release order bug on deploy, but could potentially screw up other tasks… Which might be worse, actually (picture the effect on `rollback`).

Currently, all tasks depending on `releases_list` _assume_ that the list is ordered (from last to oldest). So this is the expected contract for this function.

However, the list is `rsort`ed by file name, which only really works when release names are consistent, and is subject to breaking (timezone mix, user names their releases by git hash…).

Wouldn't it be safer to have `release_list` sort the list by release directory create time? Concurrency issues aside, wouldn't solve all potential breakages mentioned here so far?

oanhnn:
@coox 
I created PR #656 , you can review and confirm "Is bug fixed ?"

coox:
@oanhnn — donc, looks 👍  to me.

coox:
@oanhnn, @elfet:
Regarding the discussion on `releases_list` order, I suppose that it goes beyond the scope of current issue: `deployer` could not entirely rely on filesystem data to ensure sorting, it would be too brittle. `deployer` would have to store releases metadata next to the releases themselves, that's quite a bit of work.

Is this a feature that you have envisioned for an upcoming release? What are the plans regarding release metadata on the server? I'm asking because I noticed in #641 that there could be plans for a `dep releases` command, which I think is related, as release metadata storing would be inevitable.

oanhnn:
@coox This feature may be available in Deployer 4.x
@elfet Can you make a feature issue about this feature.
I think it may be:
```
$bin/dep releases
Realease       | Time                        | Reversion
---------------|-----------------------------|-----------
20160501100000 | 10:00:00 May 01, 2016 (UTC) | 56abc52
20160501100513 | 10:05:13 May 01, 2016 (UTC) | dde123f

```

PR #660, opened 2016-05-20
Check for locally installed composer.phar

warren5236:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

If you already have a copy of composer.phar in {{release_path}} Deployer downloads a new copy instead of using the existing file.  This pull request adds a check so it will use that local copy.

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -58,10 +58,19 @@ env('bin/composer', function () {
         $composer = run('which composer')->toString();
     }
 
-    if (empty($composer)) {
-        run("cd {{release_path}} && curl -sS https://getcomposer.org/installer | {{bin/php}}");
-        $composer = '{{bin/php}} {{release_path}}/composer.phar';
+    if (!empty($composer)) {
+        return $composer;
     }
+
+    // check for a locally installed version of composer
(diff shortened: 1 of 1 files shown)


recipe/common.php
```diff
@@ -58,11 +58,20 @@
...
 
-    if (empty($composer)) {
-        run("cd {{release_path}} && curl -sS https://getcomposer.org/installer | {{bin/php}}");
+    if (!empty($composer)) {
+        return $composer;
+    }
+
+    // check for a locally installed version of composer
+    $localFile = env()->parse('{{release_path}}/composer.phar');
+    if (is_file($localFile)) {
```
oanhnn: **Deployer** will check file is existed in remote server. Please using check block instead `is_file`
```
if (run("if [ -f $localFile ]; echo 'true'; endif")->toBool()) {
    // TODO
}
```

oanhnn:
Thankyou. But:
1. I think this case only useful when you commited `composer.phar` to repo or having a task copy `composer.phar` to here before it check. Because, release directory is created new.
2. Please check my comment.

Issue #658, opened 2016-05-20
Send result by email


dimoniet:
| Q | A |
| --- | --- |
| Issue Type | Feature Request |
| Deployer Version | N/A |
| Local Machine OS | N/A |
| Remote Machine OS | N/A |
### Description

What do you think about giving the possibility to send an email when the deploy is finished, telling it's result. We can include all the log that `-vvv` outputs, only some parts, or a message telling that the deploy was OK. Of course this will be optional, and the email could be sent to one or more addresses.

dimoniet:
Of course, I will developt this feature for myself because I need it, but I'm asking if you would like it to send a pull request.

Issue #394, opened 2015-07-13
Idea: Logs


Danielss89:
It would be nice if Deployer kept a log of each deployment. The logs should contain the same as -vvv outputs.

Option to enable/disable logging and choose a log directory.

WDYT?

elfet:
Yes! This is cool feature. Must have!

davidkudera:
:+1:

erickr:
Is anyone doing this? 

Is there a preferred way of passing information between tasks? For example, lets say you use the hipchat recipe, and want to push the deploy log to a hipchat channel, and then later make another recipe collecting the git changelog between this and previous version, and you want to post that to hipchat as well.

davidkudera:
What about using for example [saldaek/monolog](https://github.com/Seldaek/monolog)? Than everyone can configure where to put logs

petermein:
+1

elfet:
Monolog is a great idea!

oisvidi:
If you use a DIC in v4 it would be nice if it was overwriteable before anything was run. So I could replace the container value "logger" with my own psr3 compatible logger. 

Which is monolog but might write to a different file or greylog. 

Would be nice if in my deploy.php I just set a bunch of container values, then the deployer app is run. 

dep cli sets up basic container, loads my deployer scripts which first runs common recipe then my deploy recipe then runs the app with container as a constructor parameter (as defined in the container).

unclead:
why this issue was closed?

PR #662, opened 2016-05-21
Logs, and notificacions by email

dimoniet:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    |  No
| Deprecations? | No
| Fixed tickets | https://github.com/deployphp/deployer/issues/658 & https://github.com/deployphp/deployer/issues/394 ??

I've developed a simple logging system and an email notify system, using monolog for both.
I'm able to decide if I want one or another, or both, setting it's parameters in my deploy.php file.

The logger function expects 2 parameters, the name we want to set to the project inside the log, and the path for the log. Both are mandatory

The emailer function expects 2 parameters, the name we want to set to the project inside the log, and the email address ( or addresses in an array ) where we want to send notifications. There are 3 extra methods to configure 'from' address, the subject and the minimum logging level, based in RFC 5424, when the email will be  sent. This extra methods are optional.

It's not a big implementation, just to cover my needs.

```
logger('project_name', 'log_path');

emailer('project_name', 'email@example.com')
	->from('from@example.com')
	->subject('Error deploying!')
	->level(400);
```

Hope this can help someone with their deploy control.


src/Log/LogCollection.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace Deployer\Log;
+
+use Deployer\Collection\Collection;
+
+class LogCollection extends Collection
```
elfet: Why do you need a log collection?

test/src/Log/LogSenderTest.php
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Deployer\Log;
+
+class LogSenderTest extends \PHPUnit_Framework_TestCase
+{
+    public function testLogSender()
+    {
+        $app = new LogSender('name', 'address');
+
+        $this->assertTrue(method_exists($app, 'writeLog'), 'Class does not have method writeLog');
```
elfet: Test must cover more then that.

test/src/Log/LogWriterTest.php
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Deployer\Log;
+
+class LogWriterTest extends \PHPUnit_Framework_TestCase
+{
+    public function testLogWriter()
+    {
+        $app = new LogWriter('name', 'path');
+
+        $this->assertTrue(method_exists($app, 'writeLog'), 'Class does not have method writeLog');
```
elfet: Redundant tests.

src/Executor/Informer.php
```diff
@@ -28,6 +35,11 @@ class Informer
     public function __construct(OutputWatcher $output)
     {
         $this->output = $output;
+        
+        $deployer = Deployer::get();
+        if ($deployer->logs->has("email")) {
+            $this->sender = Deployer::get()->logs->get("email");
```
elfet: You are getting `email` which is `sender` from `logs`. Better to separate concepts.

elfet:
Log functionality is very important and will be in Deployer for sure. But why do you need email?

src/Log/LogCollection.php
dimoniet: It's not necessary, but I've seen that every function that is configurable from deployer is going to a collection, I just wanted to follow the same workflow you had.

test/src/Log/LogSenderTest.php
dimoniet: New to testing, sorry for not covering much. I will try to write more tests.

test/src/Log/LogWriterTest.php
dimoniet: Same answer than before.

src/Executor/Informer.php
dimoniet: Mails take the information from logs, this is why I set them together.

dimoniet:
In fact what I needed was a notification by mail when an error occurs, because I'm deploying from cron. When I was working on it I realised that I needed some kind of logs to see what happened, and I changed from a mailing system to a log system, taking profit of Monolog's mailing feature.

If you have any improvement or change, tell me and I will try to do it.

oanhnn:
@dimoniet I think you should make a PR with logging feature only.
Users can send email by a custom `success` task.

unclead:
For Logging better to use PSR-3 interface instead of strong coupling with Monolog. Monolog implements PSR-3 and we can easy use it as logger by default,  but user also should should be able to customize logger via configuration, e.g. if i want to use my own logger which implements PSR-3.

dimoniet:
I'm doing a new logging system, with all your requirements.

elfet:
Also there no need for separate logger function. You can use set function:
```php
set('log', './deploy.log');
```

dimoniet:
I was thinking in something like that:

```
//functions.php
function logger($name, $value)
{
    $deployer = Deployer::get();

    if ($value instanceof \Closure) {
        $deployer->setLogger(call_user_func($value));
    }
    elseif (is_string($value)) {
        $logger = new Monolog\Logger($name);
        $logger->pushHandler(new \Monolog\Handler\StreamHandler($value));
        $deployer->setLogger($logger);
    }
    else {
        throw new InvalidArgumentException('Log should be a logger object or the name and path for the default log.');
    }
}
```



Here the user has two options, the "easy" one where he can write only a name and a path and deployer will use Monolog, and the "pro" where the user can configure his own logger configuration.

For example:

```
//deploy.php
logger('deploy',function(){
	$logger = new Monolog\Logger('deploy');
    $logger->pushHandler(new \Monolog\Handler\ErrorLogHandler());

    return $logger;
});
```

What do you think about that? Perhaps we can set only one parameter that would be the path, or the logger object, and put a default name to the log entry.

dimoniet:
Using this "configurable" option I could use the email notiication system embedded in Monolog, that is what I need and wanted to do at first :-)

dimoniet:
One more question, do you want to log -vvv every deploy, or you want to log output console only?

rlacerda83:
@dimoniet, do you finished this implementation?
I need to log all the tasks...

muratbsts:
@elfet, Is there a way saving logs on set function currently? I've tried set like below. But it doesn't work.

```php
set('log', './deploy.log');
````

elfet:
Log system was implemented in deployer v4, but there is no simple way of configuring it right now.

Issue #683, opened 2016-06-03
setfacl: illegal option -- R


uradvd85:
| Q | A |
| --- | --- |
| Issue Type | Bug |
| Remote Machine OS | Freebsd 9.1 |
### Description

root@work:/data/www/site/releases/20160603115438 # cd /data/www/site/releases/20160603115438 && sudo setfacl -R -m u:"uradvd85":rwX -m u:`whoami`:rwX wp-content/uploads
setfacl: illegal option -- R
usage: setfacl [-bdhkn] [-a position entries] [-m entries] [-M file] [-x entries] [-X file] [file ...]
root@work:/data/www/site/releases/20160603115438 #
### Output log

Executing task deploy:writable
↳ on [old]

> cd /data/www/site/releases/20160603115438 && chmod 2>&1; true
> < usage: chmod [-fhv] [-R [-H | -L | -P]] mode file ...
> <
> cd /data/www/site/releases/20160603115438 && if hash setfacl 2>/dev/null; then echo 'true'; fi
> < true
> <
> cd /data/www/site/releases/20160603115438 && sudo setfacl -R -m u:"uradvd85":rwX -m u:`whoami`:rwX wp-content/uploads

  Unable to setup correct permissions for writable dirs.
  You need to configure sudo's sudoers files to not prompt for password,
  or setup correct permissions manually.

  [RuntimeException]
  We trust you have received the usual lecture from the local System
  Administrator. It usually boils down to these three things:
      #1) Respect the privacy of others.
      #2) Think before you type.
      #3) With great power comes great responsibility.
  sudo: no tty present and no askpass program specified

nemoinho:
`-R` and `--recursive` are not supported on a typical freebsd installation (see: [FreeBSD manpage for setfacl](https://www.freebsd.org/cgi/man.cgi?query=setfacl&sektion=1)).

But to make this clear this problem may occur on every non-GNU system (except OSX 😉), so this will also happens on Solaris and maybe other BSD's as well.

A possible simple workaround which also works on GNU systems and OSX could be this solution, but it may needs some more investigation:
    
    $ find DIRECTORY_PATH | xargs setfacl OPTIONS

elfet:
> $ find DIRECTORY_PATH | xargs setfacl OPTIONS
Really need to test this. If some one can test it and create PR – ❤️

elfet:
Solved in v4.

PR #684, opened 2016-06-04
Avoid -R for setfacl and use find instead

nemoinho:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #683 

`setfacl` may has no `-R` option on non-GNU systems therefore we want to avoid this plattform-specific option and use a more general solution instead. This solution is build on `find` and `xargs` because these tools can emulate the whole behaviour smoothly and are widely available.

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -291,8 +291,8 @@ task('deploy:writable', function () {
             } elseif (commandExist('setfacl')) {
                 if (!empty($httpUser)) {
                     if (!empty($sudo)) {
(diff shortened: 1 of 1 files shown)


elfet:
Hmm, do you test it on which platforms?

nemoinho:
I know that Problem from Solaris an used a snippet for that System which I used earlier, but I'll provide also a vagrant configuration with freebsd next week when I have more time.

Maybe I can create a repository with some boxes to test the whole deployer.

elfet:
No progress, closing it. May be we need to return to it after implementing Docker test runner.

PR #688, opened 2016-06-05
Refactoring/Cleanup/Fixes for Symfony-related Recipes

robfrawley:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | Yes
| BC breaks?    | Yes
| Deprecations? | No
| Fixed tickets | N/A

- Removed overridden `deploy:assetic:dump` task in `symfony3` recipe, as it implied this command was no longer valid and should be overridden with `deploy:assets:install` command. In reality, `deploy:assets:install` should exist in all symfony versions, as should `deploy:assetic:dump` (although this is arguable, as `deploy:assetic:dump` isn't a Symfony command, it is a kriswallsmith/assetic command). 
- Changed default for `deploy:assetic:dump` to `false` (via `dump_assets`) as this is not a core Symfony package and should not be assumed to be installed. It relies on kriswallsmith/assetic and symfony/assetic-bundle.
- Added `deploy:assets:install` task as this is a symfony-core command that is intended to be called prior to other asset compilation commands (such as `deploy:assetic:dump`). Its purpose to is to install any bundle assets (within `<bundleName>/Resource/public`) and put them in `web/bundles/<bundleName>`.
- `composer_action` created
- Fixed `SYMFONY_ENV` value in `env_vars` to correctly refer to the configured `env` value.
- Created `bin/console` to cleanup calls to Symfony console command (there was a ton of duplicated lines prior)
- created `console_options` so the options passed to `bin/console` reflect what should be passed per the configured environment (for example, the `--no-debug` options cannot be passed for any environment except for `prod` and will likely break hard when passed for `dev` environment)
- General cleanup...

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -18,13 +18,21 @@ set('http_user', null);
 set('clear_paths', []);         // Relative path from deploy_path
 set('clear_use_sudo', true);    // Using sudo in clean commands?
 
+
+/**
+ * Composer options
+ */
+env('composer_action', 'install');
+env('composer_options', '{{composer_action}} --verbose --prefer-dist --no-progress --no-interaction --no-dev --optimize-autoloader');
+
+
 /**
  * Environment vars
  */
 env('timezone', 'UTC');
 env('branch', ''); // Branch to deploy.
-env('env_vars', ''); // For Composer installation. Like SYMFONY_ENV=prod
-env('composer_options', 'install --no-dev --verbose --prefer-dist --optimize-autoloader --no-progress --no-interaction');
+env('env_vars', ''); // Variable assignment before cmds (for example, SYMFONY_ENV={{env}})
+
 env('git_cache', function () { //whether to use git cache - faster cloning by borrowing objects from existing clones.
     $gitVersion = run('{{bin/git}} version');
     $regs       = [];
@@ -43,6 +51,7 @@ env('release_name', function () {
 
     return date('YmdHis');
 }); // name of folder in releases
+
 
 /**
  * Custom bins.
@@ -73,6 +82,7 @@ env('bin/composer', function () {
 argument('stage', \Symfony\Component\Console\Input\InputArgument::OPTIONAL, 'Run tasks only on this server or group of servers.');
 option('tag', null, \Symfony\Component\Console\Input\InputOption::VALUE_OPTIONAL, 'Tag to deploy.');
 option('revision', null, \Symfony\Component\Console\Input\InputOption::VALUE_OPTIONAL, 'Revision to deploy.');
+
 
 /**
  * Rollback to previous release.
@@ -336,10 +346,7 @@ task('deploy:writable', function () {
  * Installing vendors tasks.
  */
 task('deploy:vendors', function () {
-    $composer = env('bin/composer');
-    $envVars = env('env_vars') ? 'export ' . env('env_vars') . ' &&' : '';
-
-    run("cd {{release_path}} && $envVars $composer {{composer_options}}");
+    run('cd {{release_path}} && {{env_vars}} {{bin/composer}} {{composer_options}}');
 })->desc('Installing vendors');
 
 
--- a/recipe/symfony.php
+++ b/recipe/symfony.php
@@ -7,9 +7,13 @@
 
 require_once __DIR__ . '/common.php';
 
+
 /**
  * Symfony Configuration
  */
+
+// Symfony build env
+env('env', 'prod');
 
 // Symfony shared dirs
 set('shared_dirs', ['app/logs']);
@@ -22,16 +26,28 @@ set('writable_dirs', ['app/cache', 'app/logs']);
 
 // Assets
 set('assets', ['web/css', 'web/images', 'web/js']);
-// Default true - BC for Symfony < 3.0
-set('dump_assets', true);
+
+// Requires non symfony-core package `kriswallsmith/assetic` to be installed
+set('dump_assets', false);
 
 // Environment vars
-env('env_vars', 'SYMFONY_ENV=prod');
-env('env', 'prod');
+env('env_vars', 'SYMFONY_ENV={{env}}');
 
 // Adding support for the Symfony3 directory structure
 set('bin_dir', 'app');
 set('var_dir', 'app');
+
+// Symfony console bin
+env('bin/console', function () {
+    return sprintf('{{release_path}}/%s/console', trim(get('bin_dir'), '/'));
+});
+
+// Symfony console opts
+env('console_options', function () {
+    $options = '--no-interaction --env={{env}}';
+
+    return env('env') !== 'prod' ? $options : sprintf('%s --no-debug', $options);
+});
 
 
 /**
@@ -60,21 +76,25 @@ task('deploy:assets', function () {
         return "{{release_path}}/$asset";
     }, get('assets')));
 
-    $time = date('Ymdhi.s');
-
-    run("find $assets -exec touch -t $time {} ';' &> /dev/null || true");
+    run(sprintf('find %s -exec touch -t %s {} \';\' &> /dev/null || true', $assets, date('Ymdhi.s')));
 })->desc('Normalize asset timestamps');
+
+
+/**
+ * Install assets from public dir of bundles
+ */
+task('deploy:assets:install', function () {
+    run('{{env_vars}} {{bin/php}} {{bin/console}} assets:install {{console_options}} {{release_path}}/web');
+})->desc('Install bundle assets');
 
 
 /**
  * Dump all assets to the filesystem
  */
 task('deploy:assetic:dump', function () {
-    if (!get('dump_assets')) {
-        return;
+    if (get('dump_assets')) {
+        run('{{env_vars}} {{bin/php}} {{bin/console}} assetic:dump {{console_options}}');
     }
-
-    run('{{bin/php}} {{release_path}}/' . trim(get('bin_dir'), '/') . '/console assetic:dump --env={{env}} --no-debug');
 })->desc('Dump assets');
 
 
@@ -82,7 +102,7 @@ task('deploy:assetic:dump', function () {
  * Warm up cache
  */
 task('deploy:cache:warmup', function () {
-    run('{{bin/php}} {{release_path}}/' . trim(get('bin_dir'), '/') . '/console cache:warmup  --env={{env}} --no-debug');
+    run('{{env_vars}} {{bin/php}} {{bin/console}} cache:warmup {{console_options}}');
 })->desc('Warm up cache');
 
 
@@ -90,7 +110,7 @@ task('deploy:cache:warmup', function () {
  * Migrate database
  */
 task('database:migrate', function () {
-    run('{{bin/php}} {{release_path}}/' . trim(get('bin_dir'), '/') . '/console doctrine:migrations:migrate --env={{env}} --no-debug --no-interaction');
+    run('{{env_vars}} {{bin/php}} {{bin/console}} doctrine:migrations:migrate {{console_options}}');
 })->desc('Migrate database');
 
 
@@ -98,10 +118,11 @@ task('database:migrate', function () {
  * Remove app_dev.php files
  */
 task('deploy:clear_controllers', function () {
-    run("rm -f {{release_path}}/web/app_*.php");
-    run("rm -f {{release_path}}/web/config.php");
+    run('rm -f {{release_path}}/web/app_*.php');
+    run('rm -f {{release_path}}/web/config.php');
 })->setPrivate();
 
+// Run after code is checked out
 after('deploy:update_code', 'deploy:clear_controllers');
 
 
@@ -116,6 +137,7 @@ task('deploy', [
     'deploy:shared',
     'deploy:assets',
     'deploy:vendors',
+    'deploy:assets:install',
     'deploy:assetic:dump',
     'deploy:cache:warmup',
     'deploy:writable',
@@ -123,4 +145,5 @@ task('deploy', [
     'cleanup',
 ])->desc('Deploy your project');
 
+// Display success message on completion
 after('deploy', 'success');
--- a/recipe/symfony3.php
+++ b/recipe/symfony3.php
@@ -10,20 +10,6 @@ require_once __DIR__ . '/symfony.php';
 /**
  * Symfony 3 Configuration
  */
- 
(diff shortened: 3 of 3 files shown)


robfrawley:
Perhaps of consequence (is this expected behavior): prior to fixing the travis build by adding an `env` environment variable (see https://github.com/robfrawley/deployer/blob/master/recipe/common.php#L44), it failed on php 5.5 and 5.6, but passed on 7.0.

5.5: https://travis-ci.org/deployphp/deployer/jobs/135430007
5.6: https://travis-ci.org/deployphp/deployer/jobs/135430009
7.0: https://travis-ci.org/deployphp/deployer/jobs/135430011

It seems odd that it would pass on 7.0 without `env` defined...is that possibly a bug?

elfet:
Nice work! 
It's ok on 7.0, but not on 5.5, 5.6?

recipe/common.php
```diff
@@ -18,13 +18,32 @@
 set('clear_paths', []);         // Relative path from deploy_path
 set('clear_use_sudo', true);    // Using sudo in clean commands?
 
+
+/**
+ * Composer options
+ */
+env('composer_action', function () {
+    return env('env') === 'prod' ? 'install' : 'update';
```
oanhnn: I don't think using `composer update` is good action.
It will update your vendors and composer.lock file.

elfet: Agree, install!

oanhnn:
I don't think using composer update is good action.
It will update your vendors and composer.lock file when deploying.
If you want update vendors, i think you should run update on your dev server, test all features and make a commit changed composer.lock file.

recipe/common.php
robfrawley: It seems needlessly restrictive to hard-code the "install" action. Would you be more content to reverse this logic such that update is only called if the environment is set to dev?

`return env('env') === 'dev' ? 'update' : 'install';`

The update command is a valid use case for deployer. For example, I am setting up a project to automatically deploy to a staging server after a successful travis build automatically. Regardless, the default is still to use `install` unless someone changes their environment.

robfrawley:
@oanhnn The intended use-case is for someone who uses deployer to automatically build a staging environment as part of their CI chain. The default action is still to run `install` --- only when run with a non-production environment would `update` be used.

robfrawley:
@elfet The build works on all platforms now that I defined `env` https://github.com/robfrawley/deployer/blob/master/recipe/common.php#L44

Prior to that, `env` was empty, and it would fail on 5.5 and 5.6, but passed on 7.0. I mentioned this as it may have exposed an irregularity and/or bug.

As for this pull-request, it is fixed, but you may still want to look into why it broke selectively without the `env` defined.

oanhnn:
@robfrawley 
If you want using `composer update`, you can make a custom task.

> It will update your vendors and composer.lock file when deploying.
If you want update vendors, i think you should run update on your dev server, test all features and make a commit changed composer.lock file.

https://blog.engineyard.com/2014/composer-its-all-about-the-lock-file

recipe/common.php
```diff
@@ -17,14 +17,34 @@
 set('http_user', null);
 set('clear_paths', []);         // Relative path from deploy_path
 set('clear_use_sudo', true);    // Using sudo in clean commands?
+set('composer_allow_update', false);
```
robfrawley: @oanhnn How is this for a compromise?

robfrawley:
@oanhnn

> I think it makes users puzzled :scream:

Are you implying that the value of stage should define the env programatically somehow? That seems outside the scope of this pull-request.

oanhnn:
Currently, only symfony recipe using `env('env')`. But after merge this PR, all recipe will have to set `env('env')`. Variable name makes users puzzled.

robfrawley:
@oanhnn what do you want to call it? `env` makes sense to me; I don't have any idea what else it would be called...

oanhnn:
What do you think about this code?
```
env('composer_action', 'install');
env('composer_options', 'install --no-dev --verbose --prefer-dist --optimize-autoloader --no-progress --no-interaction');
env('env_vars', ''); // For Composer installation. Like SYMFONY_ENV=prod
```

robfrawley:
@oanhnn That code doesn't allow for composer options based on the environment. I changed `env` to `build_type` ... is that clearer?

robfrawley:
@elfet Do you have any comments about this pull request and what might need changing for it to be accepted?

recipe/common.php
elfet: No update, please. This is not what can be done in recipe by default.

recipe/common.php
```diff
@@ -17,14 +17,34 @@
...
+
+
+/**
+ * Composer options
+ */
+env('composer_action', function () {
+    return get('composer_allow_update') && env('build_env') === 'dev' ? 'update' : 'install';
```
elfet: Only install allowed.

elfet:
`dump` part is ok.
composer `update` is not.

oanhnn:
@robfrawley 
```
env('composer_action', 'install');
env('composer_options', '--no-dev --verbose --prefer-dist --optimize-autoloader --no-progress --no-interaction');
env('env_vars', ''); // For Composer installation. Like SYMFONY_ENV=prod
```

Why?
- No break BC, Don't need change for old using
- Allow change action of composer (`install` or `update`) very flexible and simple
- Don't need `env('env')` of `env('build_env')` for all recipe

@elfet  I am agree with you. Users can use `composer update` by hand or custom task

robfrawley:
@oanhnn and @elfet Ok - does the current commit satisfy all the requirements, taking both your comments into account?

recipe/common.php
```diff
@@ -18,13 +18,21 @@
 set('clear_paths', []);         // Relative path from deploy_path
 set('clear_use_sudo', true);    // Using sudo in clean commands?
 
+
+/**
+ * Composer options
+ */
+env('composer_action', 'install');
+env('composer_options', '{{composer_action}} --verbose --prefer-dist --no-progress --no-interaction --no-dev --optimize-autoloader');
```
robfrawley: @elfet Is this allowable?

recipe/symfony.php
```diff
@@ -22,17 +26,29 @@
 
 // Assets
 set('assets', ['web/css', 'web/images', 'web/js']);
-// Default true - BC for Symfony < 3.0
-set('dump_assets', true);
+
+// Requires non symfony-core package `kriswallsmith/assetic` to be installed
+set('dump_assets', false);
```
oanhnn: Old code
```
// Default true - BC for Symfony < 3.0
set('dump_assets', true);
```
Why do you change BC?

robfrawley: I mentioned in the PR summary that we shouldn't assume asset compilation is desired b/c it isn't a core Symfony command and there is no reason to believe the user has it installed (in any Symfony version). 

That command relies on https://github.com/kriswallsmith/assetic and https://github.com/symfony/assetic-bundle, which may not (and very often isn't) installed.

Behavior is already changing with this pull request, as before the `deploy:assetic:dump` command was *never* called on Symfony >= 2.8. Forcing it by default would be the same as calling `database:migrate` by default, which we don't do (for good reason).

Previous behavior was:
- For all versions: `dump_assets` is true
- For Symfony < `2.8`: via `deploy:assetic:dump`, call `app/console assetic:dump`
- For Symfony >= `2.8`: via an overridden `deploy:assetic:dump, call `app/console assets:install`

The above is inconsistent, odd, and incorrect behavior.

The new behavior is:
- For all versions `dump_assets` is false
- For all versions, via `deploy:assets:install`, call `app/console assets:insall` (core command)
- For all versions, via `deploy:assetic:dump`, call `app/console assetic:dump` IF the user enables it via the above environment var (non-core command)

robfrawley: @oanhnn 

The alternative to retain BC is to define `dump_assets` as `true` in symfony.php and `false` in symfony3.php but this causes inconsistent behavior between the two recipes, which I assert is more confusing than fixing the inconsistency. Thoughts?

robfrawley:
@oanhnn 

FYI, I just wanted to confirm what I had asserted quickly. For reference:

```bash
#!/bin/bash

rm -fr symfony-tmp && mkdir symfony-tmp && cd symfony-tmp
curl -LsS https://symfony.com/installer -o symfony-installer.phar

for version in 2.3 2.7 2.8 3.0 3.1; do
  php symfony-installer.phar -q new $version $version

  if [[ $(grep "assetic" $version/composer.json) ]]; then
    echo "ASSETIC INSTALLED FOR: $version"
  else
    echo "ASSETIC **NOT** INSTALLED FOR: $version"
  fi
done
```

robfrawley:
@elfet Give me a shout if anything needs to be changed for this to be merge-ready. Otherwise, thanks for such a great tool!

recipe/common.php
elfet: This still allow users to use update. And this is ban practice. @oanhnn what you think?

recipe/symfony.php
elfet: What about remove anything about assets from symfony recipe? This will break BC, but we can put in in v4.

elfet:
@robfrawley still need to think and find compromise between no BC break and moving forward. May be this is good for next major v4.

elfet:
@oanhnn what you think?

recipe/symfony.php
robfrawley: That is what *I* would do, actually. I don't it should call `deploy:assetic:dump` by default.

elfet: Yeah, me neither.  But a lot of people using it as it. Well be cool, to not break somethink for any guys :)

robfrawley: The one thing to keep in mind is the current solution *is* broken for Symfony 2.8.

elfet: Well, may be fix it right in 3.x. Need to investigate a little. @oanhnn your thoughts?

recipe/common.php
robfrawley: Crosses fingers... I understand your gut-reaction against running update ad-hoc, but it can be useful for automated deployments to dev server. `compoaser_options` already exists, so it's easy enough to overwrite the whole string, but this just pulls out `composer_action`. Give me a heads up if this is okay @oanhnn and @elfet .

oanhnn: @elfet I think it is ok. Not break BC and users can customize `composer_action` easily.
I thinnk in Deployer, `composer_options` will not include `composer_action`

recipe/symfony.php
oanhnn: @robfrawley I think you should split this problem to a PR, because it is break BC. We will discuss about it more.

robfrawley: @oanhnn Split what into a PR? There isn't really anything that can be separated from *this* PR. Can you elaborate?

recipe/common.php
robfrawley: Is this what you mentioned breaking into a separate PR?

robfrawley:
@elfet @oanhnn Where you guys at on this? Any thoughts or comments! ;-)

robfrawley:
This needs to also have the "BUG" issue label. (Deployer's Symfony recipes breaks when Symfony version 2.8 is used due to Assetic assumption, and the version 3 recipe calls unrelated commands for  "assets:dump" task that it should not.)

robfrawley:
@elfet Is this pull request particularly controversial for some reason I don't see?

# BC Breaks

- User must set `dump_assets` to true for Symfony recipe if they want their assets compiled (whereas previously they did not have to).

# Bug Fixes

- Version 2.8 of Symfony now supported
- Version 3 of Symfony correctly calls "assets:dump" instead of "assets:install" (which should never have been the case).
- All versions of Symfony now have task "assets:install"
- The `env_vars` variable now properly reflects the configured environment (`SYMFONY_ENV={{env}}`)
- Addition of `console_options` environment variable as anonymous function to generate the options to pass for the Symfony console command depending on the defined `env` (calling the console command in prod/dev mode with/without `--no-debug` option fails if the incorrect combination is passed).

# Cleanup/Improvements

- Pulled composer action out of `composer_options` into own environment variable `composer_action`
- Major simplification of "deploy:vendors" anonymous function
- The Symfony console command was moved into own environment variable as anonymous function to allow for cleaner tasks commands that can now be called simply with `{{bin/console}}` instead of having repetitious statements like `trim(get('bin_dir'), '/') . '/console'` everywhere.
- Usage of `sprintf` instead of string/variable concatenation
- Cleanup of file newlines to standardize it (some tasks had two spaces between their definitions, others had one)

robfrawley:
@elfet @oanhnn Any time to look this over and provide feedback?!

I'd love to be able to use deployer directly in my projects instead of relying on my fork with these fixes, but it's not usable on some of the Symfony versions I maintain without the bug fixes here. Can we please get this merged? ;-)

elfet:
Sorry, for delay. Will look tonight or tomorrow.

robfrawley:
@elfet Glad to hear it!

robfrawley:
@elfet Any progress?

robfrawley:
Do you guys have any intention to ever merge this; I don't want to spend the time rebasing it if you have no intention to merge it.

oanhnn:
@robfrawley This branch has conflicts that must be resolved. Please resolved (I think you should make new PR fix issue #720 )

elfet:
Hi everyone! 

Sorry for a BIG delay. After a lot of thinking i'm decided not to merge this PR into 3.x, because of BC. So i'm closing this. 

But all that must be moved into 4.x branch, and i believe that can end up with sooner release on Deployer 4.x.

robfrawley:
@elfet So, after I spend last night and part of the morning working to re-base this and fix the merge conflicts (due to the thumbs up on moving forward from @oanhnn), that work ends up being a waste of my time? This is due to **one extremely minor BC break** (that only affects the versions of Symfony most people no longer use), and apparently outweighs the **6 bug fixes** and **5 improvements**...

I'm a bit speechless guys.

elfet:
We not going to throw you work away. We a going to merge your changes into 4.x branch.

elfet:
I'll open this PR to show what we still really need your work.

elfet:
@robfrawley please check [4.x branch](https://github.com/deployphp/deployer/tree/4.x).

robfrawley:
:thumbsup:

elfet:
I'm planning to speed up Deployer development and PR integration (with BC) by increasing speed of major releases.

PR #692, opened 2016-06-07
Add Blade views composer directory to shared_dirs for Laravel

lasselehtinen:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

Laravel 5.2 cannot create the compiled view if directory does not exist. Without this, you get the following exception:

`file_put_contents(/56ccf3c0f1a96367aabe0a532b30e5bdea56cfcd.php): failed to open stream: Permission denied`

### DIFF
--- a/recipe/laravel.php
+++ b/recipe/laravel.php
@@ -10,7 +10,7 @@ require_once __DIR__ . '/common.php';
 // for older versions, please read the documentation https://github.com/deployphp/docs
 
 // Laravel shared dirs
-set('shared_dirs', ['storage']);
+set('shared_dirs', ['storage', 'storage/framework/views']);
 
 // Laravel 5 shared file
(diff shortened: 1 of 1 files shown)


elfet:
Any Laravel users can test it?

garygreen:
-1 from me. This doesn't seem like a bug to me. By default Laravel automatically has placeholder directories setup for the views cache folder, seems to me like you didn't use `laravel/laravel` as base or deleted the folder by mistake.

elfet:
From laravel docs:

> After installing Laravel, you may need to configure some permissions. Directories within the storage and the bootstrap/cache directories should be writable by your web server or Laravel will not run. 

So closing it.

PR #696, opened 2016-06-09
Fix deleting old releases, if deploy_path is a symlink

Agares:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No (?)
| Deprecations? | No
| Fixed tickets | 

When the {{deploy_path}} is a symlink, find won't find any directories within it. To solve this, the trailing slash is required. This behaviour is explained here: https://chris-lamb.co.uk/posts/find-trailing-slashes-and-symbolic-links-directories

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -357,7 +357,7 @@ task('deploy:symlink', function () {
  */
 env('releases_list', function () {
     // find will list only dirs in releases/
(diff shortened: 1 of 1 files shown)


elfet:
Cool! Thanks!

Issue #354, opened 2015-06-16
Lock and unlock releases


elfet:
We really need to implement this feature. 
- `deploy:release`  
  Then new release dir created, `LOCK` file created within it.  
  If prev release contains `LOCK` file, error throws.
- `deploy:symlink`  
  After symlink `LOCK` file deleted.
- `rollback`  
  Skip all release what contains `LOCK` file.
- `cleanup`  
  Delete all release what contains `LOCK` file (check if current not linked to it).

![](http://cdn.ebaumsworld.com/mediaFiles/picture/1989829/83031891.gif)

johnny-bit:
So... `LOCK` file is basically saying "we're in process of deploying here"? I quite get the concept for release, symlink & rollback (rolling to last complete release) however I don't get cleanup... If release acted like You said, then cleanup would have to be run twice - once befor release and once normally. maybe it would be better to add tasks like semaphores? `acquire_lock` and `release_lock`?

gordalina:
:+1:

elfet:
acquire_lock and release_lock - :+1:

Baldinof:
Is it necessary to have it on a per release basis ?

For my projects, I use these tasks, and it works good:
```php
env('lock_path', '{{deploy_path}}/deploy.lock');

task('deploy:lock', function() {
    $res = run('[ -f {{lock_path}} ] && echo Locked || echo OK');

    if (trim($res) === "Locked") {
        throw new RuntimeException("Deployement is locked.");
    }

    run('touch {{lock_path}}');
});

task('deploy:unlock', function() {
    $res = run('[ -f {{lock_path}} ] && echo Locked || echo OK');

    if (trim($res) === "Locked") {
        run('rm {{lock_path}}');
    }
});

after('deploy:prepare', 'deploy:lock');
after('deploy:symlink', 'deploy:unlock');
```

elfet:
Done in v4!

PR #714, opened 2016-06-26
Deployment lock

RicardoJ:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #354

This purposes the usage of a deployment lock. I tried using it with a release path but got unexpected results. Using the `deploy_path` seems to be the right direction.


RicardoJ:
To be done: `rollback` and `cleanup `.

PR #717, opened 2016-07-14
Prevent rolling back to empty or newer release dir

Yogarine:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A


elfet:
I think we need better solution with releases list file.

PR #740, opened 2016-08-04
Listen runtime exception

rlacerda83:
| Q | A |
| --- | --- |
| Bug fix? | No |
| New feature? | Yes |
| BC breaks? | No |
| Deprecations? | No |
| Fixed tickets | N/A |

Adds support for listener fatal exceptions (runtime exceptions) the tasks.

How it works?

By default, the listener is disabled, to enable listeners do:

```
set ( 'enable_runtime_exception_listener', true);
```

Set a function to be executed when a fatal exception occurs:

```
$notificationService =  new NotificationService();
onRunTimeException (function () use ($notificationService) {
     $notificationService->notifySlack();
});
```

If the listener is enabled you can set the task if it will have a listener or not:

```
// Will not be monitored
task ('test', function () {
     run ('git status');
     writeln ('<info> OK </ info>');
})->listenRunTimeException(false);
```

I could not capture the tasks that have fatal error in parallel processing. Can someone help me?

**NOTE: Some tests broke, I will correct.**


src/Deployer.php
```diff
@@ -58,15 +59,28 @@ class Deployer
...
-     */
-    public function __construct(Application $console, Console\Input\InputInterface $input, Console\Output\OutputInterface $output)
-    {
+     * @param Emitter $emitter
+     */
+    public function __construct(
+        Application $console,
+        Console\Input\InputInterface $input,
+        Console\Output\OutputInterface $output,
+        Emitter $emitter
```
unclead: may better set emiiter via setter? because it is not a mandatory property. Deployer will still work as expected without emitter. Usually only required properties are passed via constructor.

oanhnn:
Why not use `symfony\event-dispatcher` package? This project used many Symfony components.

elfet:
I'm for **symfony\event-dispatcher** too. And why do you need `enable_runtime_exception_listener` option? Why to turn in on all time?

rlacerda83:
I will change to the event dispatcher to symfony! No problem!

I used the lib league because I already have experiences in other projects with her!

I thought about using the listener off because:

1 - I think most users will not use the feature. I could be wrong! .
Theoretically, the fact do not set the callback function would make the listener off! 
Hmmm ... So I can leave it on by default!

Think like this: imagine that I just want to use the listener on a task:
I could globally to set to off and set only on a task that I want to use the listener. What do you think?

In this case it would make more sense to keep the parameter, but to turn the listener globally off?
Ex: set ( 'disable_runtime_exception_listener', false);
This command would turn the listener off for all tasks except those that are configured to use the listener.

About sending the listerner to the construct, if we think the concept of leaving on by default, the best option I see is to send it in the construct. The listener processing is very light.

If the listener is not set in construct, listener verification will be required for all calls.

Ex: In seriesExecutor I can not have a listener ... In this case the listener will continue on the interface, I'll have to send it null or not. Following this concept I'll have to check if it null before calling the "emmit".

elfet:
Check 4.x branch. I just added event delegator.

rlacerda83:
OK @elfet ! I will follow its implementation!

rlacerda83:
Done!

change listener to symfony event dispatcher!

```
<?php

require_once 'recipe/common.php';

localserver('test')
->stage('staging');

env('deploy_path', '/var/www/deployer');

// By default the listener is on to all tasks!
// If set to true, the listener will be off for all tasks
//set('disable_exception_task_listener', true);

task('test', function () {
    run('ls');
    writeln('<info>ok</info>');
})->desc('Execute test');

// will dispatch error!
task('test2', function () {
    run('git statdus');
    writeln('<info>ok</info>');
})->desc('Execute test2');

// This task ignores the listener, even if it goes wrong, the event will not be called
task('teste3', function () {
    writeln('<info>teste 3</info>');
})->desc('Execute artisan migrate')
->listenRunTimeException(false);

//Within the exception the user has access to the task name and the generated exception

onRunTimeException(function($event) {
    $msg = sprintf(
        'Error on task %s. Message: %s',
        $event->getTaskName(),
        $event->getErrorMessage()
    );
    writeln($msg);
});

// main task
task('deploy', [
    'teste',
    'teste2',
    'teste3'
])->desc('Deploy project');
```

elfet:
Implemented in v4 with events.

dakota:
@elfet Is there any guide on how to implement it in v4 with events?

elfet:
Take a look at this https://github.com/deployphp/deployer/blob/master/recipe/common.php#L146
You can do something like this:
```php
after('deploy:failed', 'your_task');
```

Issue #752, opened 2016-08-17
Add ability to merge with set, rather than overwrite


garygreen:
When using `set('shared_dirs', ['my_custom_dir']);` it would be nice to have the ability to merge with the existing values, rather than overwrite.

At the moment I have to do `set('shared_dirs', array_merge(get('shared_dirs'), ['my_custom_dir']))`. which seems messy and adds a lot of bloat to my deploy file.

Maybe a `merge('shared_dirs', ['my_custom_dir'])` option for the collection?

elfet:
I think creating another function for merging is bad — there is already to match functions.

What about this?

```php
$shared = get('shared_dirs');
array_push($shared, 'new', 'shared');
set('shared_dirs',  $shared);
```

garygreen:
Too verbose for me. It's pretty common to want to add additional writable dirs, shared dirs, etc rather than simply replacing the existing ones the recipe defines? Surprised a function for merging data doesn't exist...

elfet:
I think we can add some api function for what.

oanhnn:
@garygreen @elfet 
What are you think about case:
```
// recipe abc
set('shared_dirs', ['foo', 'bar']);
```
And in `deploy.php` file, i only want share `bar` and `baz` directory.
Are we need a function to this case?
I think make a new function, new API is simple but it isn't need.
This make deployer will be complex. Especially when the document is unclear!

garygreen:
@oanhnn 

> i only want to share `bar` and `baz` directory

Your last example does exactly that. I'm not sure what point your making?

elfet:
What about this name instead of `merge`?

```php
add('shared_dirs', ['path/to/dir']);
```
```

**Comment by elfet** [comment]
Or more fun way:

```php
set('shared_dirs', function ($dirs) {
    $dirs[] = 'path/to/dir';
    return $dirs;
});
```

garygreen:
I like `add` that sounds good. The closure function is pretty cool but can't see myself using it tbh.

elfet:
Done in v4 with `add`.

PR #753, opened 2016-08-17
Added ability to add values.

garygreen:
| Q | A |
| --- | --- |
| Bug fix? | No |
| New feature? | Yes |
| BC breaks? | No |
| Deprecations? | No |
| Fixed tickets | #752 |

This makes it easier to add values into existing recipes rather than overwriting the values.

### DIFF
--- a/src/Collection/Collection.php
+++ b/src/Collection/Collection.php
@@ -47,6 +47,14 @@ class Collection implements CollectionInterface, \Countable
     /**
      * {@inheritdoc}
      */
+    public function add($name, $object)
+    {
+        $this->collection[$name] = array_merge($this->collection[$name], $object);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     public function getIterator()
     {
         return new \ArrayIterator($this->collection);
--- a/src/functions.php
+++ b/src/functions.php
@@ -88,9 +88,9 @@ function localServer($name)
 function cluster($name, $nodes, $port = 22)
 {
     $deployer = Deployer::get();
-    
+
     $cluster = ClusterFactory::create($deployer, $name, $nodes, $port);
-    
+
     return $cluster->getBuilder();
 }
 
@@ -296,7 +296,7 @@ function runLocally($command, $timeout = 60)
     if (!$process->isSuccessful()) {
         throw new \RuntimeException($process->getErrorOutput());
     }
-    
+
     return new Result($process->getOutput());
 }
 
@@ -379,6 +379,15 @@ function write($message)
 function set($key, $value)
 {
     Deployer::get()->parameters->set($key, $value);
+}
+
+/**
+ * @param string $key
+ * @param mixed $value
+ */
+function add($key, $value)
+{
+    Deployer::get()->parameters->add($key, $value);
 }
 
 /**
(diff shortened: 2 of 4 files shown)


oanhnn:
I think make a new function, new API is simple but it isn't need.
This make deployer will be complex. Especially when the document is unclear!
Using `set` and `get` method is enough.

garygreen:
> Using `set` and `get` method is enough.

It may be enough, but it's not as nice. By default `set` overrides any existing values that the recipe defined. It's not particularly nice at the moment to 'add to' data for an existing recipe, which is pretty common imo. This doesn't break anything -- it just adds a more 'friendly' way of merging with the existing values.

I think we should encourage to keep configs as simple as possible and if by adding another method reduces unnecessary boilerplate code and makes the deploy config easier to read and understand, I think it's worthwhile.

Another option (to avoid adding another method) could be to have a third parameter in `set` to show as 'merge values' 

```php
set('shared_dirs', ['add_this_dir'], true); // true means will merge with the existing shared_dirs values.
```

To me this isn't as nice or obvious as doing `merge` though.

oanhnn:
I think it is not necessary.
```
// way 1
set('shared_dirs', [
    'dir_from_recipe01',
    'dir_from_recipe02',
    'your_custom_dir',
])
```
```
// way 2
merge('shared_dirs', ['your_custom_dir']);
````
````
// way 3
set('shared_dirs', ['your_custom_dir'], true);
```
I think use way 1 instead of way 2 or 3, your deploy script will be clearer.
A other problem is merge method can error when parameter isn't array
```
// case 1
set('foo', function () {
     return ['bar'];
});
merge('foo', ['baz']);
```
or
```
// case 2
set('foo', ['bar']);
merge('foo', 'baz');
```

elfet:
Take a look at capistrano: http://capistranorb.com/documentation/getting-started/configuration/
They using `append`. I think we can use it too.

garygreen:
> I think use way 1 instead of way 2 or 3, your deploy script will be clearer.

Sorry but I disagree. What your suggesting is to duplicate what the recipe defines into your own one -- what if the recipe updates or fixes something? Your have the old version still and will need to update your one. This is solved by doing `set('shared_dirs', array_merge(get('shared_dirs'), ['my_custom_dir']))` but that's too word-y and not very easy/nice to read.

> A other problem is merge method can error when parameter isn't array

Yes that's a problem. But it would throw an exception and still be pretty obvious why your deploy script is invalid. Also you could say the same about other functions -- for instance I thought I could do `after('deploy', ['task1', 'task2'])` but that doesn't work because it doesn't expect parameter 2 to be an array. 

Capistrano is a good example of having `append` which is pretty much what is being proposed here.

elfet:
`append('shared_dirs', 'new/path', ...);` or `append('shared_dirs', ['new/path', ...]);`

Also append func can throw exception if trying to append not array.

garygreen:
@elfet looks good. Should I change the PR to use `append` or `add` ?

elfet:
I think `add` is good:

```php
add('shared_dirs', ['new/path', ...]);
```

elfet:
Also please, create same PR for 4.x branch.

garygreen:
I've updated the PR. Will do one for 4.x branch

src/Collection/Collection.php
```diff
@@ -47,6 +47,14 @@ public function set($name, $object)
     /**
      * {@inheritdoc}
```
elfet: Inherent from what?

elfet:
Done in v4.

PR #760, opened 2016-08-22
Added artisan:optimize command and automatically optimize Laravel deployments

garygreen:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

This will add a `artisan:optimize` command and also automatically optimize deployments. Laravel has an internal list of classes which the framework will add in one file to optimize performance - this is separate from composer's optimize autoloader.

### DIFF
--- a/recipe/laravel.php
+++ b/recipe/laravel.php
@@ -66,6 +66,11 @@ task('artisan:route:cache', function () {
     writeln('<info>' . $output . '</info>');
 })->desc('Execute artisan route:cache');
 
+task('artisan:optimize', function () {
+    $output = run('{{bin/php}} {{deploy_path}}/current/artisan optimize --force');
+    writeln('<info>' . $output . '</info>');
+})->desc('Execute artisan optimize');
+
 /**
  * Task deploy:public_disk support the public disk.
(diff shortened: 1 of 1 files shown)


elfet:
Which laravel version?

garygreen:
> Which laravel version?

Laravel has always had the optimize command.

garygreen:
Suprisingly it's not very well documented, never has been but it's been there since 4.*. Here's the command as found in 5.1: https://github.com/laravel/framework/blob/5.1/src/Illuminate/Foundation/Console/OptimizeCommand.php#L19

It's recommended to run this when you deploy as it makes a big difference for performance.

pdbreen:
Just FYI - this has been a composer post install / post update script since at least v5.0 (didn't check earlier).  

https://github.com/laravel/laravel/blob/master/composer.json

So, explicit execution may not be needed and could actually cause issues if run after the symlink swap takes place.

garygreen:
@pdbreen that's true, on second thoughts for default Laravel installs this PR isn't really needed.

PR #761, opened 2016-08-22
Show which file is being uploaded when in debug mode

peschee:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

Currently, there's no way of actually seeing which files are being uploaded to the server. This PR adds the filename display of the file being uploaded when in debug mode (`-vvv`).

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -334,7 +334,7 @@ function runLocally($command, $timeout = 60)
     if (!$process->isSuccessful()) {
         throw new \RuntimeException($process->getErrorOutput());
     }
-    
+
     return new Result($process->getOutput());
 }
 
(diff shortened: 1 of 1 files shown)


oanhnn:
I see a out in line 320 
https://github.com/deployphp/deployer/blob/master/src/functions.php#L320

Is it not enough?

peschee:
Well, no, that only shows what path is being uploaded to where. However, if you want to see the progress and output of each single file and directory that is being uploaded, that's not enough. What do you think?

elfet:
I think it's okay to have this in debug mode.

elfet:
@deployphp/maintainer what you think?

oanhnn:
Ok. I am agree.

peschee:
Thanks!

PR #770, opened 2016-09-08
Add option to set branch to deploy

osrecio:
| Q | A |
| --- | --- |
| Bug fix? | No |
| New feature? | Yes |
| BC breaks? | No |
| Deprecations? | No |
| Fixed tickets | N/A |

Add option --branch to set branch to deploy. Best way than get of env.

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -73,6 +73,7 @@ env('bin/composer', function () {
 argument('stage', \Symfony\Component\Console\Input\InputArgument::OPTIONAL, 'Run tasks only on this server or group of servers.');
 option('tag', null, \Symfony\Component\Console\Input\InputOption::VALUE_OPTIONAL, 'Tag to deploy.');
 option('revision', null, \Symfony\Component\Console\Input\InputOption::VALUE_OPTIONAL, 'Revision to deploy.');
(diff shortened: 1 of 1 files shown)


pdbreen:
This looks like it would be a breaking change.  My server definitions set the branch via env.  While maybe not obvious at first, this makes perfect sense as I suspect many existing server definitions are environment specific (the documentation certainly guides people to use different server def for prod vs beta, etc).

oanhnn:
Branch should be defined in environment variables and `--tag` option should be 
more preferred than `branch`.
If you want deploy on different branch, i think you should make more stages.

elfet:
I like `--branch` options and i think we can add it in no BC.

elfet:
Better to 4.x

osrecio:
I agree wit you @pdbreen . I think the purpose is only add the option `--branch` without delete env('branch'). But If you make `dep deploy --branch=x` wins over env=branch of server because you can decide in deploy what branch you want deploy IMHO.

osrecio:
Ok I will don't change my PR beacuse @oanhnn make a PR with correct code of my point of view: https://github.com/deployphp/deployer/pull/771 

Thanks!

elfet:
Done in v4.

PR #783, opened 2016-09-21
fix bug recipe common deploy:writable httpUser run ps user to user:20

sdhou:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  |  No
| BC breaks?    |  No
| Deprecations? |  No
| Fixed tickets | N/A

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -280,7 +280,7 @@ task('deploy:writable', function () {
     if (!empty($dirs)) {
         try {
             if (null === $httpUser) {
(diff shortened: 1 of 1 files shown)


elfet:
Hi, what is this?

jessiejia:
When i have a user named  gitlab-www for nginx , the `ps axo user,comm` will get 'gitlab-+'. 
This will lead to the failure of the following tasks

elfet:
So use can simple set http_user param?

jessiejia:
yes, http_user is useful for that .
but i think this one is more cheap .
besides that ,  `http_user` is unsetable  in current code

elfet:
How is unsettable? You can use set.

jessiejia:
Yes, it works.

PR #790, opened 2016-09-23
Upgrade symfony recipe and template

fchris82:
| Q | A |
| --- | --- |
| Bug fix? | No |
| New feature? | Yes |
| BC breaks? | No |
| Deprecations? | No |
| Fixed tickets | N/A |

Add new configurations options:
- `removable_files` (default `['web/app_*.php', 'web/config.php']`)
- `interaction` (default `true`)
- `enable_database_create` (default `false`)
- `use_database_migration_strategy` (default `false`)
- `doctrine_migration_path` (default `app/DoctrineMigrations`)

New or modified tasks:
- `database:create`
- `database:migrate`
- `database:migrate:rollback` - rollback the database if the migration strategy is enabled
- `deploy:init-parameters-yml` - ask the new parameters (use the `composer.json` and symulate the composer behavior)

+Upgrade the template


recipe/symfony.php
```diff
@@ -107,26 +120,152 @@
...
+before('rollback', 'database:migrate:rollback');
 
 /**
  * Remove app_dev.php files
  */
 task('deploy:clear_controllers', function () {
-    run('rm -f {{release_path}}/web/app_*.php');
-    run('rm -f {{release_path}}/web/config.php');
+    foreach (get('removable_files') as $file) {
+        run('rm -f {{release_path}}/' . $file);
```
fchris82: Comment for me: here is a bug. Remove the `/` sign.

recipe/symfony.php
```diff
@@ -107,26 +120,152 @@
...
-})->desc('Migrate database');
+    if (env('use_database_migration_strategy')) {
+        run('{{env_vars}} {{bin/php}} {{bin/console}} doctrine:migrations:migrate {{console_options}}');
+    }
+})->desc('Migrate database (only if set `use_database_migration_strategy` true)');
+
+/**
+ * Database migration rollback
+ */
+task('database:migrate:rollback', function () {
```
elfet: This is a really complicated task. There is no way to do it simple? 
This code reley on a lot of stuff what can be changed, and can bring a lot of bugs. 
But i like a work. Maybe it's better to create separate project for this task where it's possible to write tests for this?

PR #803, opened 2016-10-05
Use git_cache also when --revision is defined

qtsd:
| Q | A |
| --- | --- |
| Bug fix? | Yes |
| New feature? | No |
| BC breaks? | No |
| Deprecations? | No |
| Fixed tickets | N/A |

When `--revision` option is setted, `update_code` does not take care of `git_cache`.
There is a comment "_To checkout specified revision we need to clone all tree_".

But if `git_cache` is used, all the tree is already cloned.
Else we just need to not shallow clone (setting `$depth=''`).

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -193,15 +193,14 @@ task('deploy:update_code', function () {
     // If option `tag` is not set and option `revision` is set
     if (empty($tag) && input()->hasOption('revision')) {
         $revision = input()->getOption('revision');
+        if (!empty($revision)) {
+            $depth = '';
+        }
     }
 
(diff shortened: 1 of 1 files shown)


elfet:
Cool! Need more time to test and review. 
Pls, take a look at this PR @deployphp/maintainer

elfet:
@qtsd what if first deploy was with `--depth 1`, then second with `--revision` option?

qtsd:
@elfet it throws an exception, which do a complete clone when catched.

qtsd:
(same comportment as gitcache)

elfet:
Cool!

PR #806, opened 2016-10-11
Don't use /current in Laravel recipe

vicdelfant:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A or xx

The Laravel recipe assumes the existence of the `current` directory, but this is not the case on the very first release:

```
[RuntimeException]
  Could not open input file: foo/current/artisan
```

This PR fixes this issue by using `{{release_path}}` and using an inline `if -d` for the`artisan:up` and `artisan:down` tasks.

### DIFF
--- a/recipe/laravel.php
+++ b/recipe/laravel.php
@@ -22,47 +22,47 @@ set('writable_dirs', ['bootstrap/cache', 'storage']);
  * Helper tasks
  */
 task('artisan:up', function () {
-    $output = run('{{bin/php}} {{deploy_path}}/current/artisan up');
-    writeln('<info>'.$output.'</info>');
+    $output = run('if [ -d {{deploy_path}}/current ]; then {{bin/php}} {{deploy_path}}/current/artisan up; fi');
+    writeln('<info>' . $output . '</info>');
 })->desc('Disable maintenance mode');
 
 task('artisan:down', function () {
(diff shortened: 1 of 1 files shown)


oanhnn:
:+1: Thank you for this case, but `{{deploy_path}}/current` is a symlink.

vicdelfant:
You're welcome :) I am aware that current is a symlink, but the symlink doesn't exist until after the first successful release. It isn't uncommon to run any of the artisan tasks before the symlink task, so using the actual release path will always work.

recipe/laravel.php
```diff
@@ -22,47 +22,47 @@
  * Helper tasks
  */
 task('artisan:up', function () {
-    $output = run('{{bin/php}} {{deploy_path}}/current/artisan up');
-    writeln('<info>'.$output.'</info>');
+    $output = run('if [ -d {{deploy_path}}/current ]; then {{bin/php}} {{deploy_path}}/current/artisan up; fi');
```
oanhnn: Why not?
```
run('if [ -f {{deploy_path}}/current/artisan ]; then {{bin/php}} {{deploy_path}}/current/artisan up; fi');
```

vicdelfant: Sure, that'd work too. Commit coming up...

PR #835, opened 2016-11-03
[WIP] Allow hiding of secrets in deployer log output

x3ro:
| Q             | A
| ------------- | ---
| Bug fix?      |  No
| New feature?  | Yes
| BC breaks?    |  No
| Deprecations? |  No
| Fixed tickets | N/A 

I'm currently building a continuous deployment system using GitLab CI and deployer. Being able to deploy from within GitLab CI requires me to provide it with some secrets, such as the password of the server I'm going to deploy to (since that shouldn't be part of the git repository). I can create such secrets in GitLab, and they will not be visible to common users, which is good. 

For being able to check deployment failures in the build logs, I'm always running the deployment with `-vv`. As a result, when running a command which contains a secret, it is echoed to the build log in plain text. I need to avoid this, since build logs are visible to all users that can see the repository in GitLab. 

Because of this I've introduced the `[secret]foobar[/secret]` syntax, which will omit everything between the "tags" when printed to the console, but will still run the correct command when used with `run` or `runLocally`. 

Is this something that you (the maintainers) could see being merged into deployer? :)

PS: Thinking about it, I should've probably used `<secret></secret>` instead to conform with Symfony Console. I'll fix that, along with any other feedback there might be.

- [x] Change syntax to `<secret>`
- [ ] Figure out where to put the `ShellCommand` class (root namespace doesn't seem nice)
- [ ] Clean up `use`s in `ShellCommand.php`
- [x] Figure out why tests are not failing for me locally

### DIFF
--- a/src/ShellCommand.php
+++ b/src/ShellCommand.php
@@ -1,0 +1,45 @@
+<?php
+/* (c) Anton Medvedev <anton@medv.io>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Deployer;
+
+use Deployer\Console\InitCommand;
+use Deployer\Console\WorkerCommand;
+use Deployer\Console\Application;
+use Deployer\Server;
+use Deployer\Stage\StageStrategy;
+use Deployer\Task;
+use Deployer\Collection;
+use Deployer\Console\TaskCommand;
+use Symfony\Component\Console;
+
+class ShellCommand
+{
+    private $command;
+
+    public function __construct($command)
+    {
+        $this->command = $command;
+    }
+
+    public function getForRunning()
+    {
+        $command = str_replace("<secret>", "", $this->command);
+        $command = str_replace("</secret>", "", $command);
+        return $command;
+    }
+
+    public function __toString()
+    {
+        return $this->getForRunning();
+    }
+
+    public function getForPrinting()
+    {
+        return preg_replace('|<secret>.*</secret>|s', "[SECRET HIDDEN]", $this->command);
+    }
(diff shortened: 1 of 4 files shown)


x3ro:
If I could get feedback on

- [ ] Figure out where to put the ShellCommand class (root namespace doesn't seem nice)

That'd be great :) Not really sure where to put it!

src/ShellCommand.php
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    public function __toString()
+    {
+        return $this->getForRunning();
+    }
+
+    public function getForPrinting()
+    {
+        return preg_replace('|<secret>.*</secret>|s', "[SECRET HIDDEN]", $this->command);
```
glensc: use non-greedy matchin for cases there are multiple `<secret>`s on the same line:

```php
'|<secret>.*?</secret>|s'
```

test/src/ShellCommandTest.php
```diff
@@ -0,0 +1,51 @@
+<?php
+namespace Deployer;
+
+class ShellCommandTest extends \PHPUnit_Framework_TestCase
+{
+    protected function setUp()
+    {
+        $this->command = new ShellCommand("echo '<secret>important\nsecret</secret>' > /dev/null");
+        $this->emptySecretCommand = new ShellCommand("echo '<secret></secret>' > /dev/null");
```
glensc: add test for multiple `<secret>` on same line!

pluseg:
@x3ro Lucas, could you update your PR with comments above please?

antonmedv:
Do work on this feature. Closing it.

PR #836, opened 2016-11-03
Don't stop the deploy process when zero migrations

mathieu-gilloots:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | ?
| Deprecations? | No
| Fixed tickets | N/A

When i want to keep a "database:migrate" on my deploy file and when I have no migration in DoctrineMigrations Folder, cause break of the deploy.


elfet:
Thanks!

Issue #791, opened 2016-09-23
before()/after() wildchar or beforeAll()/afterAll()


luke83:
| Q | A |
| --- | --- |
| Issue Type | Feature Request |
| Deployer Version | N/A |
| Local Machine OS | N/A |
| Remote Machine OS | N/A |
### Description

_I think it could be useful to have a `catch 'em all` option for after()/before() to run specific task before/after all other task. This could be useful in use cases like #777._
### Proposed interface:

```
before('*', 'timer:start');
after('*', 'timer:stop');
-or-
beforeAll('timer:start');
afterAll('timer:stop');
```

gulios:
that's right and for parallel too because now when i do some task after('count_time') i get the longest time for server node(because task ends when all servers done) not for each server. For statistics usage it could be very useful to get information about: task_name, execution_time, server...

thx

pluseg:
@luke83 i think that beforeAll and afterAll can lead to breaking of callback execution order. It will become less obvious. According to #777 it's easier to add profiling feature in verbose mode.

elfet:
I think it's better to use event system here.

elfet:
Events in v4.

radutopala:
@elfet is the event system already implemented in 4.1?

elfet:
No, not in v4. Probably will be in 4.1. Or maybe will back to wildchar with *.

PR #893, opened 2016-11-22
added onStart and onEnd tasks

radutopala:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #791 

A quick fix for added on start and on end tasks, similar to beforeAll / afterAll from #791.

@elfet pls close this PR if you have a better way soon. Thanks!

### DIFF
--- a/src/Console/TaskCommand.php
+++ b/src/Console/TaskCommand.php
@@ -61,7 +61,15 @@ class TaskCommand extends Command
     {
         $stage = $input->hasArgument('stage') ? $input->getArgument('stage') : null;
 
-        $tasks = $this->deployer->getScriptManager()->getTasks($this->getName(), $stage);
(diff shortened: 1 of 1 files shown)


elfet:
I think this is not solution for #791. Now working of this by myself. But need more time.

radutopala:
I needed a fast solution to a problem I have here, but happy to use yours when it's done.

radutopala:
@antonmedv was this issue addressed in any of the newer versions of deployer?

antonmedv:
Actually we are using simple before/after as this has same api. 

Main idea here is to try to keep deployer simple, and allow user to do anything inside deploy.php

PR #904, opened 2016-11-25
Adjusted runLocally to throw ProcessFailedException in case of error

maurice2k:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

Replaced the `\RuntimeException` in `runLocally` with Symfony's `Symfony\Component\Process\Exception\ProcessFailedException` which has more information on the command that failed (including command line and error output).

The `ProcessFailedException` extends the `\RuntimeException` so there shouldn't be a BC problem.


elfet:
Cool, thanks!

Issue #713, opened 2016-06-23
Magento2 Recepipe


barryvdh:
| Q | A |
| --- | --- |
| Issue Type | Feature Request |

There is a Magento1 recipe, should we start working on Magento2?
- Works with Composer
- uses bin/magento binary file for commands
- needs to share files/folders

Flush cache

``` php
task('magento:cache:flush', function () {
    $output = run('{{bin/php}} {{deploy_path}}/current/bin/magento cache:flush');
    writeln('<info>' . $output . '</info>');
})->desc('Flush cache');
```

Generate static content

``` php
task('magento:static-content', function () {
    $output = run('{{bin/php}} {{deploy_path}}/current/bin/magento setup:static-content:deploy nl_NL');
    writeln('<info>' . $output . '</info>');
})->desc('Clear cache');
```

Needs a first-time install script, that copies the env.php/config.php? How do you usually handle that?

``` php
task('install', [
    'deploy:prepare',
    'deploy:release',
    'deploy:update_code',
    'deploy:vendors',
    'deploy:symlink',
    'magento:install',
]);

env('magento_install_options',
'--db-host=localhost --db-name=magento --db-user=magento ' .
'--admin-firstname=Magento --admin-lastname=User --admin-email=admin@example.com ' .
'--admin-user=admin --admin-password=admin123 --language=nl_NL '.
'--currency=EUR --timezone=Europe/Amsterdam --cleanup-database '.
'--session-save=db --use-rewrites=1');

task('magento:install', function() {
    $confirm = askConfirmation('Are you sure? This will (re-)install Magento!');
    if (!$confirm) {
        writeln('<comment>Aborted</comment>');
        return 0;
    }
    $dbPass = ask('What is the DB password?','');
    $output = run('{{bin/php}} {{deploy_path}}/current/bin/magento setup:install --base-url=http://'.env('server.host').' --db-password="'. $dbPass.'" {{magento_install_options}}');
    writeln('<info>' . $output . '</info>');

  //Not sure what this should be exactly
   // run("cp {{release_path}}/app/etc/env.php {{deploy_path}}/shared/app/etc/env.php");
    // run("cp {{release_path}}/app/etc/config.php {{deploy_path}}/shared/app/etc/config.php");

});
```

Deploy

``` php
task('deploy', [
    'deploy:prepare',
    'deploy:release',
    'deploy:update_code',
    'deploy:shared',
    'deploy:vendors',
    'magento:static-content',
    'deploy:symlink',
[...]

PR #911, opened 2016-11-28
add magento2 base recipe

unailopez16:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #713

### DIFF
--- a/recipe/magento2.php
+++ b/recipe/magento2.php
@@ -1,0 +1,80 @@
+<?php
+namespace Deployer;
+
+require_once __DIR__ . '/common.php';
+
+// Configuration
+set('shared_files', ['app/etc/env.php', 'var/.maintenance.ip']);
+set('shared_dirs', ['var/log', 'var/backups', 'pub/media']);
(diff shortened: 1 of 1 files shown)


peterjaap:
I did a quick review @unailopez16 

I don't think enabling all modules in the deployment recipe is a good idea - I might have a good reason (or a not so good reason) to disable an extension but still deploy the code.

Also, enabling/disabling maintenance mode during deployment isn't necessary when there is no database upgrade. I'd recommend moving this to an `askConfirmation` that defaults to true in silent mode.

You might also want to add an option to set a parameter in some config file for which theme and/or language to deploy the static content for. You're now deploying ALL static content, which takes a considerable amount of time for M2. When we're building a Dutch shop, I don't need to deploy other languages, so adding `nl_NL` to the static content deploy command considerably speeds things up. This goes for `--theme` as well.

suspiciousfellow:
Thanks for this, was looking for something like it.

PR #912, opened 2016-11-29
Set the correct port parameter to ssh/scp command

stz184:
| Q             | A |
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

According to man scp -P (uppercase) is the correct parameter to set custom port number.
-p (lowercase) is valid scp parameter and according to the man page "Preserves modification times, access times, and modes from the original file."

The problem was discovered while transferring a file from /tmp directory, because the temp files are usually created with 0600 permission which is preserved on the remote machine. Without changing the ownership or permissions. it will not be readable by www-data user.

### DIFF
--- a/src/Server/Remote/NativeSsh.php
+++ b/src/Server/Remote/NativeSsh.php
@@ -58,7 +58,7 @@ class NativeSsh implements ServerInterface
         $hostname = $serverConfig->getHost();
 
         if ($serverConfig->getPort()) {
-            $sshOptions[] = '-p ' . escapeshellarg($serverConfig->getPort());
+            $sshOptions[] = '-P ' . escapeshellarg($serverConfig->getPort());
         }
 
         if ($serverConfig->getPrivateKey()) {


t33k:
Could you show me the link to man page of ssh you reffering to?

What I found is:

In man of ssh port is set with "-p"
https://linux.die.net/man/1/ssh

In man of scp port is set with "-P"
https://linux.die.net/man/1/scp

So current version is ok and your change is wrong.

stz184:
Hi, t33k,

The current implementation is wrong, because you merge both ssh and scp parameters and pass it to scp command, used inside the scpCopy method and so inside upload one.
That's way you effectively DO NOT change the port for scp transfers, instead, you change the function behavior. -p (lowercase) modification times, access times, and modes from the original file, which is not always the desired behavior. Example usage: I generate config files using .dist files for template and put them in /tmp directory from which I then upload it to the server. The files are transferred with 0600 file permissions, which is not the desired behavior. And most important, passing -p instead of -P to scp, you DO NOT change the port number, so I do not think that this is OK.

t33k:
hi stz184,

could you show me place in code where there is "merge both ssh and scp parameters" ?

For my understanding of code in src/Server/Remote/NativeSsh.php there is separate options for ssh and separate for scp. There is no object property that holds ssh options and there is no passing of ssh options as method parameters.

stz184:
Hi t33k, 

You are right, I double checked this. 
I am closing this one,

PR #918, opened 2016-12-03
Add support for multiplexing for NativeSsh

t33k:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

Implementing ssh multiplexing speeds up NativeSsh extremely. Its faster than other methods.
https://en.wikibooks.org/wiki/OpenSSH/Cookbook/Multiplexing

How to test? In your deploy.php file put:
```
   set('ssh_type', 'native');
   set('ssh_type_native_mux', true);
```

You should observe considerable increase of speed. When using mux the ssh authentication is done only once.

### DIFF
--- a/src/Server/Remote/NativeSsh.php
+++ b/src/Server/Remote/NativeSsh.php
@@ -13,6 +13,8 @@ use Symfony\Component\Process\Process;
 
 class NativeSsh implements ServerInterface
 {
+    const UNIX_SOCKET_MAX_LENGTH = 104;
+
     /**
      * @var array
      */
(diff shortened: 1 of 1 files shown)


elfet:
Awesome stuff!

elfet:
Maybe it's better to turn on it by default?

t33k:
Tnx elfet! I was also amazed by the speed results :)

For "default". In my understanding multiplexing can be received as unsafe for some folks because it creates linux socket that can be then accessed without any authorization. The mux in the proposed implementation creates socket for time of deploy (+5 sec). So in the time of deploying someone who has access to our local machine from which we do deploy can use the socket to gain access to remote server.

The judgment if this is safe enough or not is probably individual perception.

So if this would be done by default then documentation must, imho, clearly state the fact that:
a) multiplexing is used by default,
b) multiplexing brings some security problems, 
c) multiplexing can be deactivated with set('ssh_type_native_mux', false);

Other than that I am for setting mux as default :)

elfet:
So socket can be accessed during deployment time? Then deploying from local machine there may be no security issue as keys for example available for everybody.

I'd like to make it by default (btw, what about Windows?) as a lot of users does not change default, so this can bring a speedup in deployment time. It will be good if we can find proper solution for this.

t33k:
Yes. Socket can be accessed during deployment time. 
Keys can be additionally protected with passphrase and then the mux is less secure.

For windows. Well - I do not know and I will have not way to test because I am on mac.

t33k:
hi @elfet 
is there anything I can do to push this forward ? Do you want to change this PR somehow? Plz tell me - I can make some changes.

elfet:
I'd like to test in on mac, linux and windows to make in default by default. Or write better docs with explanation of how tu turn in on.

Issue #922, opened 2016-12-09
release_path should return current_path in non-deploy context


elfet:
| Q                 | A
| ----------------- | ---
| Issue Type        | Feature Request
| Deployer Version  | next
| Local Machine OS  | N/A
| Remote Machine OS | N/A

### Description 
~~Implement `release_or_current_path` param which will be returning `release_path` or `current_path`.~~

`release_path` should return `current_path` in non-deploy context.

elfet:
I found is better solution: release_path should return current_path. 🤗

elfet:
Done.

Issue #921, opened 2016-12-09
Cannot use get function in task with once method


ozgurky:
| Q                 | A
| ----------------- | ---
| Issue Type        | Question
| Deployer Version  | 4.0.2
| Local Machine OS  | Ubuntu 16.04 LTS
| Remote Machine OS | Ubuntu 16.04 LTS

 I use a custom task for running artisan migrate on a lumen project and I want this task to run once. 

```
task('deploy:artisan_migrate', function () {
    run("cd {{release_path}} && php artisan migrate --force");
})->once();
```
It returns `Configuration parameter deploy_path does not exists.` error but this parameter is defined in yml file. I've also tried get('release_path') and got same error. When i remove once method it seems ok.

Is not possible to use **get** function in a task with once method?

elfet:
What do you think `get('release_path')` should return in `once()` task?

ozgurky:
Thanks for reply, I've multiple servers for production but they don't have any git installation. So i decided to run git on local server, compress codes with tar and upload tar file to these servers. I want to run git clone command only once for fast deployment. Here is my code:

```
task('local:update_code', function () {
    $repository = trim(get('repository'));
    $branch = get('branch');
    $git = get('bin/git');
    $gitCache = get('git_cache');
    $depth = $gitCache ? '' : '--depth 1';

    if (input()->hasOption('tag')) {
        $tag = input()->getOption('tag');
    } elseif (input()->hasOption('revision')) {
        $revision = input()->getOption('revision');
    }

    $at = '';
    if (!empty($tag)) {
        $at = "-b $tag";
    } elseif (!empty($branch)) {
        $at = "-b $branch";
    }

    if (!empty($revision)) {
        runLocally("$git clone $at --recursive -q $repository {{local_path}} 2>&1");
        runLocally("cd {{local_path}} && $git checkout $revision");
    } else {
        runLocally("$git clone $at $depth --recursive -q $repository {{local_path}} 2>&1");
    }
})->once()->desc('Updating local code');
```

branch parameter is defined in server.yml file but when i use task with once get('branch') returns empty.

Is there a way for using yml file parameters with once method?

elfet:
```yml
prod:
    ...
    branch: prod

beta:
    branch: beta
```


What should `gert('branch')` return?

ozgurky:
```
prod-b:
  ...
  stage: develop
  branch: develop

prod-c:
  stage: production
  branch: master
```
[...]

Issue #920, opened 2016-12-08
laravel migrate show error release path does not found


nyrf:
| Q                 | A
| ----------------- | ---
| Issue Type        | Bug
| Deployer Version  | 4.0.2
| Local Machine OS  | OS X 10.12.1
| Remote Machine OS | ubuntu 14.04

### Description 
After deploy, run `dep artisan:migrate`

### Output log
*  [RuntimeException]
  Release path does not found.
  Run deploy:release to create new release.*

elfet:
This task can be called only during deployment process:
```php
before('deploy:symlink', 'artisan:migrate');
```
Maybe it's should be able to run it separately.  @deployphp/maintainer ?

nyrf:
@elfet It works, thanks.

Issue #879, opened 2016-11-17
Parameter freezing step


player259:
| Q                 | A
| ----------------- | ---
| Issue Type        | Feature Request
| Deployer Version  | 4
| Local Machine OS  | N/A
| Remote Machine OS | N/A

### Description 
In case of strange environment parameters behavior (like #871) there should be declared parameter freezing step
Such unpredictable parameter evaluation can appear on task skipping or specific execution order
I'd suggest to use two-step deployment with preparing and execution as symfony container does (container compilation)
Symfony uses FrozenParameterBag (DependencyInjection component) to store parameters in runtime
And Deployer already has a part of described feature, Deployer::get()->config service, where parameters are defined before the real execution.

Other parameters, as "releases_list", should not be cached or freezed at all
After "cleanup", "deploy:release" tasks or other custom tasks, which do manipulations in releases folder, value of "releases_list" should be evaluated again

So parameter subsystem should became more transparent, predictive and determinate
By the way, i agreed with deprecating env()

And it will be great to use Config component to store servers.yml structure definition
Symfony\Component\Config\Definition\Processor
Symfony\Component\Config\Definition\Builder\TreeBuilder

elfet:
It will be cool if "releases_list" did not freezes. What about some parameter?
```php
set('releases_list`, function () {
}, true); // true - not freeze
```

Or some method for recalling callback?

```php
fetch('releases_list'); // Recall callback and update config param.
```

player259:
As some parameters are defined in third-party recipes, developer should not know how they are working.
Developer can miss the moment when parameter should be fetched again, or don't know that parameters were affected by his actions.
So, first solution seems more meanful.

elfet:
Or maybe something like:
```php
factory('releases_list', ...);
```

player259:
Yes, functions should stay small and simple. Another function to implement another behavior is the way to go.
But if you introduce factory() as runtime-evaluated parameter, set() should become once-evaluated non-runtime parameter, as opposite to factory(). And the expected way is on-the-start evaluation, just before the tasks.
I think it will be more clear for developers to have parameter initialization step. If keep thing as-is, it will be ambiguous and uncertain. In some cases there will be no difference what to use - factory() or set(), and this is not good.

elfet:
I think freeze all parameters before every task is unproductive. It's slow down deployment process. 
Now is more like lazy evaluating params, which is familiar to many programmers. So, i think we don't need freezing step.

[...]

PR #923, opened 2016-12-10
Unify paths

jordisala1991:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes 
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #922 #921 #920 #879 

I'm trying to unify all paths into a variables (similar to Capistrano). What do you think about this PR? I know it's not complete, and will require to change some things and some documentation maybe.

With this PR, one should be able to modify the structure of deployer using the new variables.

We can include new paths like: `current_release` or `previous_release` to simplify and make simple to use `release_path` outside a release.

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -131,7 +131,9 @@ option('branch', null, InputOption::VALUE_OPTIONAL, 'Branch to deploy');
  */
 desc('Show current release');
 task('current', function () {
-    writeln('Current release: ' . basename(realpath(get('current_path'))));
+    $current = run("readlink {{current_path}}")->toString();
+
+    writeln('Current release: ' . basename($current));
 });
 
 
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -69,12 +69,19 @@ set('git_cache', function () { //whether to use git cache - faster cloning by bo
 
 
 /**
- * Return current release path.
+ * Custom paths.
  */
-set('current_path', function () {
-    $link = run("readlink {{deploy_path}}/current")->toString();
-    return substr($link, 0, 1) === '/' ? $link : get('deploy_path') . '/' . $link;
-});
+set('current_dir', 'current');
+set('release_dir', 'release');
+set('releases_dir', 'releases');
+set('shared_dir', 'shared');
+set('dep_dir', '.dep');
+
+set('current_path', '{{deploy_path}}/{{current_dir}}');
+set('release_path', '{{deploy_path}}/{{release_dir}}');
+set('releases_path', '{{deploy_path}}/{{releases_dir}}');
+set('shared_path', '{{deploy_path}}/{{shared_dir}}');
+set('dep_path', '{{deploy_path}}/{{dep_dir}}');
 
 
 /**
@@ -124,7 +131,7 @@ option('branch', null, InputOption::VALUE_OPTIONAL, 'Branch to deploy');
  */
 desc('Show current release');
 task('current', function () {
-    writeln('Current release: ' . basename(get('current_path')));
+    writeln('Current release: ' . basename(realpath(get('current_path'))));
 });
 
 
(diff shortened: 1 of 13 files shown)


jordisala1991:
Ping @elfet. Does this have any change to get merged? I have another PR depending on this one.

elfet:
This is cool release. But what about this?

>  I know it's not complete, and will require to change some things and some documentation maybe.

Also what PR you want to?

> I have another PR depending on this one.

Also i'm wondering is this functionality is really needed? I mean how many people will be using it? Because this PR bring a lot of config option which lead to complicate thinks and refactoring.

For example _releases_path_ and _release_path_ differ only in one *s* letter which will lid to potential  bugs. I think good naming will solve problems.

recipe/common.php
```diff
@@ -69,12 +69,19 @@
...
-    $link = run("readlink {{deploy_path}}/current")->toString();
-    return substr($link, 0, 1) === '/' ? $link : get('deploy_path') . '/' . $link;
-});
+set('current_dir', 'current');
+set('release_dir', 'release');
+set('releases_dir', 'releases');
+set('shared_dir', 'shared');
+set('dep_dir', '.dep');
+
+set('current_path', '{{deploy_path}}/{{current_dir}}');
```
elfet: Now current_path will be returning symlink instead of real path with will lead to bugs. Also does not work with relative symlinks. 

Why did you drop this code?

```
 -set('current_path', function () {
 -    $link = run("readlink {{deploy_path}}/current")->toString();
 -    return substr($link, 0, 1) === '/' ? $link : get('deploy_path') . '/' . $link;
 -});
```

recipe/deploy/release.php
```diff
@@ -74,53 +72,34 @@
...
-    $releaseExists = run("if [ -h {{deploy_path}}/release ]; then echo 'true'; fi")->toBool();
-    if (!$releaseExists) {
-        throw new \RuntimeException(
-            "Release path does not found.\n" .
-            "Run deploy:release to create new release."
-        );
-    }
-
-    $link = run("readlink {{deploy_path}}/release")->toString();
-    return substr($link, 0, 1) === '/' ? $link : get('deploy_path') . '/' . $link;
```
elfet: Same thing here. Why you drop this code? Now it's does not support relative symlinks.

recipe/common.php
```diff
@@ -69,12 +69,19 @@
...
  */
-set('current_path', function () {
-    $link = run("readlink {{deploy_path}}/current")->toString();
-    return substr($link, 0, 1) === '/' ? $link : get('deploy_path') . '/' . $link;
-});
+set('current_dir', 'current');
+set('release_dir', 'release');
+set('releases_dir', 'releases');
+set('shared_dir', 'shared');
+set('dep_dir', '.dep');
```
elfet: Do you really need to modify dir name of `.dep`?

jordisala1991: The reason to create variables for all directories is to be consistent in the code using those directories, like using a variable to avoid repetition of a constant value.

recipe/common.php
jordisala1991: Why using a symlink will lead to bugs on the code? I am letting the OS to resolve the symlink instead of forcing a readlink.

recipe/deploy/release.php
jordisala1991: In this case, I wanted to change `release_dir`variable to `current_release` like capistrano, but I wanted to mantain the PR with not too much changes.

jordisala1991:
I have another PR which focus on moving all `if` on the code to the `test()` function. It was already there, but not used consistently on the code. But since it touched a lot of code too, didn't want to rebase a lot of times.

Capistrano variables structure: https://www.admon.org/scripts/capistrano-variables-list/
Capistrano code: https://github.com/capistrano/capistrano/blob/0965373a4b239665360e15e3ee409f130b6c094b/lib/capistrano/dsl/paths.rb

PR #934, opened 2016-12-23
Added --hook option

peterjaap:
Added --hook option to disable running other tasks than the one specified

| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

I wanted a way to run certain tasks without having the before/after hooks set in deploy.php to be run, because some of those tasks might assume we are in a deploy context while maybe we're not.

We can now do;

```bash
dep custom:command:here staging --hooks false
```

While we have

```php
before('custom:command:here', 'another:command:before:custom');
after('custom:command:here', 'another:command:after:custom');
```

So not to run `another:command:after:custom` or `another:command:before:custom` but still run `custom:command:here`.

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -118,6 +118,7 @@ argument('stage', InputArgument::OPTIONAL, 'Run tasks only on this server or gro
 option('tag', null, InputOption::VALUE_OPTIONAL, 'Tag to deploy');
 option('revision', null, InputOption::VALUE_OPTIONAL, 'Revision to deploy');
 option('branch', null, InputOption::VALUE_OPTIONAL, 'Branch to deploy');
+option('hooks', null, InputOption::VALUE_OPTIONAL, 'Set to false to skip after/before hooks', true);
 
 /**
  * Tasks
--- a/src/Console/TaskCommand.php
+++ b/src/Console/TaskCommand.php
@@ -63,7 +63,7 @@ class TaskCommand extends Command
 
         $tasks = $this->deployer->getScriptManager()->getTasks($this->getName(), $stage);
 
-        if (!$input->getOption('hooks') || $input->getOption('hooks') == 'false') {
+        if (!$input->getOption('hooks') || $input->getOption('hooks') === 'false') {
             $tasks = array_filter($tasks, function ($task) {
                 return $this->getName() == $task->getName();
             });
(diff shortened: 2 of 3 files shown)


peterjaap:
Tests are failing, though I don't understand why.

elfet:
I think test failing in parallel tests. Try fix it myself.

elfet:
@peterjaap also i think test for this options will be cool.

peterjaap:
The test actually revealed a bug in the code, heh, who would've thought. All checks passing!

elfet:
What about renaming to --no-hooks?

peterjaap:
Yes, perfect.

Op do 19 jan. 2017 om 05:41 schreef Anton Medvedev <notifications@github.com

src/Console/TaskCommand.php
```diff
@@ -62,6 +62,13 @@ protected function execute(Input $input, Output $output)
         $stage = $input->hasArgument('stage') ? $input->getArgument('stage') : null;
 
         $tasks = $this->deployer->getScriptManager()->getTasks($this->getName(), $stage);
+
+        if (!$input->getOption('hooks') || $input->getOption('hooks') === 'false') {
```
elfet: But what about group task? Them will not be working like this :/

recipe/common.php
```diff
@@ -118,6 +118,7 @@
 option('tag', null, InputOption::VALUE_OPTIONAL, 'Tag to deploy');
 option('revision', null, InputOption::VALUE_OPTIONAL, 'Revision to deploy');
 option('branch', null, InputOption::VALUE_OPTIONAL, 'Branch to deploy');
+option('hooks', null, InputOption::VALUE_OPTIONAL, 'Set to false to skip after/before hooks', true);
```
pluseg: Need to change to `no-hooks`. Also please set $mode = InputOption::VALUE_NONE and remove default value. Then in TaskCommand:66 you can use just one simple condition `if ($input->getOption('no-hooks')) {`.

src/Console/TaskCommand.php
```diff
@@ -62,6 +62,13 @@ protected function execute(Input $input, Output $output)
...
+            $tasks = array_filter($tasks, function ($task) {
+                return $this->getName() == $task->getName();
+            });
```
pluseg: @peterjaap I guess you should implement `ScriptManager::setHooksEnabled(true|false)` and use the `$isHooksEnabled` flag in `ScriptManager::getTasks()` method. It will work both for single and group tasks.

Also nice to have tests for that. Check test/src/FunctionsTest.php with pretty same ones.

Peter, could you do that? Do you have a free time?

pluseg:
@elfet I've rewritten this PR and added tests. Please close this ticket and review https://github.com/deployphp/deployer/pull/1061.

elfet:
Closing as duplicate of #1061 (this pr fixes group scripts)

Issue #929, opened 2016-12-19
Why  deployer/phar-update as dependency?


ceesvanegmond:
| Q                 | A
| ----------------- | ---
| Issue Type        |Question
| Deployer Version  | 4.0.2
| Local Machine OS  | Mac Sierra
| Remote Machine OS | Ubuntu 16

### Description 
Just a question, why is deployer/phar-update a dependency of this specific project? For instance when using a other framework than Symfony, you might not need phar-update. 

Is there an specific reason for this?

elfet:
It's used only when deployer is a phar.

ceesvanegmond:
But why is it a 'requirement'? I think it needs to be an suggestion instead of a require. Don't you think? In this case, I got installation errors, because of an interfering package  ```justinrainbow/json-schema``` because this is used by ```herrera-io/json``` which is ultimatly used by ```deployer/phar-update```.

Don't you think it's better to suggest instead of require this ```phar-update``` package? It is optional though.

elfet:
Yes, i think it's better to move it to suggests. Will do it in next release. For now you can use phar version.

PR #937, opened 2016-12-28
#929: Moved deployer/phar-update to suggest instead of raw dependency

ceesvanegmond:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #929

This small pullrequest moves deployer/phar-update to the suggest. See ticket #929

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,6 @@
         "symfony/yaml": "~2.6|~3.0",
         "symfony/process": "~2.6|~3.0",
         "phpseclib/phpseclib": "~2.0",
-        "deployer/phar-update": "^1.0",
         "pimple/pimple": "~3.0",
         "monolog/monolog": "^1.21"
     },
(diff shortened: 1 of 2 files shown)


ceesvanegmond:
@elfet The build is failing because of the use of ```Deployer\Component\PharUpdate\Console\Command``` and ```Deployer\Component\PharUpdate\Console\Helper```. Does deployer really need those in ```Application```?

elfet:
For phar - yes. Need to check if it's phar.

ceesvanegmond:
@elfet Can't we just remove those lines of code?

elfet:
@ceesvanegmond no. phar update will be broken then.

lucasmichot:
Maybe using https://github.com/padraic/phar-updater would be a good alternative then ?

elfet:
@lucasmichot no, i think current impl is better. I think it's possible only to install it in dev mode, and only for building phar.

lucasmichot:
Sorry @elfet , it's a misunderstanding; I meant:
instead of using `herrera-io/phar-update` (abandonned) in dev depdencies, we should use `padraic/phar-updater` in dev-dependecies

elfet:
@lucasmichot no :) adraic/phar-updater is another impl with different features, we will stick with deployer/phar-updater (fork of herrera-io/phar-update).

elfet:
Done. Updated deployer/phar-update.

PR #942, opened 2017-01-05
Add "testLocally" function

mbrodala:
This mirrors the run/runLocally functions.

| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -379,6 +379,20 @@ function test($command)
 }
 
 /**
+ * Run test command locally.
+ * Example:
+ *
+ *     testLocally('[ -d {{local_release_path}} ]')
+ *
+ * @param string $command
+ * @return bool
+ */
+function testLocally($command)
+{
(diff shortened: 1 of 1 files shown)


elfet:
What about different naming? I was also thinking about renaming `runLocally` to something like `exec` too.

Maybe `checkCommand` or new namespace `Local\test` `Local\run`?

mbrodala:
For now this could be added as is since the impact is low and migrating to something else should be easy.

Not sure about a better API in general for this:

- `exec` does not clearly convey the "local" execution part, one can also execute remotely.
- `checkCommand` implies checking Deployer commands which does not match its purpose

Something like `local\test(...)` and `local\run(...)` could work. This could also be used for `localServer` which could be come `local\server` then.

mbrodala:
Thanks for merging. I'll keep an eye on the local naming plans.

PR #947, opened 2017-01-06
CN-411 Fix exit code for parallel execution

aheart:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    |  No
| Deprecations? | No
| Fixed tickets | N/A

Backporting fix from 4.0
See https://github.com/deployphp/deployer/commit/3e40fcd958e2ac6eeb292fe1911f652803fe50db#diff-ecc506bcfd58cb6526cd2db33d5121cdR195

### DIFF
--- a/src/Executor/ParallelExecutor.php
+++ b/src/Executor/ParallelExecutor.php
@@ -120,6 +120,11 @@ class ParallelExecutor implements ExecutorInterface
     private $hasNonFatalException = false;
 
     /**
+     * @var string|bool
+     */
+    private $lastExceptionMessage = false;
+
+    /**
      * @param InputDefinition $userDefinition
      */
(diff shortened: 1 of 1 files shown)


elfet:
Why you just can upgrade to 4.0 version?

aheart:
One of our build servers is stuck with PHP 5.4 and will stay that way for an unknown period of time.
Everyone who is in a similar position could benefit from this, but If you would rather not invest time into 3.x I completely understand.
We can use our fork until an upgrade to 4.x is possible.

Thank you for taking the time to review this PR.

elfet:
@aheart can you use it via composer installation?

aheart:
@elfet No, but I've created a phar using your build script. Works well, thank you.

PR #948, opened 2017-01-07
Fix bug with $httpGroup guard clause

carlalexander:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

Just found this bug reading the source code. I think you meant to use `httpGroup` and not `httpUser` 😂


elfet:
Thanks:)

PR #950, opened 2017-01-11
always success

ad3n:
always success when running unlock

| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

Run `deploy:unlock` always return success instead of exception when `deploy.lock` not exist

### DIFF
--- a/recipe/deploy/lock.php
+++ b/recipe/deploy/lock.php
@@ -23,5 +23,8 @@ task('deploy:lock', function () {
 
 desc('Unlock deploy');
 task('deploy:unlock', function () {
-    run("rm {{deploy_path}}/.dep/deploy.lock");
+    try {
+        run("rm {{deploy_path}}/.dep/deploy.lock");//always success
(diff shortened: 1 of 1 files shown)


ad3n:
I checked, the error is not related with this PR

recipe/deploy/lock.php
```diff
@@ -23,5 +23,8 @@
 
 desc('Unlock deploy');
 task('deploy:unlock', function () {
-    run("rm {{deploy_path}}/.dep/deploy.lock");
+    try {
+        run("rm {{deploy_path}}/.dep/deploy.lock");//always success
+    } catch (\Exception $e) {
```
elfet: It's bad solution to mute all errors, better to use bash or rm options.

ad3n: Ok, i change to use `-f` instead of `try catch` block

Issue #951, opened 2017-01-11
variable collision from deploy.php


glensc:
| Q                 | A
| ----------------- | ---
| Issue Type        | Bug
| Deployer Version  | v4.0.2
| Local Machine OS  | Linux
| Remote Machine OS | N/A

### Description 

deployer entrypoint [bin/dep](https://github.com/deployphp/deployer/blob/v4.0.2/bin/dep) should protect itself from variable and function collisions.

at least it must not rely on any variable after reading user `deploy.php`:
- [bin/dep:107](https://github.com/deployphp/deployer/blob/v4.0.2/bin/dep#L107)

also shouldn't expose it's internals to user `deploy.php` either.

i suggest to move everything to a class under `Deployer` namespace.

### Content of `deploy.php`
```php
<?php
$deployer = null;
```

### Output log
*With enabled option for verbose output `-vvv`.*

```
$ deployer.phar -vvv
PHP Fatal error:  Call to a member function run() on null in phar:///home/glen/deployer.phar/bin/dep on line 110

$ ./vendor/bin/dep -vvv
PHP Fatal error:  Call to a member function run() on null in /home/glen/vendor/deployer/deployer/bin/dep on line 111
```

elfet:
So it's only $deployer var is out there? What about renaming to $__deployer__?
Or just use Deployer::get()->run();

glensc:
yes, for now `$deployer` may be overwritten by user code. but in the future there may be more. also the second point that user code should not access internal variables.

PR #952, opened 2017-01-11
Prevent leak of variables into deploy.php

mathbr:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #951

### DIFF
--- a/bin/dep
+++ b/bin/dep
@@ -104,7 +104,10 @@ $deployer = new \Deployer\Deployer($console, $input, $output);
 
 // Require deploy.php file
 if (is_readable($deployFile)) {
-    require $deployFile;
+    // Prevent variable leak into deploy.php file
+    call_user_func(function() {
+        require $deployFile;
+    });
 }
 
 // Run Deployer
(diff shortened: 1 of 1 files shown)


glensc:
interesting solution. seems to solve variable overwrite and variable leak.

elfet:
Nice, does it works on php5.6?

glensc:
closures come with php 5.3, so it must work

however particular implementation i think lacks `use $deployFile` for the closure

mathbr:
@glensc Good catch, I tried to prevent one variable too much from leaking. ;-)

glensc:
LGTM now

bin/dep
```diff
@@ -104,7 +104,10 @@ $deployer = new \Deployer\Deployer($console, $input, $output);
 
 // Require deploy.php file
 if (is_readable($deployFile)) {
-    require $deployFile;
+    // Prevent variable leak into deploy.php file
```
glensc: it's actually two way protection: leak into and leak from

PR #955, opened 2017-01-13
Raise timeout for runLocally to 300 seconds

mbrodala:
Similar to other locations this is now also raised to 5 minutes.

| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -331,7 +331,7 @@ function run($command)
  * @return Result Output of command.
  * @throws \RuntimeException
  */
-function runLocally($command, $timeout = 60)
+function runLocally($command, $timeout = 300)
 {
     $command = parse($command);
 


elfet:
Does it really needed?

elfet:
What about creating global config for setting local timeout?

mbrodala:
@elfet It is necessary for all commands within recipes which do not provide a configuration option for the timeout argument of `runLocally()`. We've discovered that e.g. [`rsync` does this](https://github.com/deployphp/recipes/blob/30967b0/rsync.php#L23), however not all locations do, e.g. [`local:vendors`](https://github.com/deployphp/recipes/blob/ea7e8a2/local.php#L166).

That's the background and the fact that the timeout was already raised to 300 seconds for the `Local` server (592e132).

elfet:
Ok, what about config option for that?

mbrodala:
@elfet Yes, I'll probably add PRs for recipes to add config options. Thanks for merging. :-)

PR #958, opened 2017-01-17
Try to remove deploy.lock when deploy is fail

ad3n:
Try to remove deploy.lock when deploy is fail

| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

Try to remove `deploy.lock` when deploy is fail to make it simpler

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -141,4 +141,6 @@ task('success', function () {
  */
 task('deploy:failed', function () {
 })->setPrivate();
+
 onFailure('deploy', 'deploy:failed');
+before('deploy:failed', 'deploy:unlock');


elfet:
I don't know if is best for default behavior. Maybe it is. What @deployphp/maintainer thinking?

ad3n:
@elfet : my team have a problem how to determine between failed deploy or locking by someone that running deployer. So from my side, it's better to remove `deploy.lock` when deploy had been fail

elfet:
Ok, lets give it a try)

ad3n:
Ok thank you :+1:

elfet:
@ad3n  I am reverting this. Got problems with race cond: one start deploy, another start deploy, fails and removes deploy.lock, so second can start another deploy interupting first deploy.

ad3n:
hhhmmm yes, you are right.

elfet:
Or there is another solution, throw GrasefulExitException from deploy:lock and do not fail here. In other cases deploy:unlock will be triggered.

ad3n:
I think that is not simple, but if we can make sure we have one folder in `release` that is better solution to make sure only one deploy that running

elfet:
What about removing this from common recipe and move to template instead? To make it optional for users. So they can choose what to do then deploy fails? What if they do not want to remove deploy.lock?

ad3n:
Not really that is good idea. Can we use `http://www.tldp.org/HOWTO/Bash-Prompt-HOWTO/x700.html` to counting how many deploy that running in same time

elfet:
I think it's better to make this configurable, and allow users to turn off: `before('deploy:failed', 'deploy:unlock');` 
Now it is not possible.

ad3n:
Not really that is good idea. Can we use `http://www.tldp.org/HOWTO/Bash-Prompt-HOWTO/x700.html` to counting how many deploy that running in same time

ad3n:
@elfet : :+1:

PR #962, opened 2017-01-18
Recursively add configuration options

mbrodala:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | Yes
| BC breaks?    | Maybe
| Deprecations? | No
| Fixed tickets | N/A

This ensures that something like the following works:

```php
set('top', [
  'first' => [
    'second' => [
      'value1',
      'value2',
    ],
  ],
]);

add('top', [
  'first' => [
    'second' => [
      'value3',
    ],
  ],
]);

/*
The following should yield:

Array
(
    [first] => Array
        (
            [second] => Array
                (
                    [0] => value1
                    [1] => value2
                    [2] => value3
                )
        )

)
*/

print_r(get('top'));

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
 - Added option `-L` to `setfacl` [#956](https://github.com/deployphp/deployer/pull/956)
 - Run `deploy:unlock` when deploy is fail [#958](https://github.com/deployphp/deployer/pull/958)
 - Now throw exception on duplicates in `shared_dirs`
-- `add()` now merges configuration options recursively
+- `add()` now merges configuration options recursively [#962](https://github.com/deployphp/deployer/pull/962)
 
 ### Fixed
 - Fixed native ssh scp option
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Added option `-L` to `setfacl` [#956](https://github.com/deployphp/deployer/pull/956)
 - Run `deploy:unlock` when deploy is fail [#958](https://github.com/deployphp/deployer/pull/958)
 - Now throw exception on duplicates in `shared_dirs`
+- `add()` now merges configuration options recursively
 
 ### Fixed
 - Fixed native ssh scp option
--- a/src/Deployer.php
+++ b/src/Deployer.php
@@ -167,7 +167,7 @@ class Deployer extends Container
             if (!is_array($config)) {
                 throw new \RuntimeException("Configuration parameter `$name` isn't array.");
             }
-            self::setDefault($name, array_merge($config, $array));
+            self::setDefault($name, array_merge_recursive($config, $array));
         } else {
             self::setDefault($name, $array);
         }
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by elfet** [comment]
Please, add test for this.

**Comment by mbrodala** [comment]
I will. :-)

**Comment by mbrodala** [comment]
@elfet Can you have another look?

**Comment by mbrodala** [comment]
Thanks a lot. :-)

**Comment by thadin** [comment]
This breaks recipes that use strings in their config, like rsync:

```php
set('rsync', [
    'exclude' => [
        '.git',
        'deploy.php',
    ],
    'exclude-file' => false,
    'include' => [],
    'include-file' => false,
    'filter' => [],
    'filter-file' => false,
    'filter-perdir' => false,
    'flags' => 'rz',
    'options' => ['delete'],
    'timeout' => 60,
]);

add('rsync', [
    'exclude' => [
        'uploads'
    ],
    'flags' => 'rzc'
]);
```

The function `array_merge_recursive` merges 2 strings with the same key into an array, as a result my `flags` become `['rz', 'rzc']`, which causes the rsync commando to have the arguments `-Array` (due to the toString conversion.

mbrodala:
@thadin You are absolutely right, nice catch. I'll make a PR with a possible fix.

mbrodala:
Short status message: `array_merge_recursive` fails on overriding scalar values, `array_replace_recursive` fails at keeping numerical keys. The latter is essential e.g. for `exclude` in the config above.

thadin:
One solution could be to create a custom method that loops (recursively) through the options and merges arrays while overriding scalars, but this might cause unexpected behavior for users.

elfet:
```php
I refactored the Daniel's function and I got it:

<?php
/**
* array_merge_recursive does indeed merge arrays, but it converts values with duplicate
* keys to arrays rather than overwriting the value in the first array with the duplicate
* value in the second array, as array_merge does. I.e., with array_merge_recursive,
* this happens (documented behavior):
*
* array_merge_recursive(array('key' => 'org value'), array('key' => 'new value'));
*     => array('key' => array('org value', 'new value'));
*
* array_merge_recursive_distinct does not change the datatypes of the values in the arrays.
* Matching keys' values in the second array overwrite those in the first array, as is the
* case with array_merge, i.e.:
*
* array_merge_recursive_distinct(array('key' => 'org value'), array('key' => 'new value'));
*     => array('key' => array('new value'));
*
* Parameters are passed by reference, though only for performance reasons. They're not
* altered by this function.
*
* @param array $array1
* @param array $array2
* @return array
* @author Daniel <daniel (at) danielsmedegaardbuus (dot) dk>
* @author Gabriel Sobrinho <gabriel (dot) sobrinho (at) gmail (dot) com>
*/
function array_merge_recursive_distinct ( array &$array1, array &$array2 )
{
  $merged = $array1;

  foreach ( $array2 as $key => &$value )
  {
    if ( is_array ( $value ) && isset ( $merged [$key] ) && is_array ( $merged [$key] ) )
    {
      $merged [$key] = array_merge_recursive_distinct ( $merged [$key], $value );
    }
    else
    {
      $merged [$key] = $value;
    }
  }

  return $merged;
}
```
Use this.

mbrodala:
@elfet That one also doesn't work, I now went with the one [mentioned here](http://stackoverflow.com/a/36366886/6812729).

PR #974, opened 2017-01-23
Set infinite timeout for local commands

mbrodala:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

Very long running commands (e.g. DB import) can exceed even 5 minutes (see #955) thus switch to an infinite timeout for local commands.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@
 ## master
 [v4.1.0...master](https://github.com/deployphp/deployer/compare/v4.1.0...master)
 
-—
+### Changed
+- Set infinite timeout for `runLocally()` and `Local::run()` [#974](https://github.com/deployphp/deployer/pull/974)
 
 ## v4.1.0
 [v4.0.2...v4.1.0](https://github.com/deployphp/deployer/compare/v4.0.2...v4.1.0)
--- a/src/Server/Local.php
+++ b/src/Server/Local.php
@@ -11,8 +11,6 @@ use Symfony\Component\Process\Process;
 
 class Local implements ServerInterface
 {
-    const TIMEOUT = 300;
-
     /**
      * @var Configuration
      */
@@ -64,8 +62,8 @@ class Local implements ServerInterface
     {
         $process = new Process($command);
         $process
-            ->setTimeout(self::TIMEOUT)
-            ->setIdleTimeout(self::TIMEOUT)
+            ->setTimeout(null)
+            ->setIdleTimeout(null)
             ->mustRun($callback);
 
         return $process->getOutput();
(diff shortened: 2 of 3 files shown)


elfet:
Why not `runLocally('...', null)`?

mbrodala:
@elfet Yes, for `runLocally()` this is possible but not with `run()` on a `localServer()`.

elfet:
> not with run() on a localServer().
I think we need to fix API and allow to set timeout for this too.

```php
run('...', INF);
```

mbrodala:
I'd be fine with that too. Notice that Symfony expects `null` for infinite timeouts, not `INF`.

pluseg:
@elfet What do you think about `null` value for infinite timeouts? It blocks the PR.

elfet:
Null is okay, but i like INF for verbosity. Also i'm thinking about changing `run` api in v5 to something like this: `run('...', $options)`. 
To becable pass timeouts, pty and other option for current run.

mbrodala:
@elfet Sounds good. How to proceed on this PR? I can rebase to fix the conflicts if you think it makes sense to have this added.

A little more background: with custom tasks you can of course manually pass `null` to enable infinite timeout, however this is not possible with 3rd party tasks/recipes.

elfet:
I will manually merge this pr later.

> however this is not possible with 3rd party tasks/recipes.
This is interesting point, we need found good API for it. What are problems to solve we want?

mbrodala:
@elfet That won't be easy to fix. We can reconfigure tasks by now but not the function calls (e.g. `runLocally()`) within those tasks.

elfet:
Done in #1092
```php
run('...', ['timeout' => null]);
runLocally('...', ['timeout' => null]);
```

elfet:
Also i was thinking about renaming `runLocally` to `exec`.

mbrodala:
@elfet The change you mentioned does not solve the issue mentioned here: I'd still need to copy the body of all tasks from 3rd party recipes to set the `timeout` option.

Thus please reconsider this PR.

staabm:
> Also i was thinking about renaming runLocally to exec.

I like run and runLocally as they cleary communicate their purpose.
exec and run dont make a statement whether they run remotely or locally... this would harm readability

elfet:
@mbrodala i see, maybe it's should be 3rd party issue? I like simplicity of options in _run_ function. For example, this is how i pass tty to git clone command: https://github.com/deployphp/deployer/pull/1092/commits/ace288978c633f621edbcd07c7512accf1cdf544#diff-b4ebd1b9bf76c38d2568949ec0c70c51

Of cause we can add some global param, but i don't like it for some reason.

mbrodala:
You are probably right that the 3rd party recipes should offer options to customize the timeout.

PR #977, opened 2017-01-24
Add non-recursive chmod flag to deploy:writable

greew:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

This commit enables the user to disable the recursive flag when setting writable dirs with `chmod`, by setting `writable_chmod_recursive` to `false`. It is per default set to `true`.

I have prefixed the setting with `writable_chmod_` to use the same naming convention as with `writable_chmod_mode`.

My personal scenario is that when I run the deployment, writable dirs are created by my own user and are then given 777 permission. The files within these writable dirs are then created by the web server user, and when I deploy the next time (without sudo), the deployment fails because the recursive flag makes the command trying to change permissions on the files as well, which my personal user doesn't own.

By removing the recursive flag, only the directories are being changed, which doesn't makes the deployment crash.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 ### Changed
 - `add()` now merges configuration options recursively [#962](https://github.com/deployphp/deployer/pull/962)
+- Added `writable_chmod_recursive` boolean option to enable non-recursive `chmod`
 
 ## v4.1.0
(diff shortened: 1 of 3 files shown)


elfet:
Cool, will merge it. Please add PR about this option to docs too.

greew:
Added to docs now :+1:

PR #987, opened 2017-01-27
use bin/php when composer is found

lewactwo:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -91,15 +91,15 @@ set('bin/git', function () {
 
 set('bin/composer', function () {
     if (commandExist('composer')) {
-        $composer = '{{bin/php}} ' . run('which composer')->toString();
+        $composer = run('which composer')->toString();
     }
 
     if (empty($composer)) {
         run("cd {{release_path}} && curl -sS https://getcomposer.org/installer | {{bin/php}}");
-        $composer = '{{bin/php}} {{release_path}}/composer.phar';
+        $composer = '{{release_path}}/composer.phar';
     }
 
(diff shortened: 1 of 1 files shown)


elfet:
Composer itself is a bin and can be called directly.

mbrodala:
@elfet True, but it will use `/usr/bin/env php` which is not the correct PHP CLI version to run on all servers, thus I consider this PR very valid.

lewactwo:
@elfet Yes it can, but on my hosting I have more php versions - `php55`, `php56`, `php70` and `php71`. Default php binary is symlink to `php56`. In config I set `bin/php` to `php71` and it works for everything but not for composer. Some of my packages require PHP 7.0 and without that I got error. I think if somebody set `bin/php` he want to use that php binary everywhere.

recipe/common.php
```diff
@@ -91,7 +91,7 @@
 
 set('bin/composer', function () {
     if (commandExist('composer')) {
-        $composer = run('which composer')->toString();
+        $composer = '{{bin/php}} ' . run('which composer')->toString();
```
elfet: Agree.

pluseg: @lewactwo Kamil, could you please update your PR with @mbrodala comment? It's about little duplication of `{{bin/php}} ` part.

lewactwo: @pluseg It's done.

PR #989, opened 2017-01-30
Properly install composer dependencies

b1rdex:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

Code should be tested on package versions defined in `composer.lock` rather then last versions of packages from `composer update`.

Commit c3564a28ac5289465494c23a1ad763edec4737b2 introduced `composer update` to workaround symfony php 5.4 compability.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,16 +11,11 @@ matrix:
   allow_failures:
     - php: 7.1
 
-env:
-  matrix:
-    - COMPOSER_FLAGS="--prefer-lowest"
-    - COMPOSER_FLAGS="--prefer-stable"
-
 install:
   - eval `ssh-agent -s`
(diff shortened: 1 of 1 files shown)


elfet:
But what is deployer installed via `composer require deployer/deployer`? Then composer.lock will be ignored. In this case testing with lower/higher deps is better.

b1rdex:
@elfet If you want to follow this scenario, then `composer.lock` should be removed completely. Then both cases (`composer require` and direct use + `composer install`) would behave same — `install` will get last versions.

elfet:
`composer.lock` is used when building phar version on build server.

PR #990, opened 2017-01-30
Use the old vendors directory for composer

yisraeldov:
| Q             | A
| ------------- | ---
| Bug fix?      |  No
| New feature?  | Yes 
| BC breaks?    | No
| Deprecations? |  No
| Fixed tickets | N/A 

### DIFF
--- a/recipe/deploy/vendors.php
+++ b/recipe/deploy/vendors.php
@@ -9,5 +9,11 @@ namespace Deployer;
 
 desc('Installing vendors');
 task('deploy:vendors', function () {
+    $releases = get('releases_list');
+    if (isset($releases[1])) {
+        if (run("if [ -d {{deploy_path}}/releases/{$releases[1]}/vendor ]; then echo 'true'; fi")->toBool()) {
+            run("cp --recursive {{deploy_path}}/releases/{$releases[1]}/vendor {{release_path}}");
+        }
+    }
     run('cd {{release_path}} && {{env_vars}} {{bin/composer}} {{composer_options}}');
 });


elfet:
Composer itself has cache.

yisraeldov:
Even if it is installed locally, not globally?

elfet:
Yes, you will see this: `Loaded from cache...`

recipe/deploy/vendors.php
```diff
@@ -9,5 +9,11 @@
 
 desc('Installing vendors');
 task('deploy:vendors', function () {
+    $releases = get('releases_list');
+    if (isset($releases[1])) {
+        if (run("if [ -d {{deploy_path}}/releases/{$releases[1]}/vendor ]; then echo 'true'; fi")->toBool()) {
```
mbrodala: Aside from what was said: `vendor-dir` is configurable and thus needs to be read from Composer config.

yisraeldov:
Npm directory is not configurable ?

mbrodala:
@yisraeldov [In Composer you can](https://getcomposer.org/doc/06-config.md#vendor-dir). ;-)

yisraeldov:
I asked about NPM because I took this code straight from there.

elfet:
@yisraeldov You can do this by using the --prefix flag and the --global* flag.

There is also better approach using `previous_release` from #991 like this:

```php
task('deploy:copy_vendors', function () {
    run('cp -r {{previous_release}}/vendor {{release_path}}/vendor');
});

before('deploy:vendors', 'deploy:copy_vendors');
```

yisraeldov:
Ok, that could be correct, but I just took the code from the NPM recipie,
so there should probably be an update there too

elfet:
> so there should probably be an update there too
Yes)

yisraeldov:
@elfet I would like to use the `previous_release` but I don't see that there is any PR for #991 Also the current way of getting previous_release like in NPM, and `update_code` doesn't work:

 ```
➤ Executing task deploy:update_code
[backend_dev] > which git
[backend_dev] > /usr/bin/git version
[backend_dev] > if [ -h /tmp/release ]; then echo 'true'; fi
[backend_dev] > readlink /tmp/release
[backend_dev] > /usr/bin/git clone  --recursive -q --reference /tmp/releases/22 --dissociate git@github.com:xxx  /tmp/releases/24 2>&1
```
It is cloneing using reference of  **22** when it should be using **23**, 

I see the same thing in other places when I try to refer to the previous release.

elfet:
Take look at deploy:copy_dirs

PR #996, opened 2017-02-04
Fix `parse` not replacing variables containing a `-` like `{{webserver-user}}`

mwr:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes 
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? |  No
| Fixed tickets | N/A

CHANGELOG.md is updated.

In the current master it is not possible to use a `_` in variables that are to be replaced using `\Deployer\parse()`.
This pull request updates the affected regex so it is possible to have variables like `{{webserver-user}}` replaced with the value from the config.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## master
 [v4.2.1...master](https://github.com/deployphp/deployer/compare/v4.2.1...master)
 
--
+- Fixed `parse` not replacing variables containing a `-` like `{{webserver-user}}`
 
 ## v4.2.1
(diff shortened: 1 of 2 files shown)


elfet:
You can use `_` instead of.

cmuench:
👍 for "-" in dynamic vars.

elfet:
I do not see any reasons for adding this. I think if there will be some restrictions: better recipes.

cmuench:
@elfet I agree if there is a concrete code style available.

elfet:
I will add this to docs, about option naming.

PR #997, opened 2017-02-05
Improve the WordPress recipe

khromov:
* Configure files commonly shared across deploys (`wp-config.php` and the uploads folder)
* Run Composer (many WordPress sites use Composer with WPackagist)
* chown file to the web server owner. (Might be platform-dependent, but this is a good default)

| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | Yes (If someone  had include the old recipe)
| Deprecations? | No
| Fixed tickets | N/A

### DIFF
--- a/recipe/wordpress.php
+++ b/recipe/wordpress.php
@@ -26,7 +26,7 @@ task('deploy', [
     'deploy:release',
     'deploy:update_code',
     'deploy:shared',
-    'deploy:vendors', 
+    'deploy:vendors',
     'deploy:writable',
     'deploy:symlink',
     'deploy:chown',
--- a/recipe/wordpress.php
+++ b/recipe/wordpress.php
@@ -9,15 +9,27 @@ namespace Deployer;
 
 require_once __DIR__ . '/common.php';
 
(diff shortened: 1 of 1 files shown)


recipe/wordpress.php
```diff
@@ -9,15 +9,27 @@
...
+set('shared_files', ['wp-config.php']);
+set('shared_dirs', ['wp-content/uploads']);
+
 /**
- * Main task
+ * Chown files to correct web server user
+ * for your OS, so uploads work.
  */
+task('deploy:chown', function () {
+    run('chown -R www-data:www-data ' . env('deploy_path'));
```
elfet: Http user maybe different. Also why not using only deploy:writable task?

khromov: You are right, it's a better solution to use `writable` task.

WordPress has a quirk that disables the plugin/theme editor and automatic updates if WP core files are not owned by the web server process. However, in this case it shouldn't matter because if you are using Deployer you probably don't want people to go and edit files or update through the web interface anyway.

PR #1000, opened 2017-02-06
Verify that writable directories exist

yisraeldov:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    |  No
| Deprecations? | No
| Fixed tickets | N/A 

If there is a directory in the `$writables` list that is not in the git repository, then attempting to set that directories permissions will cause the deploy to fail.

### DIFF
--- a/recipe/deploy/writable.php
+++ b/recipe/deploy/writable.php
@@ -18,6 +18,9 @@ task('deploy:writable', function () {
         return;
     }
 
+    //verify that directories exist
+    run("mkdir -vp $dirs");
+
     if ($httpUser === false && $mode !== 'chmod') {
         // Detect http user in process list.
         $httpUser = run("ps axo user,comm | grep -E '[a]pache|[h]ttpd|[_]www|[w]ww-data|[n]ginx' | grep -v root | head -1 | cut -d\\  -f1")->toString();


pluseg:
@yisraeldov I guess that requires a test. What do you mean by `that is not in the git repository`? Is it outside of a project root or just in .gitignore?

recipe/deploy/writable.php
```diff
@@ -18,6 +18,9 @@
         return;
     }
 
+    //verify that directories exist
+    run("mkdir -vp $dirs");
```
pluseg: @yisraeldov 
It doesn't work. Need to move it under `cd('{{release_path}}');`.
Also please add test in test/recipe/CommonTest.php:testWriteable(). It requires just adding non-existed folder in  `writable_dirs` and checking existence after `deploy:writable`.

Do you have time for that? If not then I'll do it.

yisraeldov:
Not sure when I'll have a chance to do it...

pluseg:
@yisraeldov it's not a problem, thanks for response! I will create a patch.

pluseg:
New PR: https://github.com/deployphp/deployer/pull/1049

PR #1007, opened 2017-02-13
Add task queue:restart for Laravel recipe

lasselehtinen:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

Laravel has an artisan command to send a restart signal to the queue workers. This is kind of a must during deployment. I noticed that the Laravel recipe does not have this artisan command, so here is a PR to add this command.

https://laravel.com/docs/5.4/queues#queue-workers-and-deployment

### DIFF
--- a/recipe/laravel.php
+++ b/recipe/laravel.php
@@ -14,12 +14,12 @@ require_once __DIR__ . '/common.php';
 
 // Laravel shared dirs
 set('shared_dirs', [
-    'storage'
+    'storage',
 ]);
 
 // Laravel shared file
 set('shared_files', [
-    '.env'
+    '.env',
 ]);
 
 // Laravel writable dirs
@@ -138,7 +138,6 @@ task('deploy', [
     'artisan:view:clear',
     'artisan:cache:clear',
     'artisan:config:cache',
-    'artisan:queue:restart',
     'artisan:optimize',
     'deploy:symlink',
     'deploy:unlock',
--- a/recipe/laravel.php
+++ b/recipe/laravel.php
@@ -98,6 +98,11 @@ task('artisan:optimize', function () {
     run('{{bin/php}} {{release_path}}/artisan optimize');
 });
 
+desc('Execute artisan queue:restart');
+task('artisan:queue:restart', function () {
+    run('{{bin/php}} {{release_path}}/artisan queue:restart');
+});
+
 /**
  * Task deploy:public_disk support the public disk.
  * To run this task automatically, please add below line to your deploy.php file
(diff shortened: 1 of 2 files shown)


recipe/laravel.php
```diff
@@ -98,6 +98,12 @@
     run('{{bin/php}} {{release_path}}/artisan optimize');
 });
 
+desc('Execute artisan queue:restart');
+task('artisan:queue:restart', function () {
+    run('{{bin/php}} {{release_path}}/artisan queue:restart');
+    writeln('<info>' . $output . '</info>');
```
elfet: `$output` is undefined. There is no need of writing output here.

elfet:
Also, please add this to deploy flow and CHANGELOG.md.

lasselehtinen:
I am not sure actually if there is a point for displaying the output at all, since [it is pretty much fixed](https://github.com/laravel/framework/blob/7212b1e9620c36bf806e444f6931cf5f379c68ff/src/Illuminate/Queue/Console/RestartCommand.php#L33). So I removed it.

I added task to the the deployment flow as you asked. Please note that all Laravel applications use queues, but I guess it is a sensible default. But the command works even if you have set the queue driver to "sync" so there is no harm done.

elfet:
@lasselehtinen so it's safe for all laravel users? If them not using queue?

lasselehtinen:
@elfet Yes it is safe to run always but might not be always necessary. Personally I have queue jobs on pretty much on all Laravel projects, but that is just me. But with a gut feeling as there is no statistics that larger majority uses the queue for sending emails, notifications etc. It is a non-blocking operation so it takes only one seconds or less.

elfet:
Well, maybe its better to not include this to recipe, but to template of laravel? So users can customize it easily.

lasselehtinen:
Yes I guess that is a best option, you can easily add `after('artisan:optimize', 'artisan:queue:restart');` in your deployment if you are using queues. I've updated the PR.

I guess that is something that could be covered in the recipe documentation, but I noticed there is no documentation for the Laravel recipe. I will try to contribute on that when I have time.

pluseg:
@elfet, it waits your approve and merge.

barryvdh:
I think this should be added by default. It doesn't hurt anyone, it's only creating a cache instance: https://github.com/laravel/framework/blob/5.4/src/Illuminate/Queue/Console/RestartCommand.php

So no effect for people not using queue, but very much required for everyone who is using queues!

Issue #1005, opened 2017-02-09
phar-update not a dev-dependency


KingCrunch:
| Q                 | A
| ----------------- | ---
| Issue Type        | Bug
| Deployer Version  | 4.2.1
| Local Machine OS  | Linux DELL-E5540-SEKR 4.9.8-1-ARCH #1 SMP PREEMPT Mon Feb 6 12:59:40 CET 2017 x86_64 GNU/Linux
| Remote Machine OS | N/A

### Description 
`deployer/phar-updater` is a `require-dev`. I'd expect, that a development requirement is only required für development, but once I installed deployer with `composer --no-dev` deployer fails to do _anything_ at all.

Expectation: Deployer works even when installed with `composer --no-dev`.

### Steps to reproduce

```
~$ dep -vvv
```

### Content of `deploy.php`

N/A

### Output log

```
PHP Fatal error:  Uncaught Error: Class 'Deployer\Component\PharUpdate\Console\Helper' not found in phar:///usr/bin/dep/src/Console/Application.php:72
Stack trace:
#0 phar:///usr/bin/dep/vendor/symfony/console/Application.php(83): Deployer\Console\Application->getDefaultHelperSet()
#1 phar:///usr/bin/dep/bin/dep(99): Symfony\Component\Console\Application->__construct('Deployer', '4.2.1')
#2 /usr/bin/dep(4): require('phar:///usr/bin...')
#3 {main}
  thrown in phar:///usr/bin/dep/src/Console/Application.php on line 72
```

elfet:
You are running phar version of deployer, not composer installed?

KingCrunch:
@elfet Exactly. I find it ugly to mix tools with dependencies

elfet:
Phar from deployer.org site contains this deps.

KingCrunch:
@elfet Don't get my wrong, I simply want to understand

So it is intended, that a Phar created without _development_ dependencies is actually broken?

elfet:
No, dev contains this package. What we can do is to check if class exists before adding it.

KingCrunch:
I've seen your comment after I've created the pull request #1006 (which I've created, because I didn't want to complain only ;)). Sure, you can close it, but I'd still disagree. The package is used during production. This makes it a concrete requirement and not something to enhance development.

[...]

PR #1016, opened 2017-02-15
Make dev-dependency to "required" because it's required

KingCrunch:
"deployer/phar-update" is not a development dependency, but it is required
within the phar and to build the phar

| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #1005

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,11 +31,11 @@
         "symfony/process": "~2.6|~3.0",
         "phpseclib/phpseclib": "~2.0",
         "pimple/pimple": "~3.0",
-        "monolog/monolog": "^1.21"
+        "monolog/monolog": "^1.21",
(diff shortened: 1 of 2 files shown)


elfet:
But what about #937 #929 ? Some people disagree this your pr. I think it's required only for building phar, not needed in composer installed version. 
But, maybe you are right and this is more strait forward solution.

elfet:
Maybe suggest section is best of all?

elfet:
Well, phar-update does not required. Will add check if it's available.

elfet:
Done 31fd582c40adf2bacf9e4d175808b9f93a2f7fef

Issue #995, opened 2017-02-02
Can not share same dirs


greatwitenorth:
| Q                 | A
| ----------------- | ---
| Issue Type        | Bug
| Deployer Version  | 4.2.1
| Local Machine OS  | OSX 10.11.6
| Remote Machine OS | Centos 6.8

### Description 
I'm seeing this error when Deployer is trying to create the shared folders:
```
[Deployer\Exception\ConfigurationException]
Can not share same dirs `img/newsletters` and `img/news`.
```

### Steps to reproduce
I use the following command:
```vendor/bin/dep deploy:shared staging -vvv```
and I see the error:
```
[Deployer\Exception\ConfigurationException]
  Can not share same dirs `img/newsletters` and `img/news`.
```

It appears as though there's a strpos match happening in ```vendor/deployer/deployer/recipe/deploy/shared.php``` on line 19. I'm not sure why partial matches should be throwing an exception. Shouldn't it only be exact matches?

### Content of `deploy.php`
```php
namespace Deployer;

require 'vendor/deployer/deployer/recipe/common.php';
require 'vendor/deployer/recipes/phinx.php';

// Set configurations
set('repository', 'git@bitbucket.org:myuser/myrepo.git');
set('shared_files', [
	'inc/env.php'
]);
set('shared_dirs', [
	'builds',
	'cache',
	'docs',
	'free-ride-html',
	'img/ads',
	'img/awards',
	'img/board',
	'img/congress',
	'img/gallery',
	'img/mailchimp',
	'img/news',
	'img/newsletters',
	'img/raffle',
	'img/sponsors',
	'img/staff',
	'kml',
	'slir/cache/rendered',
	'slir/cache/request',
]);

set('phinx.configuration', './migration/.phinx.yml');

// Configure servers
server('production', 'mysite.com')
    ->user('myuser')
	->identityFile()
	->set('branch', 'master')
	->stage('production')
	->set('phinx.environment', 'development')
    ->set('deploy_path', '/home/myuser/production');

server('staging', 'mysite.com')
	->user('myuser')
	->identityFile()
	->set('branch', 'dev')
	->stage('staging')
	->set('phinx.environment', 'staging')
	->set('deploy_path', '/home/myuser/staging');


task('deploy:set_permissions', function(){
	run("chmod -R go-w {{deploy_path}}");
}); 

task('deploy:empty_slir', function () {
	$sharedPath = "{{deploy_path}}/shared";
[...]

PR #1022, opened 2017-02-17
Fix sharing folders having similar names

pluseg:
`Can not share same dirs` bug fix.

| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #995

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,9 @@
 ## master
 [v4.2.1...master](https://github.com/deployphp/deployer/compare/v4.2.1...master)
 
--
+### Fixed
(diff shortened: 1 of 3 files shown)


pluseg:
@elfet please note that Scrutinizer passes but the build is actually failed. It seems that Scrutinizer uses wrong phpunit version (`phpunit` rather than `vendor/bin/phpunit`). The same issue about Travis-CI in PhpUnit repo — https://github.com/sebastianbergmann/phpunit/issues/2432.

elfet:
Cool, thanks! Also there is possible to implement better version of this check using virtual dir tree for checking.

PR #1026, opened 2017-02-17
Parallel executor does not wait for once() task even if they can have an impact of the future tasks

ahfeel:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 - Fixed `Can not share same dirs` for shared folders having similar names [#995](https://github.com/deployphp/deployer/issues/995)
 - Fixed scalar override on recursive option merge [#1003](https://github.com/deployphp/deployer/pull/1003)
 - Fixed incompatible PHP 7.0 syntax [#1020](https://github.com/deployphp/deployer/pull/1020)
-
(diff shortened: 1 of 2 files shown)


elfet:
Cool. Can you show example of deploy.php file for tests?

ahfeel:
Sure !

Here's an example that would fail :

```
<?php

namespace Deployer;

require 'recipe/common.php';

set('ssh_type', 'native');
set('ssh_multiplexing', true);

serverList('serverlist.yml');

option('global_flag', null, InputOption::VALUE_OPTIONAL, 'Set a super flag.');
task('test:step1', function() {
    Deployer::setDefault('global_flag', explode(',', input()->getOption('global_flag')));
})->once();

task('test:step2', function() {
    writeln('Has the flag ? ' . Deployer::getDefault('global_flag'));
});

task('test:all', [
    'test:step1',
    'test:step2',
]);
```

If you run this with the parallel execution flag, the step2 will be triggered at the same time than the step1, and the global option won't be parsed.

elfet:
This scenario can't work in parallel more. Because on share-nothing memory model. When you do `Deployer::setDefault` in task `test:step1`, it update only master process, and other does not changed.

PR #1028, opened 2017-02-17
Enable TTY for local commands

mbrodala:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

This is essential when executing interactive commands via `runLocally()`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 ### Added
 - Add a way to retrieve a defined task [#1008](https://github.com/deployphp/deployer/pull/1008)
+- Enable TTY for local commands [#1028](https://github.com/deployphp/deployer/pull/1028)
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


elfet:
I don't think what interactive by default is good solution. It's better to make if optional. Will do it in next v5.

mbrodala:
@elfet What issues exactly do you see here?

t33k:
+1 for NOT having that by default. I am planing to do deploy not from local computer but by calling one single command on remote server so I suppose I will use LocalServer() extensively and I should not have interaction there.

elfet:
Some command doesn't require tty.

elfet:
In v5 will be something like this:
```php
runLocally('...', ['tty' => true]);
```

Issue #1013, opened 2017-02-14
Implement task ssh for connecting to server


antonmedv:
| Q                 | A
| ----------------- | ---
| Issue Type        | Feature Request
| Deployer Version  | next

Implement task `dep ssh [server]` for opening ssh shell to remote server and cd to `deploy_path`.

pluseg:
@elfet as I know you are going to support only Native SSH in the nearest future, right? I have a good solution for Native SSH and I'm not sure do we need to implement same functionality for other ssh connection methods if they will be cut? What do you think?

elfet:
Only for native ssh is okay. in v5 two other will be droped.

antonmedv:
📦 Done in #1092

PR #1035, opened 2017-02-21
Implement `ssh` task for opening ssh connection and going to a release path

pluseg:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #1013 

After review and approval I'll add a PR into docs repo.

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -21,11 +21,6 @@ require __DIR__ . '/deploy/cleanup.php';
 require __DIR__ . '/deploy/copy_dirs.php';
 require __DIR__ . '/deploy/rollback.php';
 
-use Deployer\Server\SSHPipeInterface;
-use Deployer\Task\Context;
-use Symfony\Component\Console\Input\InputArgument;
-use Symfony\Component\Console\Input\InputOption;
-
 /**
  * Configuration
  */
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -115,35 +115,11 @@ set('bin/symlink', function () {
 });
 
 /**
- * Default arguments and options.
- */
-argument('stage', InputArgument::OPTIONAL, 'Run tasks only on this server or group of servers');
-option('tag', null, InputOption::VALUE_OPTIONAL, 'Tag to deploy');
-option('revision', null, InputOption::VALUE_OPTIONAL, 'Revision to deploy');
-option('branch', null, InputOption::VALUE_OPTIONAL, 'Branch to deploy');
-
-/**
  * Tasks
  */
 desc('Show current release');
 task('current', function () {
     writeln('Current release: ' . basename(get('current_path')));
-});
-
-desc('Connect to ssh and `cd` to release path');
-task('ssh', function () {
-    $server = Context::get()->getServer();
-
-    if (!$server instanceof SSHPipeInterface) {
-        throw new \RuntimeException('The ssh type `' . get('ssh_type') . '` doesn\'t support `ssh` task.');
-    }
-
-    $initialCommand = null;
-    if (get('current_path')) {
-        $initialCommand = 'cd ' . escapeshellarg(get('current_path')) . '; exec \$SHELL';
-    }
-
-    $server->createSshPipe($initialCommand);
 });
 
 
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -21,6 +21,8 @@ require __DIR__ . '/deploy/cleanup.php';
 require __DIR__ . '/deploy/copy_dirs.php';
 require __DIR__ . '/deploy/rollback.php';
 
+use Deployer\Server\SSHPipeInterface;
+use Deployer\Task\Context;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputOption;
 
@@ -126,6 +128,22 @@ option('branch', null, InputOption::VALUE_OPTIONAL, 'Branch to deploy');
 desc('Show current release');
 task('current', function () {
     writeln('Current release: ' . basename(get('current_path')));
+});
+
+desc('Connect to ssh and `cd` to release path');
+task('ssh', function () {
+    $server = Context::get()->getServer();
+
+    if (!$server instanceof SSHPipeInterface) {
+        throw new \RuntimeException('The ssh type `' . get('ssh_type') . '` doesn\'t support `ssh` task.');
+    }
+
+    $initialCommand = null;
+    if (get('current_path')) {
+        $initialCommand = 'cd ' . escapeshellarg(get('current_path')) . '; exec \$SHELL';
+    }
+
+    $server->createSshPipe($initialCommand);
 });
 
 
--- a/src/Server/Configuration.php
+++ b/src/Server/Configuration.php
@@ -340,6 +340,16 @@ class Configuration
     }
 
     /**
+     * Get user and host
+     *
+     * @return string
+     */
+    public function getUserAndHost()
+    {
+        return (!empty($this->getUser()) ? $this->getUser() . '@' : '') . $this->getHost();
+    }
+
+    /**
      * Get name
      *
      * @return string
--- a/src/Server/Remote/NativeSsh.php
+++ b/src/Server/Remote/NativeSsh.php
@@ -47,7 +47,7 @@ class NativeSsh implements ServerInterface, SSHPipeInterface
      */
     public function run($command)
     {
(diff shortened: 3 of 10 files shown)


elfet:
Cool. But better use posix exec.

recipe/common.php
```diff
@@ -128,6 +130,22 @@
     writeln('Current release: ' . basename(get('current_path')));
 });
 
+desc('Connect to ssh and `cd` to release path');
+task('ssh', function () {
```
elfet: Better to create it as ssh console command, not task.

src/Server/SSHPipeInterface.php
```diff
@@ -0,0 +1,19 @@
+<?php
+/* (c) Anton Medvedev <anton@medv.io>, Maxim Kuznetsov <skypluseg@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Deployer\Server;
+
+interface SSHPipeInterface
```
elfet: This it really strange interface.

pluseg: What's strange for you? I think that there is a chance in the future that we need to add new type of `Server`, so we need to define which of them supports `ssh` task and which one does not. If we have an interface then we can be sure that a class implemented one has concrete method like `createSshPipe()`. Or do you think that checking `getenv('ssh_type')` is enough?

recipe/common.php
pluseg: @elfet could you please explain the better side of this decision? I'm pretty new in the project so I can miss something important.

pluseg:
@elfet 

> But better use posix exec.

Do you mean [pcntl_exec](http://php.net/manual/en/function.pcntl-exec.php)? Yeah, it's the best solution but it requires `pcntl` extension. Also PHP.net says "Currently, this module will not function on non-Unix platforms (Windows)". If you mean simple `exec()` function then it doesn't work for me, terminal just freezing. Is there another solution that I missed?

elfet:
Use pcntl_exec if possible otherwise proc_open.

pluseg:
Mini-changelog:
1. Added pcntl support for running `ssh` process separately.
2. pcntl_exec gets different arguments and escape them rather than proc_open. So it leads to light refactoring of building ssh command line options.
3. Added `ssh` console command for ssh connections
4. The item above leads to moving getting stage, tag and revision options to TaskCommand. I guess that it's more suitable place.

@elfet do i need to reflect all these items in CHANGELOG.md or not? I added only third item as the most valuable.

recipe/common.php
```diff
@@ -113,14 +115,6 @@
...
 
 /**
- * Default arguments and options.
- */
-argument('stage', InputArgument::OPTIONAL, 'Run tasks only on this server or group of servers');
-option('tag', null, InputOption::VALUE_OPTIONAL, 'Tag to deploy');
-option('revision', null, InputOption::VALUE_OPTIONAL, 'Revision to deploy');
-option('branch', null, InputOption::VALUE_OPTIONAL, 'Branch to deploy');
-
-/**
```
pluseg: I moved this section into TaskCommand. I have no idea why they are called here. @elfet do you have any thoughts here?

src/Server/Configuration.php
```diff
@@ -340,6 +340,16 @@ public function setUser($user)
     }
 
     /**
+     * Get user and host
+     *
+     * @return string
+     */
+    public function getUserAndHost()
+    {
+        return (!empty($this->getUser()) ? $this->getUser() . '@' : '') . $this->getHost();
```
pluseg: It's used in several places so I create separate method.

src/Server/Remote/NativeSsh.php
```diff
@@ -141,23 +99,25 @@ public function scpCopy($target, $target2)
...
 
         if ($serverConfig->getPort()) {
-            $scpOptions[] = '-P ' . escapeshellarg($serverConfig->getPort());
+            $scpOptions = array_merge_recursive($scpOptions, ['-P' => $serverConfig->getPort()]);
         }
 
         if ($serverConfig->getPrivateKey()) {
-            $scpOptions[] = '-i ' . escapeshellarg($serverConfig->getPrivateKey());
+            $scpOptions = array_merge_recursive($scpOptions, ['-i' => $serverConfig->getPrivateKey()]);
+        } elseif ($serverConfig->getPemFile()) {
```
pluseg: I've noticed that pem file support for ssh connection was added recently but it was missed for scp.

src/Server/Remote/NativeSsh.php
```diff
@@ -230,12 +206,55 @@ protected function getConnectionHash()
...
     {
         return [
-            '-o ControlMaster=auto',
-            '-o ControlPersist=5',
-            '-o ControlPath=\'' . $this->getConnectionHash() . '\'',
+            '-o' => [
+                'ControlMaster=auto',
+                'ControlPersist=5',
+                'ControlPath=' . $this->getConnectionHash(),
+            ],
```
pluseg: I changed structure of options due to new way of building command line arguments.

src/Server/Remote/NativeSsh.php
```diff
@@ -244,32 +263,118 @@ protected function getMultiplexingSshOptions()
...
+            $serverConfig->getUserAndHost() => null,
+        ]);
+
+        exec('ssh -O check -S ' . $this->getConnectionHash() . ' ' . $this->buildCommandLineOptions($sshOptions, true) . ' 2>&1', $checkIfMuxActive);
+        if (!preg_match('/Master running/', $checkIfMuxActive[0])) {
             if (\Deployer\isVerbose()) {
                 \Deployer\writeln('  SSH multiplexing initialization');
             }
-            exec('ssh ' . implode(' ', $sshOptions) . ' ' . escapeshellarg($username . $hostname) . "  'echo \"SSH multiplexing initialization\"' ");
+            exec('ssh ' . $this->buildCommandLineOptions($sshOptions, true) . "  'echo \"SSH multiplexing initialization\"' ");
```
pluseg: Actually I didn't test it after options refactoring because I'm not sure how to do that correctly.

src/Server/Remote/NativeSsh.php
```diff
@@ -244,32 +263,118 @@ protected function getMultiplexingSshOptions()
...
+     * @param string|null $initialCommand
+     * @throws \RuntimeException if command is invalid
+     */
+    private function createPcntlSshPipe($initialCommand = null)
+    {
+        /**
+         * [HACK] PCNTL works only on Unix and `pcntl_exec` requires full path of running command.
+         * So it defines path of `ssh` using `which` utility.
+         */
+        $sshPath = trim(shell_exec('which ssh'));
```
pluseg: I have no idea how to do it more cross-platform.

src/Server/Remote/NativeSsh.php
```diff
@@ -230,12 +206,55 @@ protected function getConnectionHash()
...
+            ],
         ];
     }
 
+    /**
+     * Returns final ssh connection command with configs, username and host.
+     *
+     * @return array
+     */
+    private function getSshConnectionOptions()
```
ossinkine: I think this these options should be reusable. Maybe make this method public?
For example, I use rsync in my docker.php and it also needs these options. How should I pass them into it?

src/Server/SSHPipeInterface.php
pluseg: @elfet I don't understand what I should do here. My position is written down above. I think that name of the interface might be better but I have no idea.

src/Server/Remote/NativeSsh.php
pluseg: @ossinkine through configuration in deploy.php.

pluseg: @ossinkine In fact we need to create separate AbstractOptionsBuilder and SshOptionsBuilder class for working with all that stuff. But the PR is already too big so I'm going to do it next time.

elfet:
A lot of refactoring. Will try t review it all. Or maybe will cherry pick some of your code to new v5 branch.
Great work, by the way.

pluseg:
This refactoring isn't final. There is still code duplication and high complexity. But now it is prepared for further improvements. E.g we can move command line options to separate class which requires only $this->configuration and can build options for different targets (scp, ssh, etc).

elfet:
@pluseg okay. I can finish your pr by mysql. 🙂

antonmedv:
Done in #1092

Issue #1004, opened 2017-02-09
Keep releases keeps adding directories (named 1, 1.1, 1.1.2, 1.1.2.3)


filipponeri:
| Q                 | A
| ----------------- | ---
| Issue Type        | Bug
| Deployer Version  | 4.2
| Local Machine OS  | Ubuntu 16.10
| Remote Machine OS | Ubuntu 16.10

### Description 
the set('keep_releases', 3); command does not appear to work.
Deployer keeps adding directories (with names and in this order: '1', '2' , '1.1' , '1.1.2' , '1.1.2.3' ) without deleting any previous release directories. This results in quickly filling up the disk space.

Apart for this problem, Deploy deploys correctly the new software on the remote web server and does not output any error.
Thank you for your help. 
Best regards, Filippo

Running deploy
### Steps to reproduce
dep deploy.php

### Content of `deploy.php`
```php<?php

namespace Deployer;
//require_once 'recipe/common.php';
require 'recipe/laravel.php';

//use function Deployer\{server, task, run, set, get, add, before, after};

set('ssh_type', 'native');
set('ssh_multiplexing', true);

set('repository', 'git@gitlab.com:repository');

// Laravel writable dirs
add('shared_files', []);
add('writable_dirs', ['storage', 'vendor']);

server('staging', 'hosting', 22232)
    ->user('usename')
    ->password('pwd')
    ->set('branch', 'master')
    ->set('deploy_path', '/home/provrtsp/public_html/promarket11.com')
    ->pty(true);

set('keep_releases', 3);

//->identityFile('~/.ssh/id_rsa.pub', '~/.ssh/id_rsa')

/**
 * Setup the environment file in the new release
 */
task('environment', function () {
    run('cp /home/provrtsp/shared/env {{release_path}}/.env');
   
})->desc('Environment setup');

task('composer_rerun', function () {
    run('cd /home/provrtsp/public_html/promarket11.com/current; php ~/composer.phar install ');
})->desc('Environment setup');

// Laravel writable dirs
set('writable_dirs', ['storage', 'vendor']);

// to be added to task call below 'deploy:symlink',

/**
 * Main task
 */
task('deploy', [
    'deploy:prepare',
    'deploy:release',
    'deploy:update_code',
    'deploy:vendors',
    'deploy:symlink',
    'cleanup',
    'environment',
])->desc('Deploy your project');

after('deploy', 'success');
after('success', 'composer_rerun');                                                                                                           
                                                    
```

### Output log   
[...]

PR #1036, opened 2017-02-21
Fixed an issue with the output of ls in releases_list that was breaking cleanup.

crosma:
Forces single column output and trims newlines from ls. On our servers, we have an alias for ls that improves the default output but was causing cleanup to not work.

| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | Unsure


elfet:
#1004

PR #1042, opened 2017-02-22
Fixed old releases not being cleaned up when keep_releases reduced by than half.

crosma:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | Side effect noticed after fixing #1004 #1036

If as a user reduced their keep_releases by more than half of the number of releases they have on their server, the oldest would be ignored and not cleaned up. 

For example, if you have 100 releases on your server and reduce keep_releases to 20, only the most recent 45 releases would be read from the CSV. This would cause the oldest 55 releases to be ignored and not cleaned up.

The simple fix was to make sure it always reads as many lines from the CSV as there are release directories.


elfet:
Nice solution 👍🏻

PR #1043, opened 2017-02-22
Removed dependency for man package

thadin:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

The man package isn't present on our servers (space savings & speed-ups during apt-get actions). The same info could be retrieved with the `--help` attribute.

Without this fix Deployer won't use atomic deployments causing a small time frame in which our website returns 404 errors.

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -105,7 +105,7 @@ set('bin/composer', function () {
 set('bin/symlink', function () {
     if (get('use_relative_symlink')) {
         // Check if target system supports relative symlink.
-        if (run('if [[ "$(man ln)" =~ "--relative" ]]; then echo "true"; fi')->toBool()) {
+        if (run('if [[ "$(ln --help)" =~ "--relative" ]]; then echo "true"; fi')->toBool()) {
             return 'ln -nfs --relative';
         }
     }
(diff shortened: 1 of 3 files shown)


elfet:
On macOS got this:
```
$ mv --help
mv: illegal option -- -
usage: mv [-f | -i | -n] [-v] source target
       mv [-f | -i | -n] [-v] source ... directory
```

I think it's better to add configuration option, similar to `use_relative_symlink`.

jordisala1991:
Would be better to just do like capistrano, they create release directory in other folder with the same name 'current', then at the end just move the symlink to the old symlink location and it works.

For example, create 'release' dir called 'current' inside .dep dir, then at the end move one directory out with '..'

recipe/common.php
```diff
@@ -104,10 +105,7 @@
 
 set('bin/symlink', function () {
     if (get('use_relative_symlink')) {
-        // Check if target system supports relative symlink.
-        if (run('if [[ "$(man ln)" =~ "--relative" ]]; then echo "true"; fi')->toBool()) {
-            return 'ln -nfs --relative';
-        }
+        return 'ln -nfs --relative';
```
elfet: Why do you remove relative detection at all?

elfet:
I see what your are trying to implement, i will do it myself, okay? 🙂

elfet:
Done 14a8f8f9c4ebbc7da45c2b6b7c3c00a51b563ccf

Issue #1012, opened 2017-02-14
NativeSsh, improve user experience for connection failures


pniederlag:
| Q                 | A
| ----------------- | ---
| Issue Type        | Feature Request
| Deployer Version  | 4.2.1
| Local Machine OS  | N/A
| Remote Machine OS | N/A

### Description 
*As suggested I switched to the native ssh connection via set('ssh_type', 'native')*

When I do have an error in the setup that prevents a succesfull (ssh-)connection the error shown for the native-ssh implementation is very irritating. 

### Steps to reproduce

* add some invalid server configuration
* try to run an arbitrary task on this server
* EXCEPTION  [RuntimeException]                              
  Unable to login with the provided credentials is clear to me.
* switch to native ssh implementation as the warning suggests
* EXCEPTION: [Symfony\Component\Process\Exception\ProcessFailedException] is irritating for me

### Content of `deploy.php`
```php
// compat with php < 5.6
namespace Deployer;

use Symfony\Component\Console\Input\InputArgument;

//
require 'recipes/rsync.php';

argument('stage', InputArgument::OPTIONAL, 'Run tasks only on this server or group of servers');

//set('ssh_type', 'native');
//set('ssh_multiplexing', true);

// Configuration
serverList('servers.yml');

desc('Say Hello');
task('say_hello', function () {
     writeln('Hello');
     run("pwd");
});%  
```

elfet:
Done.

PR #1051, opened 2017-02-24
Change output of errors for native ssh (#1012)

pluseg:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #1012

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
 
 ### Changed
 - Autoload functions via Composer [#1015](https://github.com/deployphp/deployer/pull/1015)
+- Add task queue:restart for Laravel recipe [#1007](https://github.com/deployphp/deployer/pull/1007)
+- Changed output of errors for native ssh [#1012](https://github.com/deployphp/deployer/issues/1012)
 
 ### Fixed
 - Fixed `Can not share same dirs` for shared folders having similar names [#995](https://github.com/deployphp/deployer/issues/995)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -18,8 +20,6 @@
 - Fixed possibility to use PEM files with Native SSH
 - Fixed old releases not being cleaned up when keep_releases reduced by more than half.
 
-### Changed
-- Add task queue:restart for Laravel recipe [#1007](https://github.com/deployphp/deployer/pull/1007)
```
pluseg: Just combined with same section above.

src/Server/Remote/NativeSsh.php
```diff
@@ -49,7 +50,6 @@ public function run($command)
         $serverConfig = $this->getConfiguration();
         $sshOptions = [
             '-A',
-            '-q',
```
pluseg: @elfet Is there a real reason of adding "-q" option? Without it `ssh` returns 255 status code (Unknown error) and empty output.

elfet: Don't know 🙄 It came from PR.

src/Server/Remote/NativeSsh.php
```diff
@@ -85,12 +85,17 @@ public function run($command)
...
-            ->mustRun();
+        try {
+            $process = new Process($sshCommand);
+            $process
+                ->setPty($serverConfig->getPty())
+                ->setTimeout(null)
+                ->setIdleTimeout(null)
+                ->mustRun();
+        } catch (ProcessFailedException $exception) {
+            $errorMessage = \Deployer\isVerbose() ? $exception->getMessage() : $process->getErrorOutput();
```
elfet: Why in verbose mode using message and otherwise error output?

pluseg: @elfet Error output returns only error string like "invalid credentials" and an exception message has current look with full broken command, current directory, error string, etc. I think that common user needs only concise error string.

elfet: But why not isDebug then?

pluseg: @elfet I used it but then noted that there is one usage of Verbose in the file. Okay, I'll update the PR.

PR #1054, opened 2017-02-25
Fix some issues from SensioLabsInsight

pluseg:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

It's about refactoring.

### DIFF
--- a/src/Util/Reporter.php
+++ b/src/Util/Reporter.php
@@ -14,40 +14,47 @@ class Reporter
     /**
      * @param array $stats
      */
-    public static function report($stats)
+    public static function report(array $stats)
     {
-        $send = function () use ($stats) {
-            if (extension_loaded('curl')) {
-                $body = json_encode($stats, JSON_PRETTY_PRINT);
-                $ch = curl_init(self::ENDPOINT);
-                curl_setopt($ch, CURLOPT_HTTPHEADER, [
-                    'Content-Type: application/json',
-                    'Content-Length: ' . strlen($body)
-                ]);
-                curl_setopt($ch, CURLOPT_CUSTOMREQUEST, 'GET');
-                curl_setopt($ch, CURLOPT_POSTFIELDS, $body);
-                curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
-                curl_setopt($ch, CURLOPT_FOLLOWLOCATION, true);
-                curl_setopt($ch, CURLOPT_MAXREDIRS, 10);
-                curl_setopt($ch, CURLOPT_CONNECTTIMEOUT, 5);
-                curl_setopt($ch, CURLOPT_TIMEOUT, 5);
-                curl_exec($ch);
-            } else {
-                file_get_contents(self::ENDPOINT . '?' . http_build_query($stats));
-            }
-        };
-
-
+        $pid = null;
         if (extension_loaded('pcntl')) {
             declare(ticks = 1);
             $pid = pcntl_fork();
-            if ($pid === 0) {
-                posix_setsid();
-                $send();
-                exit(0);
-            }
+        }
+
+        if (is_null($pid) || $pid === -1) {
+            // Fork fails or there is no `pcntl` extension.
+            self::send($stats);
+        } elseif ($pid === 0) {
+            // Child process.
+            self::send($stats);
+            // Close child process after doing job.
+            exit(0);
+        }
+    }
+
+    /**
+     * @param array $stats
+     */
+    private static function send(array $stats)
+    {
+        if (extension_loaded('curl')) {
+            $body = json_encode($stats, JSON_PRETTY_PRINT);
+            $ch = curl_init(self::ENDPOINT);
+            curl_setopt($ch, CURLOPT_HTTPHEADER, [
+                'Content-Type: application/json',
+                'Content-Length: ' . strlen($body)
+            ]);
+            curl_setopt($ch, CURLOPT_CUSTOMREQUEST, 'GET');
+            curl_setopt($ch, CURLOPT_POSTFIELDS, $body);
+            curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+            curl_setopt($ch, CURLOPT_FOLLOWLOCATION, true);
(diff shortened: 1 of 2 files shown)


src/Util/Reporter.php
```diff
@@ -14,40 +14,47 @@ class Reporter
...
-                curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
-                curl_setopt($ch, CURLOPT_FOLLOWLOCATION, true);
-                curl_setopt($ch, CURLOPT_MAXREDIRS, 10);
-                curl_setopt($ch, CURLOPT_CONNECTTIMEOUT, 5);
-                curl_setopt($ch, CURLOPT_TIMEOUT, 5);
-                curl_exec($ch);
-            } else {
-                file_get_contents(self::ENDPOINT . '?' . http_build_query($stats));
-            }
-        };
```
pluseg: Extracted in separate method.

src/Util/Reporter.php
```diff
@@ -14,40 +14,47 @@ class Reporter
...
-
-
+        $pid = null;
         if (extension_loaded('pcntl')) {
             declare(ticks = 1);
             $pid = pcntl_fork();
-            if ($pid === 0) {
-                posix_setsid();
```
pluseg: @elfet Do we really need to use `posix_setsid()` here?

src/Util/Reporter.php
```diff
@@ -14,40 +14,47 @@ class Reporter
...
+
+        if (is_null($pid) || $pid === -1) {
+            // Fork fails or there is no `pcntl` extension.
+            self::send($stats);
+        } elseif ($pid === 0) {
+            // Child process.
+            self::send($stats);
+            // Close child process after doing job.
+            exit(0);
+        }
```
pluseg: Tried to decrease complexity code here.

src/Util/Reporter.php
elfet: > posix_setsid — Make the current process a session leader

It's needed to create fully separate fork. That way main process can finish successfuly, while sending stats in background.

src/Util/Reporter.php
```diff
@@ -14,40 +14,47 @@ class Reporter
...
             declare(ticks = 1);
             $pid = pcntl_fork();
-            if ($pid === 0) {
-                posix_setsid();
-                $send();
-                exit(0);
-            }
+        }
+
+        if (is_null($pid) || $pid === -1) {
```
pluseg: Null returns if there is no pcntl extension in the system.

src/Util/Reporter.php
pluseg: I've tested it and all works fine without setsid. Moreover as I understand if a leader process finishes then it closes all children. I guess that main process will run longer than child will send stats.

src/Util/Reporter.php
elfet: > Fatal error: Uncaught Error: Call to undefined function pcntl_fork() in

src/Util/Reporter.php
elfet: I will be working, but main process will be waiting to child. Creating another session allow to finish current process successfully and send stats in background.

src/Util/Reporter.php
pluseg: @elfet did you ran simple test `php -r "echo pcntl_fork();"` or execute my code? It seems that's impossible in my code as there is a check `extension_loaded('pcntl')`.
```
$pid = null;
if (extension_loaded('pcntl')) {
    ...
    $pid = pcntl_fork();
```
So if the extension is off then `$pid = null;`.

elfet: I see. Sorry, mislead code flow. But `posix_setsid()` still required. You can check this by `time dep task`

src/Util/Reporter.php
pluseg: @elfet 
> posix_setsid — Make the current process a session leader.
So a child process becomes a parent.

My test script:
```
<?php

$pid = null;
if (extension_loaded('pcntl')) {
    declare(ticks = 1);
    $pid = pcntl_fork();
}

if (is_null($pid) || $pid === -1) {
    print_r('Parent');
} elseif ($pid === 0) {
    sleep(10);
    file_put_contents('test.txt', 'Child');
    exit(0);
}

print_r('Final!');
```
What I see is fast printing "Final!" (without "Parent") and finishing script. In 10 seconds `test.txt` file appears with "Child!" content.

elfet: This is correct. Final mast go faster, no parent. Child in bg.

pluseg: @elfet So there is no need to promote child process with sending stats to session leader by calling`posix_setsid()`, right? The main process forks child for sending stats in background and goes further to "Final!" part without waiting a sending result. It seems correct way.

elfet: Yes, but if you write something like this file exec.php:
```php
<?php

exec('php test.php');
```
and then start and kill this process, test.txt will not be created, but with posix_setsid() will be.

I don't see and issues with creating separate session for sending data. What cons of that?

pluseg: Hmm, I'll test parent process killing case in the evening.

elfet: I think creating new session haven't any down cases, but implicitly says what going on.

pluseg: @elfet I guess that in case when sending stats fails with an exception it will close main process. BTW ideologically the leader process is one which executes deployer tasks not sends anonymous stats.

elfet: That is true. But sending stats in separate sid is okay.

pluseg: @elfet it's strange but `php.net` says that it promote one of process to leader and `man setsid` says that it creates totally new session and process group. Btw I returned `posix_setsid()` (see below https://github.com/deployphp/deployer/pull/1054/files#diff-aeff3570ce4587c4c14fc640f7a8a412R30)

elfet: Ok, thanks.

PR #1061, opened 2017-03-03
Add `--no-hooks` option for running commands without `before()` and `after()` (#934)

pluseg:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | Yes
| Fixed tickets | N/A

Updated version of https://github.com/deployphp/deployer/pull/934.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 ### Added
 - Add a way to retrieve a defined task [#1008](https://github.com/deployphp/deployer/pull/1008)
 - Add support for configFile in the NativeSsh implementation [#979](https://github.com/deployphp/deployer/pull/979)
+- Add `--no-hooks` option for running commands without `before()` and `after()` [#934](https://github.com/deployphp/deployer/pull/934)
 
 ### Changed
 - Autoload functions via Composer [#1015](https://github.com/deployphp/deployer/pull/1015)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 ### Added
 - Add a way to retrieve a defined task [#1008](https://github.com/deployphp/deployer/pull/1008)
 - Add support for configFile in the NativeSsh implementation [#979](https://github.com/deployphp/deployer/pull/979)
-- Add `--no-hooks` option for running commands without `before()` and `after()` [#934](https://github.com/deployphp/deployer/pull/934)
+- Add `--no-hooks` option for running commands without `before()` and `after()` [#1061](https://github.com/deployphp/deployer/pull/1061)
 
 ### Changed
 - Autoload functions via Composer [#1015](https://github.com/deployphp/deployer/pull/1015)
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -119,7 +119,6 @@ argument('stage', InputArgument::OPTIONAL, 'Run tasks only on this server or gro
 option('tag', null, InputOption::VALUE_OPTIONAL, 'Tag to deploy');
 option('revision', null, InputOption::VALUE_OPTIONAL, 'Revision to deploy');
 option('branch', null, InputOption::VALUE_OPTIONAL, 'Branch to deploy');
-option('no-hooks', null, InputOption::VALUE_NONE, 'Run task without after/before hooks');
 
 /**
  * Tasks
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -119,6 +119,7 @@ argument('stage', InputArgument::OPTIONAL, 'Run tasks only on this server or gro
 option('tag', null, InputOption::VALUE_OPTIONAL, 'Tag to deploy');
 option('revision', null, InputOption::VALUE_OPTIONAL, 'Revision to deploy');
 option('branch', null, InputOption::VALUE_OPTIONAL, 'Branch to deploy');
+option('no-hooks', null, InputOption::VALUE_NONE, 'Run task without after/before hooks');
 
 /**
  * Tasks
--- a/src/Task/GroupTask.php
+++ b/src/Task/GroupTask.php
@@ -50,12 +50,24 @@ class GroupTask extends Task
     }
 
     /**
-     * List of tasks names.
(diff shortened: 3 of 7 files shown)


src/Task/GroupTask.php
```diff
@@ -50,12 +50,24 @@ public function onlyOn($servers = [])
     }
 
     /**
-     * List of tasks names.
+     * List of tasks names with before and after hooks.
+     *
+     * @deprecated v5.0
```
pluseg: I offer deprecate this as useless.

src/Task/Task.php
```diff
@@ -227,6 +227,10 @@ public function setPrivate()
      */
     public function addBefore($task)
     {
+        if (!is_string($task)) {
+            throw new \RuntimeException('Invalid argument to `before` hook of "' . $this->getName() . '" task.');
+        }
```
pluseg: I mentioned that there is no check that $task is a string.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 ### Added
 - Add a way to retrieve a defined task [#1008](https://github.com/deployphp/deployer/pull/1008)
 - Add support for configFile in the NativeSsh implementation [#979](https://github.com/deployphp/deployer/pull/979)
+- Add `--no-hooks` option for running commands without `before()` and `after()` [#934](https://github.com/deployphp/deployer/pull/934)
```
elfet: #1061 – not ~934~

recipe/common.php
```diff
@@ -119,6 +119,7 @@
 option('tag', null, InputOption::VALUE_OPTIONAL, 'Tag to deploy');
 option('revision', null, InputOption::VALUE_OPTIONAL, 'Revision to deploy');
 option('branch', null, InputOption::VALUE_OPTIONAL, 'Branch to deploy');
+option('no-hooks', null, InputOption::VALUE_NONE, 'Run task without after/before hooks');
```
elfet: Move this to task command, as many users doesn't based on common recipe.

src/Task/Task.php
elfet: When it should be an InvalidArgumentException.

src/Task/Task.php
```diff
@@ -235,6 +239,10 @@ public function addBefore($task)
      */
     public function addAfter($task)
     {
+        if (!is_string($task)) {
+            throw new \RuntimeException('Invalid argument to `after` hook of "' . $this->getName() . '" task.');
```
elfet: Same here: InvalidArgumentException.

src/Task/ScriptManager.php
```diff
@@ -34,26 +39,34 @@ public function getTasks($name, $stage = null)
...
+
+            if ($this->isHooksEnabled()) {
+                $relatedTasks = array_merge($relatedTasks, array_map($collect, $task->getAfter()));
+            }
+
+            return $relatedTasks;
         };
 
-        $script = $collect($name);
+        $taskHierarchy = $collect($name);
```
elfet: Why are you rename this? 🙂

CHANGELOG.md
pluseg: @elfet I don't understand this rule. I think that the link must point to an issue or PR which describes the problem. Therefore I pointed out previous PR. Isn't it correct?

recipe/common.php
pluseg: @elfet Should I move only this options or all above?

src/Task/ScriptManager.php
pluseg: @elfet I refactored it as non obvious name. `$script` doesn't reflect the contents as for me and `$taskHierarchy` does. `$collect` returns exactly hierarchal collection of all related tasks. If you want I will return previous name or rename to something better.

CHANGELOG.md
elfet: I think it's petter to send to actual pr with implementation. Issues if is there is commit to master, but maybe better to send to commit (or both).

recipe/common.php
elfet: Only this one. Other used by common recipe. So if your deploy.php does not require common.php, i may no need in tag/rev/branch options.

src/Task/ScriptManager.php
elfet: But call name is `ScriptManager`. What way we define a new term inside deployer. I think it is cool)

CHANGELOG.md
pluseg: @elfet then I need to predict Pull Request's number to specify it in changelog.md? That's weird, but okay.

elfet: Kind of, or you can add more commit after pr creation. I think other contributors doing that way.

src/Task/ScriptManager.php
pluseg: @elfet Actually, I'd like to rename this class to TaskManager too :) It is responsible for managing tasks, not scripts. It isn't obvious for me as a newcomer. But of course it's up to you. If you want, I will return $script variable.

recipe/common.php
pluseg: @elfet Okay, I'll move it to TaskCommand. Meanwhile I think that it's okay to move all options to TaskCommand because of they are optional. Moreover, this way of defining options doesn't allow to add new ConsoleCommand with different arguments. I've already tried to add SSHCommand and faced that. I was not able to add required argument as there were defined options before. I guess that you will face it in version 5 refactoring.

src/Task/ScriptManager.php
elfet: Please return script) I don't like naming for everything manager-it, manager-that. All what we write some sort of managers.

recipe/common.php
elfet: Yes, i'm already started working on tasks args/options.

src/Task/ScriptManager.php
pluseg: Updated. Hmm, it's already called `manager` so I've just suggested to rename to another `manager` :) Meanwhile, it's returned to previous state.

PR #1067, opened 2017-03-07
Fix merge of string array config options

mbrodala:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

This unbreaks e.g. the following:

```php
add('shared_files', ['foo']);
add('shared_files', ['bar']);
```

Without this fix, this would become `['bar']` instead of `['foo', 'bar']`.


elfet:
Thanks a lot. You save me this this PR.

I'll give you contributor star ⭐️

PR #1072, opened 2017-03-08
Copy nested directories

gulios:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

Possibility to copy nested directories  ex:
set('copy_dirs', ['vendor', 'wp-content/plugins', 'wordpress']);

now it's not possible to copy for example: wp-content/plugins to the same location.

### DIFF
--- a/recipe/deploy/copy_dirs.php
+++ b/recipe/deploy/copy_dirs.php
@@ -16,11 +16,12 @@ task('deploy:copy_dirs', function () {
         foreach ($dirs as $dir) {
             $path = "{{deploy_path}}/releases/{$releases[0]}/$dir";
 
-            // Create destination dir(needed for nested dirs)
-            run("mkdir -p {{release_path}}/$dir");
-
             // Copy if dir exists.
             if (test("[ -d $path ]")) {
+
+                // Create destination dir(needed for nested dirs)
+                run("mkdir -p {{release_path}}/$dir");
+
                 run("rsync -av $path/ {{release_path}}/$dir");
             }
         }
--- a/recipe/deploy/copy_dirs.php
+++ b/recipe/deploy/copy_dirs.php
@@ -16,9 +16,12 @@ task('deploy:copy_dirs', function () {
         foreach ($dirs as $dir) {
             $path = "{{deploy_path}}/releases/{$releases[0]}/$dir";
 
+            // Create destination dir(needed for nested dirs)
+            run("mkdir -p {{release_path}}/$dir");
+
             // Copy if dir exists.
             if (test("[ -d $path ]")) {
-                run("cp -rpf $path {{release_path}}");
+                run("rsync -av $path/ {{release_path}}/$dir");
             }
         }
     }


elfet:
Why do you need this for?
Does rsync have same options on linux and bsd systems?

gulios:
We are using https://github.com/composer/installers  for wordpress instances which can install plugins not in vendors directory but in {{release_path}}/wp-content/plugins
Current version has:
`run("cp -rpf $path {{release_path}}"); `     
so when we set 
`set('copy_dirs', ['vendor', 'wp-content/plugins', 'wordpress']);` 
deployer copy files into 
`{{release_path}}/plugins` 
instead of 
`{{release_path}}/wp-content/plugins` 
so you can copy directories only in root of {{release_path}          

I think it's useful when your prepare releases like this(composer cache).
Options checked on OSX and GNU/Linux

elfet:
Ok, cool.

gulios:
Can you release v4.2.2 with that change?

Thx

elfet:
Will do.

nvaken:
I've just overruled this task with the following, to significantly speed up our deployment process. 

```
task('deploy:copy_dirs', function () {
  if (has('previous_release')) {
    foreach (get('copy_dirs') as $dir) {
      if (test("[ -d {{previous_release}}/$dir ]")) {
        run("mkdir -p {{release_path}}/$dir");
        run("cp -R {{previous_release}}/$dir/ {{release_path}}/$dir");
      }
    }
  }
});
```

gulios:
sorry @nvaken  but now i don't remember why it was rsync.... but probably there was a reason ;)
anyway copy nested directories was a main issue.

antonmedv:
Okay guys @nvaken @gulios can you make a PR with tests for cp/rsync copy_dirs with nested support? Will be cool to solve this mystery :)

nvaken:
I will not be able to create a new PR or test coming days or weeks, sorry about that. 

Though I did just check out what this specific PR replaced in the original piece of code, which was `cp -rpf`. I'm no Linux *guru* by all means, but I do think those where [reasonable flags](https://explainshell.com/explain?cmd=cp+-rpf) which could be placed back into place. The new `mkdir -p` does make sense as well, I'm using that in our new setup as well to make sure the `$dir` actually exists and it's complete range of parents. I could imagine that the absence of `mkdir -p` could cause the _nested directory_ problems as @gulios mentioned, though I'm still not sure what `rsync` would solve in nested file structures? But, again, I'm no Linux *guru*.

Based on this, I'd suggest the following and testing it before implementing it. 
```
run("mkdir -p {{release_path}}/$dir");
run("cp -rpf {{previous_release}}/$dir/ {{release_path}}/$dir");
```

gulios:
ok, i checked it quickly and i remind that was a problem with wrong copying nested directories content.

When you use "cp" ex:
`
mkdir -p /path/releases/2/wp-content/plugins
cp -R /path/releases/1/wp-content/plugins /path/releases/2/wp-content/plugins
`

it won't copy content from source "plugins" to destination "plugins" but copy whole dir "plugins" so you will have /path/releases/2/wp-content/plugins/plugins

With "cp -rpf" same problem.
Probably that was a reason i used rsync which copy content of source dir to destination.
Second reason is that i didn;t want to modify body of method and change value of "$dir" because if you want use copy in nested dirs you have to explode('/', $dir) and remove last element. IMO we shouldn't do that but if you want ;) maybe there is real value from speed in rsync vs cp but in my case i don;t see much difference.

nvaken:
Isn't that simply a matter of adding a trailing slash? See below a small PoC:

```
# Creating dummy data, nested source folder with nested folder in it. Also add a destination folder.
$ mkdir -p src/foo/bar/test/nested
$ mkdir -p dest
$ echo 'Test content' > src/foo/bar/test/dummy{0001..0003}.txt
$ echo 'Test content' > src/foo/bar/test/nested/dummy{0004..0007}.txt

# Actual copying, PoC. src/foo/bar/test/ being the source folder to copy.
$ mkdir -p dest/foo/bar/test
$ cp -rpf src/foo/bar/test/ dest/foo/bar/test/

# Resulting contents of dest/foo/bar/test/
$ ls -1 dest/foo/bar/test
dummy0001.txt
dummy0002.txt
dummy0003.txt
nested
```
_Note that if `cp` would have been without trailing slashes, we'd create a new nested `test` folder_

Still, I'm no expert on this area, so I might be wrong! :)

# Performance
We're warming up our new release folders by copying the composer vendor packages so only changes have to be installed / updated by composer. We've got a pretty massive amount of composer packages and while using `rsync` this copy will take ~60s, but if using `cp` we're around ~35s. Run on a Macbook (Retina, 13-inch, Early 2015) 2.9 GHz Intel Core i5 with Flashdrive, 16 GB 1867 MHz DDR3.

antonmedv:
Actually will be cool to go back to `cp`.

nvaken:
Hm... I just noticed that adding a trailing slash might cause some issues in different situations. I _think_ the advised way to go is making sure the source folder does not yet exist in the destination though the parents of the folder should. So, something like:

```
mkdir -p dest/foo/bar 
cp -rpf src/foo/bar/test dest/foo/bar/test
```

So, I think this is the way that @gulios suggested in his latest comment above. Though, I still think it's worth exploding that path?

antonmedv:
Performance is important. So I think it's worth. And I feel better of `cp` as it more common in diff environments.

gulios:
try add -W option to rsync

-a is for archive, which preserves ownership, permissions etc.
-v is for verbose, so I can see what's happening (optional)
-h is for human-readable, so the transfer rate and file sizes are easier to read (optional)
-W is for copying whole files only, without delta-xfer algorithm which should reduce CPU load
--no-compress as there's no lack of bandwidth between local devices
--progress so I can see the progress of large files (optional)

I've tested it and it looks like rsync is faster but it depends on disk usage etc...

103M	vendor/
time cp -rpf  /path/current/vendor/ /path/gulios/vendor

real	0m26.136s
user	0m0.156s
sys	0m8.140s


time rsync -avhW /path/current/vendor/ /path/gulios/vendor

sent 77.02M bytes  received 220.49K bytes  3.03M bytes/sec
total size is 76.24M  speedup is 0.99

real	0m24.746s
user	0m0.556s
sys	0m16.500s

nvaken:
Interesting. Would that not only be a performance boost if the destination folder + files already (partially) exist? When there is no destination existant, will it not just skip the "delta-xfer algorithm"? Not sure and I'm only basing this on "common sense". 

My results. _Note_: I'm making sure `new_vendor` does not exist before running these tests, since that is what we're doing when building a new release; the destination folder does not yet exist.

## rsync without -W flag
~23s
```
$ gtime rsync -avh vendor new_vendor
sent 33.63M bytes  received 200.31K bytes  1.44M bytes/sec
total size is 32.98M  speedup is 0.98
0.55user 4.69system 0:23.18elapsed 22%CPU (0avgtext+0avgdata 19021824maxresident)k
1inputs+565outputs (0major+6601minor)pagefaults 0swaps
```

## rsync with -W flag
~26s
```
$ gtime rsync -avhW vendor new_vendor
sent 33.63M bytes  received 200.31K bytes  1.28M bytes/sec
total size is 32.98M  speedup is 0.98
0.56user 4.81system 0:26.18elapsed 20%CPU (0avgtext+0avgdata 19316736maxresident)k
3inputs+647outputs (0major+7593minor)pagefaults 0swaps
```

## cp -rpf
~14s
```
$ gtime cp -rpf vendor new_vendor
0.28user 4.01system 0:14.60elapsed 29%CPU (0avgtext+0avgdata 10125312maxresident)k
0inputs+601outputs (0major+841minor)pagefaults 0swaps
```

Still seems like a major boost in performance using `cp` for me? And even stranger is that `-W` seemed to be a bit slower for me, although that might be just an anomaly and caused by my system. Are we having different results maybe due to a difference in OS or hardware @gulios? 

Other then that, as mentioned by @antonmedv, will `rsync` be a more common cross platform compatibility issue then `cp`? I'm not experienced enough to answer this question.

gulios:
I removed destination folder. It was clean test. For performance it can be a lot of things like: rsync version, cpu, disk etc so it;s hard to say. As i know it's fast on clean copy because there is no files to compare.
I think you should create PR with changes to cp if you want. For me rsync is fine and we don't need modify $dir variable special for that... but if it's ok for @antonmedv just change code to cp but please do not remove "copy nested directories" functionality ;)

nvaken:
I think it might be wise to first check out the following before altering anything: 

1. Does `cp` actually gain a performance boost? We should be able to figure that out, platform wide. Now we're seeing too much inconsistencies in our results. 
2. Is `rsync` widely enough supported to be safely used into a project this big? 
3. Check if both points are beneficial enough for `rsync` to be used in favor of a system command like `cp`.

I hope someone else is able to figure out 1 and 2, I'm a bit out of ideas _and_ time. ;-)

nvaken:
Just created a issue (see above) to make sure this is being followed up.

Issue #1073, opened 2017-03-09
Parse hyphens in environment setting names


MrHash:
| Q                 | A
| ----------------- | ---
| Issue Type        | Bug
| Deployer Version  | N/A
| Local Machine OS  | N/A
| Remote Machine OS | N/A

### Description 
*If you're reporting a bug, please include following information*

### Steps to reproduce
```
set('hyphenated-var', 'value');
parse('{{hyphenated-var}}'); 
```
Result is `{{hyphenated-var}}` when expected is `value`.

elfet:
Let's leave only PR.

PR #1074, opened 2017-03-09
Parse hyphens in environment setting names

MrHash:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #1073


elfet:
Why do you need this for? Why not some_name?

MrHash:
I have trying an integration with docker which has runnable binaries such`docker-machine` and `docker-compose`. For clarity I would like to use the actual command names and i don't see any downside to that.

Issue #371, opened 2015-06-19
ask function: Call to a member function getOutput()


puxiaokang:
Just for secure, I can't write host password in code. So I try ``ask`` function.

when I call ``ask`` or ``askHiddenResponse`` for ask password, the function throw this error:
```
Fatal error: Call to a member function getOutput() on a non-object in vendor/deployer/deployer/src/functions.php on line 511
```
*example:*
```
<?php
require 'recipe/symfony.php';
$password = askHiddenResponse("Host password: ");

server('prod', 'HOST', 22)
     ->user('USER')
     ->password($password);
```

If ``ask`` in ``task`` callback, It's success. but I can set server.password:
```
task('auth', function(){
    $password = askHiddenResponse("Host password: ");
    env('server.password', $password);
});
before('deploy', 'auth');

-- throw exception
The parameter `server.password` cannot be set, because `server` is protected.
```

elfet:
You can specify password to `null`:
```php
server('prod', 'HOST', 22)
     ->user('USER')
     ->password(null);
```
And you will be asked for password.

@ZhukV can you add docs about this functionality?

puxiaokang:
Thanks

staabm:
is `ask` and friends not supposed to be run at the top level?
(as I ran into the same error as I dont invoke it within a task it seems).

if this is the case it would be great to throw a proper exception which tells the dev that the function/method he is trying to use requires a `Context` (which means needs to be called within a `task()`)

elfet:
> if this is the case it would be great to throw a proper exception which tells the dev that the function/method he is trying to use requires a Context (which means needs to be called within a task())
This will be cool. Can you create PR with this?

elfet:
> if this is the case it would be great to throw a proper exception which tells the dev that the function/method he is trying to use requires a Context (which means needs to be called within a task())

@staabm this will be cool. Can you create PR with this?

staabm:
Will do

PR #1083, opened 2017-03-10
Provide a usefull error when ask*() is not used wihtin a task()

staabm:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

before this change a newcomer to this project sees errors like 
```
PHP Fatal error:  Uncaught Error: Call to a member function getOutput() on boolean in /cluster/www/www/www/tools/deployer/src/functions.php:677
Stack trace:
#0 /cluster/www/www/www/tools/deployer/src/functions.php(685): Deployer\output()
#1 /cluster/www/www/www/tools/deployer/src/functions.php(567): Deployer\isQuiet()
#2 /deploy/kunzmann/stage/rocket/deploy.php(51): Deployer\ask('Zielstage (prod...')
#3 /cluster/www/www/www/tools/deployer/bin/dep(106): require('/deploy/kunzman...')
#4 /cluster/www/www/www/tools/deployer/bin/dep(107): {closure}()
#5 {main}
  thrown in /cluster/www/www/www/tools/deployer/src/functions.php on line 677
```

when using a `ask()` function within global scope instead of a `task()`.

After this change the error turns into

```
PHP Fatal error:  Uncaught Deployer\Exception\Exception: 'Deployer\ask' can only be used within a 'task()'-function! in /cluster/www/www/www/tools/deployer/src/functions.php:757
Stack trace:
#0 /cluster/www/www/www/tools/deployer/src/functions.php(567): Deployer\requiresTaskContext('Deployer\\ask')
#1 /deploy/kunzmann/stage/rocket/deploy.php(51): Deployer\ask('Zielstage (prod...')
#2 /cluster/www/www/www/tools/deployer/bin/dep(106): require('/deploy/kunzman...')
#3 /cluster/www/www/www/tools/deployer/bin/dep(107): {closure}()
#4 {main}
  thrown in /cluster/www/www/www/tools/deployer/src/functions.php on line 757
```

based on a "discussion" in https://github.com/deployphp/deployer/issues/371#issuecomment-285592637

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -16,6 +16,8 @@ use Deployer\Task\Task as T;
 use Deployer\Task\Context;
 use Deployer\Task\GroupTask;
 use Deployer\Type\Result;
+use Deployer\Cluster\ClusterFactory;
+use Deployer\Exception\Exception;
 use Monolog\Logger;
 use Symfony\Component\Console\Question\ChoiceQuestion;
 use Symfony\Component\Console\Question\ConfirmationQuestion;
@@ -23,7 +25,6 @@ use Symfony\Component\Console\Question\Question;
 use Symfony\Component\Finder\Finder;
 use Symfony\Component\Process\Exception\ProcessFailedException;
 use Symfony\Component\Process\Process;
-use Deployer\Cluster\ClusterFactory;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
@@ -560,6 +561,8 @@ function has($name)
  */
 function ask($message, $default = null, $suggestedChoices = null)
 {
+    requiresTaskContext(__FUNCTION__);
+
     if (($suggestedChoices !== null) && (empty($suggestedChoices))) {
         throw new \InvalidArgumentException('Suggested choices should not be empty');
     }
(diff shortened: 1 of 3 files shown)


staabm:
This small new function can obviously be used in more places but I thought this would be a good start.

elfet:
Cool.  i think this is helpful.

elfet:
Please update changelog.

staabm:
done.

src/functions.php
```diff
@@ -742,3 +751,16 @@ function parse($value)
...
     return Context::get()->getEnvironment()->parse($value);
 }
+
+/**
+ * Throws a Exception when not called within a task-context.
+ *
+ * This method provides a usefull error to the end-user to make him/her aware
+ * to use a function in the required task-context.
+ */
+function requiresTaskContext($callerName)
```
elfet: I think better to create method in `Context`:

```php
Context::required();
```

staabm: also thought about that option. will change it.

staabm: now I know why I decided against a method in Context ;-):

the message here is rather task specific and `Context` class atm is rather generic.

```
    /**
     * Throws a Exception when not called within a task-context.
     *
     * This method provides a usefull error to the end-user to make him/her aware
     * to use a function in the required task-context.
     *
     * @throws Exception
     * @return Context
     */
    public static function required($callerName)
    {
        $ctx = self::get();
        if (!$ctx) {
            throw new Exception("'$callerName' can only be used within a 'task()'-function!");
        }
        return $ctx;
    }
```

do you still think we should move it to Context, I am fine with any solution you will accept.

elfet: I think its better.

staabm:
adjusted per feedback

PR #1084, opened 2017-03-10
Pretty-print uncaught exceptions with SymfonyStyle

staabm:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

Pretty print uncaught exceptions using SymfonyStyle. Those exceptions occur e.g. when the enduser does not respect phpdocs and uses wrong argument types etc.

repro:
```
// inside your deploy.php
before('does_not_exist_task', 'do_something');
```

before
image

after
image

### DIFF
--- a/bin/dep
+++ b/bin/dep
@@ -99,6 +99,15 @@ $input = new \Symfony\Component\Console\Input\ArgvInput();
 $output = new \Symfony\Component\Console\Output\ConsoleOutput();
 $deployer = new \Deployer\Deployer($console, $input, $output);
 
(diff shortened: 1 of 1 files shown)


staabm:
I created this patch for UX reasons. For future development it would be cool to use SymfonyStyle (or a similar self-brewed style) across all commands.

see http://symfony.com/blog/new-in-symfony-2-8-console-style-guide

elfet:
Nice.

PR #1085, opened 2017-03-10
Show a progressbar while ssh upload

staabm:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

Adds a progressbar to the scp-copy steps.

before:
the deploy progress was not visible and the running command took minutes before actual progress is visible

after::
image

see also http://symfony.com/doc/current/components/console/helpers/progressbar.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Add a way to retrieve a defined task [#1008](https://github.com/deployphp/deployer/pull/1008)
 - Add support for configFile in the NativeSsh implementation [#979](https://github.com/deployphp/deployer/pull/979)
 - Add `--no-hooks` option for running commands without `before()` and `after()` [#1061](https://github.com/deployphp/deployer/pull/1061)
+- Add a progressbar while ssh upload [1085](https://github.com/deployphp/deployer/pull/1085)
 
 ### Changed
 - Parse hyphens in environment setting names [#1073](https://github.com/deployphp/deployer/pull/1074)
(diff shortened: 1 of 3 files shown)


staabm:
hmm locally it works.. need to figure out why travis is failing

staabm:
Ready to go

elfet:
This is cool But i think to refactor upload task and upload everything gzipped (to speed up upload). No progress bar when will be able.

staabm:
cool, lets wait for the gzip refactor then. 

Faster up/down-load is higher priority

staabm:
Can I assist somehow? Are the details documented somewhere how the gzipped stuff should be implemented?

staabm:
do we talk about the `-C` scp option?

```
 -C      Compression enable.  Passes the -C flag to ssh(1) to
         enable compression.
```

or do you have something more sophiscated in mind?

elfet:
I was thinking about rsync, but i don't know if is installed on most users machines.
Also i was thinking about implementing it via ssh only `tar ... | ssh ... tar ...`.

staabm:
Any particular reason against `-C`? When The posted stack overflow answers are correct this should have the same effect as using zip and a ssh pipe.

Switching to rsync could be considered a BC break as it changes required dependencies

elfet:
Need to investigate it. Can you create benchmark for all 3 methods?

staabm:
Will do tomorrow

staabm:
I am on vacation for the rest of the week.. will check after that

elfet:
@staabm not needed any more a4beca7e078f7314c443c211aa5d9fefd1df0d36

PR #1121, opened 2017-03-22
Add a "find" mode

GaryPEGEOT:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | xx

Add a "find" mode, that prevent errors when writables folders are not empty, with files no-writables

### DIFF
--- a/recipe/deploy/writable.php
+++ b/recipe/deploy/writable.php
@@ -82,6 +82,8 @@ task('deploy:writable', function () {
             } else {
                 throw new \RuntimeException("Cant't set writable dirs with ACL.");
             }
+        } elseif ($mode === 'find') {
+            run("$sudo find $dirs -type d -exec chmod {{writable_chmod_mode}} {} \;");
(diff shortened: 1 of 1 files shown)


antonmedv:
What are use cases?

GaryPEGEOT:
If, for example, a writable directory contain a file, created by a different user that the server, chmod mod will fail:
If a www-data user has created a file in a directory, chmod -R 775 my_dir with user "testuser" will fail.

Using find allow recursivity without changing permissions to files in it.

antonmedv:
What about using sudo?

GaryPEGEOT:
In my case sudo is not an option (Debian server without sudo installed), and root either.

antonmedv:
For now it's only case in your situation, and you can simple override task. I don't think what this should be in common recipe.

PR #1132, opened 2017-03-26
Removed dependecy on the man binary

mjrider:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | No
| BC breaks?    | No
| Deprecations? |No
| Fixed tickets | N/A 

Barebone production environments often don't have the man binary or the
don't have man pages installed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 - Only native ssh client for now [#1092](https://github.com/deployphp/deployer/pull/1092)
 - Task `current` to `config:current` [#1092](https://github.com/deployphp/deployer/pull/1092)
 - `onFailure` to `fail` [#1092](https://github.com/deployphp/deployer/pull/1092)
+- Removed the dependency on the man binary
 
 
 ## v4.3.0
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -63,10 +63,10 @@ set('clear_use_sudo', false);    // Using sudo in clean commands?
 set('cleanup_use_sudo', false); // Using sudo in cleanup commands?
 
 set('use_relative_symlink', function () {
-    return test('[[ "$(man ln)" =~ "--relative" ]]');
+    return test('[[ "$(ln --help)" =~ "--relative" ]]');
 });
 set('use_atomic_symlink', function () {
-    return test('[[ "$(man mv)" =~ "--no-target-directory" ]]');
+    return test('[[ "$(mv --help)" =~ "--no-target-directory" ]]');
 });
 
 set('composer_action', 'install');


antonmedv:
On macOS i got this:
```
mv --help; echo exit code: $?
mv: illegal option -- -
usage: mv [-f | -i | -n] [-v] source target
       mv [-f | -i | -n] [-v] source ... directory
exit code: 64
```

antonmedv:
If you don't have `man` on server, you can use this:
```
 set('use_relative_symlink', true);
 set('use_atomic_symlink', true);
```
Maybe it's better to add it to docs instead of code.

mjrider:
@antonmedv i'll ask a college tomorrow if he has an idea about testing this on macos without parsing man pages.

mjrider:
found another problem when i was testing some ideas
```shell
 $ ./bin/dep -vvv test
➤ Executing task test
[localhost] > if [[ "$(ln --help )" =~ "--relative" ]]; then echo 'true'; fi
[localhost] < sh: 1: [[: not found
[localhost] > if [[ "$(mv --help ; true )" =~ "--no-target-directory" ]]; then echo 'true'; fi
[localhost] < sh: 1: [[: not found
✔ Ok [8ms]
```
bash:
```shell
 $ if [[ "$(mv --help ; true )" =~ "--no-target-directory" ]]; then echo 'true'; fi
true
dash:
```shell
$ if [[ "$(mv --help ; true )" =~ "--no-target-directory" ]]; then echo 'true'; fi
sh: 3: [[: not found
```

contents of deploy.php
```php
namespace Deployer;

task('test',function(){
	test('[[ "$(ln --help )" =~ "--relative" ]]');
	test('[[ "$(mv --help ; true )" =~ "--no-target-directory" ]]');
})->local();
```

The problem is dat /bin/sh is an symlink to dash, which is kinda default in debian/ubuntu

antonmedv:
> The problem is dat /bin/sh is an symlink to dash, which is kinda default in debian/ubuntu
In v5 solved this by calling bash directly.

mjrider:
@antonmedv Would this be an acceptable solution?

```
if [[ $(mv -h 2>&1 || mv --help 2>&1) =~ '--no-target-directory' ]] ; then echo 'true' ; fi
```
true

antonmedv:
What is so complicated. Does man package unaccessable on a lot of systems?

staabm:
maybe use the current behaviour at first and fallback to the suggested approach using something like?

```
if [[ $(man mv 2>&1 || mv --h 2>&1) =~ '--no-target-directory' ]] ; then echo 'true' ; fi
```

so we are sure we keep BC as we use `man`  as primary source as before

staabm:
alternatively build a small php-api

```php
/*
 example usage: cmdSupportsOpt('mv', '--no-target-directory');
*/
function cmdSupportsOpt($command, $optString) {
  return test('[[ "$(man '. $command .' 2>&1 || '. $command .' -h)" =~ "$optString" ]]');
}

```

mjrider:
@antonmedv clean(lean?) system images don't come with man installed
just checked an base debian install, and it doesn't have man installed
also, it looks like macos doens't support --no-target-directory, do we need macos support for this test?

antonmedv:
Okay, some sort of function will be okay.

antonmedv:
Done a1806cb77bb430bcabcdf837d84d70b9f0e72b38

PR #1133, opened 2017-03-29
fix laravel artisan:queue:restart task

zakirullin:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

Launching ```artisan:queue:restart``` command in current release's folder has no effect - old workers still running in old release's context and checking for ```'illuminate:queue:restart'``` cache key to know when to stop. By default framework's cache stored at files, so we must set cache key ``` 'illuminate:queue:restart'```in old release's folder.
When cache driver is different than default ```'file'```, queue will restarts too.

### DIFF
--- a/recipe/laravel.php
+++ b/recipe/laravel.php
@@ -100,7 +100,11 @@ task('artisan:optimize', function () {
 
 desc('Execute artisan queue:restart');
 task('artisan:queue:restart', function () {
-    run('{{bin/php}} {{release_path}}/artisan queue:restart');
+    $releases = get('releases_list');
+    if (isset($releases[1])) {
+        $previousReleasePath = "{{deploy_path}}/releases/{$releases[1]}";
+        run("{{bin/php}} $previousReleasePath/artisan queue:restart");
+    }
 });
 
 /**


antonmedv:
Maybe it's a bug in laravel? Why `{{release_path}}/artisan queue:restart` doesn't restart workers?

zakirullin:
Nope.
Simplifying things (snapshot right before ```artisan:queue:restart```) :
```releases/old/storage/framework/cache/xxx```
```releases/new/storage/framework/cache/xxx```

Snapshot after ```artisan:queue:restart```:
releases/old/storage/framework/cache/xxx
releases/new/storage/framework/cache/xxx -> contains key-value pair (```illuminate:queue:restart => 1```), set by ```artisan:queue:restart```. 

But the problem is, workers still have old code, and they will not die until ```illuminate:queue:restart``` is set in ```releases/old/storage/framework/cache/```

antonmedv:
Does your shared files contains `storage/framework/cache` too?
* releases/old/storage/framework/cache/xxx
* releases/new/storage/framework/cache/xxx
Should be same.

zakirullin:
Ahh, we don't share cache between releases for our reasons, so it leads this behaviour.
With default laravel reciepe no problem as cache is shared.

antonmedv:
Ok.

PR #1143, opened 2017-04-04
create initial shared file

vinpel:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | Yes
| BC breaks?    |  No
| Deprecations?  No
| Fixed tickets | N/A

### Yii2 advanced app receipe :
The shared step create empty config in shared & link.
After that  the "init" phase create the config file in their place.

I've inversed the order to be able for the 1st deployment a valid config file without manual edition.

### Shared /deploy/shared.php
When a shared_file is not present in shared, but present in release we use it as reference.

An exemple for yii2 receipe :

-> Init phase create for exemple "/common/config/main-local.php"
-> Shared phase **now** copy this file in the shared dir, remove it from release and create the link.
The file is now non empty in shared and the yii2 app is functional

If the file exist in shared, no copy and same as usal.


recipe/deploy/shared.php
```diff
@@ -47,15 +47,22 @@
...
+
+        // Create dir of shared file
+        run("mkdir -p $sharedPath/" . $dirname);
+
+        // Check if shared file does not exists in shared.
+        // and file exist in release
+        if (!test("[ -f $sharedPath/$file ]") && test("[ -f {{release_path}}/$file ]")) {
+            // Copy file in shared dir if not present
+            run("cp -rv {{release_path}}/$file $sharedPath/$file");
+           }
```
antonmedv: Please fix

antonmedv:
And please add to changelog.

PR #1144, opened 2017-04-05
SSH Command expects optional parameter and use default instead

stevenklar:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

Make SshCommand host parameter optional when only using the default (first) host.

> dep ssh

### DIFF
--- a/src/Console/SshCommand.php
+++ b/src/Console/SshCommand.php
@@ -39,7 +39,7 @@ class SshCommand extends Command
     {
         $this->addArgument(
             'hostname',
-            InputArgument::REQUIRED,
+            InputArgument::OPTIONAL,
             'Hostname'
         );
     }
@@ -49,7 +49,10 @@ class SshCommand extends Command
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        $host = $this->deployer->hosts->get($input->getArgument('hostname'));
+        $firstHostname = $this->deployer->hosts->getIterator()->current()->getHostname();
+        $hostname = $input->getArgument('hostname') ?? $firstHostname;
+
+        $host = $this->deployer->hosts->get($hostname);
         Context::push(new Context($host, $input, $output));
 
         $options = $host->sshOptions();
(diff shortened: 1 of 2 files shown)


src/Console/SshCommand.php
```diff
@@ -49,7 +49,10 @@ protected function configure()
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        $host = $this->deployer->hosts->get($input->getArgument('hostname'));
+        $firstHostname = $this->deployer->hosts->getIterator()->current()->getHostname();
```
staabm: I guess this only makes sense when there is actually only 1 host, right?

stevenklar: Yes. That's correct.
If multiple hosts are given the first one will be choosen.

staabm: what I meant: this fallback to the first hostname only makes sense where there is just one defined.

using the first when several are defined doesnt sound like a usefull guess.

stevenklar: What does sound like a good guess for you?
First thing which comes in mind for me would be an addtional configuration.

default_stage for example.

staabm: atm I can't think of a good guess. using the only one defined sounds straight forward.

in  a interactive mode you might "ask the user" to choose from the defined list, but non-interactive I dont think you can pre-dict which host is the "correct" one.

stevenklar: Would you see that interactive mode necessary or could that be an enhancement
`dep ssh -i`

staabm: lets see what @antonmedv thinks about such a feature

stevenklar: Allright. Maybe short explaination why I came up with this simple feature.

When using `dep ssh` I expected initial that it directly connects me instead of writing `dep ssh user@my-little-pony.de`all the time.

antonmedv: I think it's ok only if there only one host. Otherwise it's implicitly, and explicitly is always better. Inveractive mode maybe solution, but i don't like -i option, maybe it's better to be by default:
* dep ssh - interactive
* dep ssh host

stevenklar: I don't quite understand the `dep ssh - interactive`path.
User is typing `dep ssh - interactive`and get a list of hosts?

That would be an enhancement for another pull request. Would it fit with an additional "check if there is only one host registered"?

antonmedv: `dep ssh` - leads to interactive choose

antonmedv:
Done 7f74425d886aceddba3381597762c072316a0330

Issue #1086, opened 2017-03-10
Laravel overriding public folder not symlinking properly


garygreen:
If you've overridden the public path in laravel from the standard `public` to e.g. `public_html` -- the symlinking for the public storage won't work.

Laravel actually provides a command `php artisan storage:link` that can link public to the storage folder (it will know that you've overriden the path) -- maybe that should be used rather than [hardcoding it](https://github.com/deployphp/deployer/blob/master/recipe/laravel.php#L117) ?

elfet:
What will be cool. Can you make a PR with this?

barryvdh:
Would be just 
```php
desc('Execute artisan optimize');
task('artisan:storage:link', function () {
    run('{{bin/php}} {{release_path}}/artisan storage:link');
});
```

But it does have different checks: https://github.com/laravel/framework/blob/5.4/src/Illuminate/Foundation/Console/StorageLinkCommand.php

barryvdh:
See https://github.com/deployphp/deployer/pull/1152

PR #1152, opened 2017-04-07
Add storage:link option

barryvdh:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #1086

This add `artisan:storage:link`. We could add it by default, by adding it after `deploy:writable`. It only creates the link when no dir exists.


antonmedv:
I think ok by default. Please update changelog.

barryvdh:
Added by default + updated changelog

antonmedv:
Thanks.

Issue #1165, opened 2017-04-12
Allow to configure multiplexing


antonmedv:
| Q                 | A
| ----------------- | ---
| Issue Type        | Feature Request
| Deployer Version  | 5.x

### Description 

Allow to configure multiplexing with host options.

```php
host(...)
    ->options(['ControlPersist=360'])
```

mikeymike:
@antonmedv are you open to PRs on this ? (just wanted to make sure you wasn't working on it atm)

antonmedv:
No, i don't working on this right now.

Issue #1155, opened 2017-04-08
Get stuck when using multiplexing


leo108:
| Q                 | A
| ----------------- | ---
| Issue Type        | Bug
| Deployer Version  | master cdac931
| Local Machine OS  | Ubuntu 16.04.1 LTS/ homestead
| Remote Machine OS | Ubuntu 16.04.2 LTS

### Description 
See deploy file and output log

### Content of `deploy.php`
```php
<?php
namespace Deployer;

host('xxx.com')
    ->user('root')
    ->multiplexing(true)  // it works as expected if set this to false
    ->identityFile('/home/vagrant/.ssh/id_rsa');

task('test', function () {
    run('echo 123;');
});
```

### Output log

```
➤ Executing task test
[xxx.com] > echo 123;
[xxx.com] < ssh multiplexing initialization
[xxx.com] < 123        # output 123 within 1s
• done on [xxx.com] # output this line after 1min
✔ Ok [60s 804ms]
```

antonmedv:
So what is the problem?

leo108:
@antonmedv it takes 60 seconds to finish the `test` task when enabling the multiplexing, but it only takes 800ms when disabling the multiplexing.

leo108:
I changed the test task to this:

```
task('test', function () {
    run('echo 123;');
    run('echo 321;');
});
```

the output is:

```
➤ Executing task test
[xxx.com] > echo 123;
[xxx.com] < ssh multiplexing initialization
[xxx.com] < 123
[xxx.com] > echo 321;
[xxx.com] < ssh multiplexing initialization
[xxx.com] < 321
• done on [xxx.com]
✔ Ok [121s 522ms]
```

It takes 2 min to finish this simple task, don't you think there is something wrong?

antonmedv:
Looks like multiplexing isn't initialized correctly. Please chech if it's working on your local machine.

leo108:
I did some tests and find something strange

```
$cmd = 'ssh -i /home/vagrant/.ssh/id_rsa -o ControlMaster=auto -o ControlPersist=60 -o ControlPath=~/.ssh/deployer_mux_root@xxx.com root@xxx.com echo 1;';
$p = new Symfony\Component\Process\Process($cmd);
$p->run(function() {var_dump(func_get_args());});
```

it outputs

```
array:2 [
  0 => "out"
  1 => "1\n"
]
```

and then stuck there for 1min, then outputs error message:

```
[...]

PR #1172, opened 2017-04-18
Feature/configurable ssh multiplexing

mikeymike:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #1165 

I wanted more structure to the data and to be able to configure not just the options but all areas of the ssh flags/options as this really helps with debugging problems (and how I found the issue with #1155 )

To do this I created an immutable ssh arguments class that handles pretty much everything including the multiplex arguments. This was very important to me because these arguments may need to be used elsewhere and not just in the client, for example the `rsync` recipe will currently not be able to use the multiplexing unless the code is duplicated into that recipe. 

I also updated the tests where applicable including the yaml fixture with the new format I'm proposing. 

Hopefully it makes sense, I'm open to feedback and I have some changes waiting to update the `rsync` recipe if this gets merged. 

As a side note, IMO multiplexing should be enabled by default, but this is a separate discussion we can have.


src/Host/Host.php
```diff
@@ -202,31 +192,33 @@ public function multiplexing(bool $multiplexing = true)
...
-    public function addOption(string $option)
+    public function sshFlags(array $flags) : Host
+    {
+        $this->sshArguments = $this->sshArguments->withFlags($flags);
+        return $this;
+    }
+
+    public function addSshOption(string $option, $value) : Host
+    {
+        $this->sshArguments = $this->sshArguments->withOption($option, $value);
```
antonmedv: I don't like this approach (btw i like immutability). Why dot just call `$this->sshArguments->addOption($option, $value);`?

mikeymike: My idea behind this was to keep the arguments for the host from being altered accidentally outside of the class. E.g if a task called getSshArguments then added more flags / options without immutability this would change the sshArguments on the host object. 

If you don't think this is something we need to be concerned with it's a relatively simple change I can do.

mikeymike:
Example of my current usage

```php
host('xxx')
    ->port(xxxx)
    ->stage('dev')
    ->user('www-data')
    ->set('branch', 'develop')
    ->set('keep_releases', 1)
    ->set('deploy_path', 'xxxxx')
    ->multiplexing()
    ->addSshFlag('-q')
    ->addSshOption('UserKnownHostsFile', '/dev/null')
    ->addSshOption('StrictHostKeyChecking', 'no')
    ->addSshOption('ControlPersist', '600');
```

It's also possible to do the options as an array if the user would want to.

Issue #1175, opened 2017-04-18
Cleanup task not keeping releases correctly


mikeymike:
| Q                 | A
| ----------------- | ---
| Issue Type        | Bug
| Deployer Version  | master
| Local Machine OS  | N/A
| Remote Machine OS | N/A

### Description 
Example of the bug can be seen on https://3v4l.org/cXYMq

When you only want to keep 1 release it deletes it, whilst keeping more releases it seems to keep 1 less than required

mikeymike:
As mentioned in the PR #1176 my use case separates code deployment and release symlink into two separate commands because our `release` could be a DB upgrade release or may just be an atomic release, so a deploy may look something like below

```
vendor/bin/dep deploy production
vendor/bin/dep release:upgrade production
```

With this in mind the second command includes the new release in the release list while common deploys with one command will use the cache from a `get('release_list')` call from earlier in the task list. 

In my opinion this caching is causing unexpected behaviour ... what do you think ?

antonmedv:
I agree. I think we need to update releases list cache after new release and use your pr.

mikeymike:
What are your thoughts on defining if it's cachable on the `set` method .. so it's signature would change to `set(name, value, cacheable = true)` 

This way there are things that never change during deployment that can be cached, e.g. binary locations etc while other closures can be set to run each time they are called by setting cacheable to false

antonmedv:
No, i think it will make stuff more complicated. I just was thinking about adding this to `deployer:release`:
```php
add('releases_list', $new_release);
```

mikeymike:
Ahh yeah that makes sense, I'm still learning the codebase so was unaware of that function 😄 , I'll update and reopen that PR with this change later today 👍

antonmedv:
Fixed.

PR #1176, opened 2017-04-18
Cleanup keeps the right amount of releases

mikeymike:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  |  No
| BC breaks?    |  No
| Deprecations? |  No
| Fixed tickets | #1175 

Pulls out the releases to keep and then gets a diff to show the ones that need to be removed.

### DIFF
--- a/recipe/deploy/cleanup.php
+++ b/recipe/deploy/cleanup.php
@@ -18,12 +18,10 @@ task('cleanup', function () {
         return;
     }
 
-    while ($keep - 1 > 0) {
-        array_shift($releases);
-        --$keep;
-    }
+    $keeps   = array_slice($releases, 0, $keep);
+    $removes = array_diff($releases, $keeps);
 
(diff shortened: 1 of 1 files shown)


antonmedv:
Tests failed.

antonmedv:
Please update changelog too.

mikeymike:
Hmmm I've debugged this and figured out what was happening ... turns out it's my usage of the cleanup that was causing the problem.

My setup separates the code deployment and the symlink into two separate deployments. We do this to deploy using CI but still have manual control over the final release. 

Turns out because Deployer is caching the values from `get` calls the release list does not include the new release, hence why the original code works fine. 

In our case we run the symlink in technically a fresh deployment so that value isn't cached and the new release is included in the release list when the cleanup task runs.

I'll close this as it doesn't really solve anything.

PR #1210, opened 2017-05-15
Let user override phpunit locally.

lucasmichot:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A


antonmedv:
Why do you need it for?

lucasmichot:
A person contributing might tweak phpunit.xml file for different reasons: tweaking the test file, setup coverage or whatever.

When running `vendor/bin/phpunit`, if `phpunit.xml` exists it will be taken into consideration. Or else `phpunit.xml.dist` will be used.

See https://phpunit.de/manual/current/en/organizing-tests.html

antonmedv:
Why not tweak phpunit.xml itself then?

lucasmichot:
Because, if a developer change it, it would be commited

antonmedv:
> Because, if a developer change it, it would be commited

Don't `git add phpunit.xml` :)

lucasmichot:
But does it is not obvious 
Have a look at `.gitignore` on https://github.com/symfony/symfony and https://github.com/laravel/framework

antonmedv:
Okay, lets merge it.

PR #1212, opened 2017-05-15
Remove useless else.

lucasmichot:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

### DIFF
--- a/src/Collection/Collection.php
+++ b/src/Collection/Collection.php
@@ -26,9 +26,9 @@ class Collection implements CollectionInterface, \Countable
     {
         if ($this->has($name)) {
             return $this->values[$name];
-        } else {
-            return $this->throwNotFound($name);
         }
+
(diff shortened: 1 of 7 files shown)


antonmedv:
I think removing else in this cases make  code less readable.

barryvdh:
I think that in this case, switching the expected values would be more readable, when removing the `else`

```php
public function get($name)
{
	if ( ! $this->has($name)) {
		return $this->throwNotFound($name);
	}
	
	return $this->values[$name];
}
```

But I guess that is just personal preference :)

antonmedv:
Yes, personal preference. But if it make sense - ok

lucasmichot:
Feel free to re-open it if you feel so

PR #1215, opened 2017-05-15
throw the correct exception on --reference fail

snowiow:
We delete the .git folder on our live servers after deployment and so the next git clone failed. However the exception wasn't  catched, because it is of type Deployer\Exception\RuntimeException.


antonmedv:
Thanks!

antonmedv:
Can you rebase and resolve conflicts?

snowiow:
yes, merged the upstream and solved the issue

PR #1219, opened 2017-05-18
Write output on verbose, but always start multiplexing

barryvdh:
This only runs the init when verbose, I guess you meant to only write the output on verbose?


barryvdh:
Superseded by #1220

PR #1220, opened 2017-05-18
Check multiplexing until initialized

barryvdh:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    |  No
| Deprecations? | No
| Fixed tickets | #1192

Check the multiplex initialization, until it is working. Prevents async issues when prompting for a passphrase.

PR #1222, opened 2017-05-19
Bump passphrase timeout to 30secs, check if process is running

barryvdh:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #1220

This bumps the wait time for a passphrase to 30 secs. It also checks if the process is running, otherwise throw an immediate exception (for example, after entering the wrong passphrase)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 
 ### Fixed
 - Throw the correct exception on git --reference fail
-- Check if multiplexing is working before continuing
+- Check if multiplexing is working before continuing [#1192]
 - Fixed upload with non-standard SSH port [#1218]
 
 ## v5.0.0
--- a/src/Exception/InitializationException.php
+++ b/src/Exception/InitializationException.php
@@ -1,0 +1,12 @@
+<?php
+/* (c) Anton Medvedev <anton@medv.io>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Deployer\Exception;
+
+class InitializationException extends GracefulShutdownException
+{
+}
--- a/src/Ssh/Client.php
+++ b/src/Ssh/Client.php
@@ -129,12 +129,17 @@ class Client
 
             // Open master connection explicit,
             // ControlMaster=auto could not working
-            (new Process("ssh -M $sshArguments $host"))->start();
+            $process = new Process("ssh -M $sshArguments $host");
+            $process->start();
 
             $attempts = 0;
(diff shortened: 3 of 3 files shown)


antonmedv:
> It also checks if the process is running, 

What for?

barryvdh:
The ssh multiplex command is supposed to keep running in the background, right?
If it fails (eg. connection can't be made, passphrase is not entered correctly) the process exits. So if we keep waiting, it will fail after the timeout (30 secs), instead of directly.

antonmedv:
Master connection detaches, doesn't it?

barryvdh:
Not sure, doesn't seem to in my case. When I run the command manually, it just sets up an SSH connection with exiting. Or is that not intended?
I thought you were using `start` instead of `run` to run it in the background until the deployment finishes.

Should I remove the check then?

antonmedv:
> Should I remove the check then?
 
I'll test it. 

I found two cases: in homestad master stick to process, but on my macos it doesn't.

Can you check in homestad?

barryvdh:
With sticks to process, you mean it keeps the process running? That is the same as in CentOS. I can give it a try on the mac, but if there is not way to detect if it's pending, we should just let it fail. Or perhaps we could still get some output of the failing.

barryvdh:
On my Mac is seems to have the same behavior.

One thing that I'm encountering is that is run the deploy:failed task after. Depending on the hooks (eg. `after(‘deploy:failed’, ‘deploy:unlock’);`), it runs more commands, which in turn ask for passphrases on the failed task. The exception is swallowed untill all tasks are run.

I think the flow should probably be something like this:

 - First initialize SSH (multiplexing etc)
 - If working, continue with the tasks (eg. prepare etc), if then failing, run the failed task
 - If not working, abort before running any commands on the server to avoid multiple passphrase requests / swallowing the init message.

antonmedv:
Nice, i think it will be good solution.

barryvdh:
Is that something you are going to do?

antonmedv:
No

antonmedv:
Maybe you can do it?

barryvdh:
I'll see. Do you still want to merge this first?

barryvdh:
Ah, you already had something like this. I created an InitializationException whichs extends the GracefulShutdownException, which skips the failed task. This shuts down immediatly. Better?

jconroy:
> This bumps the wait time for a passphrase to 30 secs

FWIW this is working well combined with #1220, thanks for the fix @barryvdh (I was hitting the 5sec timeout)

Issue #1227, opened 2017-05-22
Deploy failed on rsync if no port is provided via YAML configuration


dakur:
| Q                 | A
| ----------------- | ---
| Issue Type        | Bug
| Deployer Version  | 5.0.0
| Local Machine OS  | Ubuntu 14.04.5 LTS
| Remote Machine OS | Ubuntu 14.04.5 LTS

### Description 
If I do not provide a `port` directive in YAML configuration (did not try in PHP array config), deploy fails on rsync command with error message `Bad port '-l'`
In the run `ssh` subcommand is `-p` parameter, but no port number provided: `rsync -azP -e 'ssh -p ' ../dist/ someuser@10.0.0.2:/var/www/web/dev/releases/75`

### Steps to reproduce
Config your deploy like this:
```yaml
.deploy-template: &deploy-template
    ssh_type: native
    ssh_multiplexing: true

    user: someuser
#    port: 22 # do not provide this
    identityFile: /root/.ssh/id_rsa


10.0.0.2:
    <<: *deploy-template
    stage: development
    deploy_path: /var/www/web/dev
```

### Content of `deploy.php`
```php
<?php

namespace Deployer;


/** @noinspection PhpIncludeInspection recipe comes from Deployer tool */
require_once 'recipe/common.php';


// deploy stages settings
inventory(__DIR__ . '/config.yml');


// task
task('deploy:copy', function () {
	upload('../dist/', '{{release_path}}');
});


// deploy process
desc('Deploy your project');
task('deploy', [
	'deploy:prepare', // checks env and prepares folder for new release
	'deploy:lock',
	'deploy:release',

	'deploy:copy',

	'deploy:copy_dirs',

	'deploy:symlink', // swap to current release

	'deploy:unlock',
	'cleanup', // remove old releases
	'success',
]);

// automatically unlock if deploy fails
after('deploy:failed', 'deploy:unlock');

```

### Output log
```
➤ Executing task deploy:prepare
[10.0.0.2] > echo $0
[10.0.0.2] < ssh multiplexing initialization
[10.0.0.2] < bash
[10.0.0.2] > if [ ! -d /var/www/web/dev ]; then mkdir -p /var/www/web/dev; fi
[10.0.0.2] > if [ ! -L /var/www/web/dev/current ] && [ -d /var/www/web/dev/current ]; then echo true; fi
[10.0.0.2] > cd /var/www/web/dev && if [ ! -d .dep ]; then mkdir .dep; fi
[...]

PR #1228, opened 2017-05-22
Added default port number into Host (#closes 1227)

dakur:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #1227

### DIFF
--- a/src/Host/Host.php
+++ b/src/Host/Host.php
@@ -18,7 +18,7 @@ class Host
 
     private $hostname;
     private $user;
-    private $port;
+    private $port = 22;
     private $configFile;
     private $identityFile;
     private $forwardAgent = true;


lucasmezencio:
Please! Merge this! :)

antonmedv:
What purpose of this?

lucasmezencio:
@antonmedv because of this error @dakur mentioned in #1227 :

```
[10.0.0.2] > rsync -azP -e 'ssh -p ' ../dist/ someuser@10.0.0.2:/var/www/web/dev/releases/75
[10.0.0.2] < Bad port '-l'
```

The default port for SSH connections is 22, then we should not have to provide this all the time.

antonmedv:
I think is no port provided when `-e 'ssh -p '` should not be added.

dakur:
@antonmedv I think you did not understand it. `-e 'ssh -p '` is what it does now. My fix is to add the port number there.

antonmedv:
What is use don't put port number in host declaration but have different port in ~/.ssh/config? This will break everything. If no port is specified, omit `-e 'ssh -p '`.

lucasmezencio:
@antonmedv sure. I think I got you.

dakur:
@antonmedv I have no port specified in `~/.ssh/config`, I have no `~/.ssh/config` at all. If I omit `port` in host declaration, `-e 'ssh -p '` is called automatically by Deployer. That's why started this issue and made that PR.

antonmedv:
@dakur your solution isn't correct (see my comment above).

> `-e 'ssh -p '` is called automatically by Deployer

This is problem in first place, and the bug what should be fixed.

dakur:
Ok, looks like it was fixed in #1240

PR #1251, opened 2017-05-30
Disable maintenance after failure

barryvdh:
This removes maintenance mode after failure, in case it fails during the DB upgrade.


barryvdh:
Done

antonmedv:
👍

PR #1254, opened 2017-06-05
Update functions.php

nicodmf:
actually, the upload function don't take any options from deploy.php, but documentation mentions option for rsync.
This commit correct that.

| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -423,13 +423,14 @@ function upload($source, $destination, array $config = [])
     $host = Context::get()->getHost();
     $source = parse($source);
     $destination = parse($destination);
+    $config['options'] = is_array($config['options']) ? $config['options'] : [];
 
     if ($host instanceof Localhost) {
         $rsync->call($host->getHostname(), $source, $destination, $config);
     } else {
         $sshArguments = $host->getSshArguments()->getCliArguments();
         if (empty($sshArguments) === false) {
-            $config['options'] = ["-e 'ssh $sshArguments'"];
+            $config['options'][] = "-e 'ssh $sshArguments'";
         }
         $rsync->call($host->getHostname(), $source, "$host:$destination", $config);
     }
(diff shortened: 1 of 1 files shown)


antonmedv:
What is this for?

Exotherme:
@antonmedv 
I just ran into this issue myseld and found this issue. The problem is that if you want to add your own options to rsync, in my case "--inplace", it will not be added because the 'options' config is always overwritten if there are SSH arguments.

src/functions.php
```diff
@@ -423,13 +423,14 @@ function upload($source, $destination, array $config = [])
...
     $source = parse($source);
     $destination = parse($destination);
+    $config['options'] = is_array($config['options']) ? $config['options'] : [];
 
     if ($host instanceof Localhost) {
         $rsync->call($host->getHostname(), $source, $destination, $config);
     } else {
         $sshArguments = $host->getSshArguments()->getCliArguments();
         if (empty($sshArguments) === false) {
-            $config['options'] = ["-e 'ssh $sshArguments'"];
```
Exotherme: Alternatively add the array check here:
if (!is_array($config['options'])) {
    $config['options'] = [];
}

nicodmf: as it, i think options are too for localhost

nicodmf:
actually, the upload function don't take any options from deploy.php, but documentation mentions option for rsync.
This commit correct that.

Exotherme:
@antonmedv & @nicodmf 
Unfortunately this change now causes a PHP notice:
PHP Notice:  Undefined index: options in src/functions.php on line 426
PHP Notice:  Undefined index: options in src/functions.php on line 452

antonmedv:
@nicodmf please, fix

Exotherme:
@antonmedv I've created a fix: #1264

Issue #1235, opened 2017-05-24
Cant use sudo in deployer


hobsRKM:
The command "cd /home/ec2-user/code_deploys && (rm -rf "$(readlink release)")" failed.
rm is failing bcoz of absence of **sudo**.
I have set **set('release_use_sudo', true);**
Didn't work.

antonmedv:
There is no sudo in this command.
Maybe add `cleanup_use_sudo` there too.

PR #1260, opened 2017-06-10
fix can't use sudo in deployer

shubhshrma:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes 
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #1235


shubhshrma:
was this not the solution or is there something else

antonmedv:
The default is false

Issue #484, opened 2015-11-22
Сustom php binary path


dmkuznetsov:
On some virtual hosting platforms for correct deploy we need to specify php binary.

For instance:

```bash
/opt/php54/bin/php composer.phar install
```

I suggest to create new environment var, for example "php_bin" with default value "php"

elfet:
Why not using `alias php='/opt/php54/bin/php'`?

dmkuznetsov:
Just "php" is busy with old version of php. 
I've tried to create .bashrc with alias, but It's not working for me.

lbausch:
`.bashrc` is only taken into account for interactive shells.  You may check for `.profile` instead.

elfet:
Then try `env('php', '/path/php')` and then use: `run('{{ php }} ...')`.

dmkuznetsov:
@lbausch thanks! It's work!

@elfet I resolved my issue by overriding tasks ```deploy:vendors```, ```deploy:assetic:dump```, etc.

In generally yes, users can create alias and put in ```.profile```, but True-way is put this variable in configuration

wesleywesley:
.bashrc does not work, .profile does not work, env('php', .. ) does not work. Exactly as @dmkuznetsov said, we need an environment variable to specify PHP version. Or a way to set the PATH variable and use it in all subsequent calls (better!)

elfet:
Will be in Deployer v4. #586

whyte624:
There is {{bin/php}} env variable now.

PR #1261, opened 2017-06-12
Mitigate missing `which` command on some systems

Addvilz:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A or xx

Some systems, especially thin Docker images, do not have `which` command, used to locate path of `php`, `git`, `composer` etc.

This pull request adds method which should mitigate this on most systems, prioritizing `command` over `which` with fallback to `type`. This should cover most systems out there.

Related to #484, #607, #586

See [this SXC article](https://unix.stackexchange.com/questions/85249/why-not-use-which-what-to-use-then) for more related info.

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -736,3 +736,30 @@ function parse($value)
 {
     return Context::get()->getConfig()->parse($value);
 }
+
+function locateBinaryPath($name)
+{
+    $nameEscaped = escapeshellarg($name);
+
+    // Try `command`, should cover all Bourne-like shells
+    if (commandExist("command")) {
+        return run("command -v $nameEscaped")->toString();
+    }
+
+    // Try `which`, should cover most other cases
+    if (commandExist("which")) {
+        return run("which $nameEscaped")->toString();
+    }
+
+    // Fallback to `type` command, if the rest fails
+    if (commandExist("type")) {
+        $result = run("type -p $nameEscaped")->toString();
+
+        if ($result) {
+            // Deal with issue when `type -p` outputs something like `type -ap` in some implementations
+            return trim(str_replace("$name is", "", $result));
+        }
+    }
+
+    throw new \RuntimeException("Can't locate [$nameEscaped] - neither of [command|which|type] commands are available");
+}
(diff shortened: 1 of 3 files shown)


src/functions.php
```diff
@@ -736,3 +736,30 @@ function parse($value)
...
+{
+    $nameEscaped = escapeshellarg($name);
+
+    // Try `command`, should cover all Bourne-like shells
+    if (commandExist("command")) {
+        return run("command -v $nameEscaped")->toString();
+    }
+
+    // Try `which`, should cover most other cases
+    if (commandExist("which")) {
```
antonmedv: Let's leave in first place.

antonmedv:
Nice feature!

src/functions.php
Addvilz: Can I ask why? `command -v` is available for `zsh`, `bash`, `sh` and `fish` by default, which covers probably 99. percentile of systems out there. It is also a built-in for most implementations I know of, instead of external binary.

antonmedv: I was thinking about keeping as much possible to previous implementation, but in command -v is common then it's okay.

Addvilz: For the sake of absolute safety just tested this on Debian Wheezy, Fedora24/25, Ubuntu Xenial/Trusty, Centos7 and latest Alpine in docker, and even Windows Subsystem for Linux (don't ask). command -v is supported by all. Centos 7, F24, F25 does not have which installed by default. Debian derivs seem all to have which, but also command, and command being builtin is probably the saner choice for default.

antonmedv: Nice work! I'm impressed.

antonmedv: Btw, how do you use docker with deployer?

Addvilz: Sorry, what do you mean?

antonmedv: How do you use docker? :)

Addvilz: Again, sorry, but how do we use docker for what? Deployments? We have couple of projects using deployer and as well dockerized. CI is then using Deployer to deploy to environments at various stages of the project from within docker, using a build file in each project. Deployer then is run from within docker, as is pretty much every other dependency of the project, including phpunit, etc.

antonmedv: Okay, thanks. I'm interested in the way people use docker.

PR #1264, opened 2017-06-15
Upload and download undefined index notice

Exotherme:
Relates to issue #1254 
Fixes a PHP undefined index notice when no options are supplied in the $config variable.

| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #1254

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -423,13 +423,15 @@ function upload($source, $destination, array $config = [])
     $host = Context::get()->getHost();
     $source = parse($source);
     $destination = parse($destination);
-    $config['options'] = is_array($config['options']) ? $config['options'] : [];
 
     if ($host instanceof Localhost) {
         $rsync->call($host->getHostname(), $source, $destination, $config);
     } else {
         $sshArguments = $host->getSshArguments()->getCliArguments();
         if (empty($sshArguments) === false) {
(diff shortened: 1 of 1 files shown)


src/functions.php
```diff
@@ -423,13 +423,15 @@ function upload($source, $destination, array $config = [])
...
     $source = parse($source);
     $destination = parse($destination);
-    $config['options'] = is_array($config['options']) ? $config['options'] : [];
 
     if ($host instanceof Localhost) {
         $rsync->call($host->getHostname(), $source, $destination, $config);
     } else {
         $sshArguments = $host->getSshArguments()->getCliArguments();
         if (empty($sshArguments) === false) {
+            if (!isset($config['options']) || !is_array($config['options'])) {
```
antonmedv: Why `is_array`?

Exotherme: It's the same as the original is_array check (line 426), but with an isset before to prevent the notice.

Consider options would contain a string (or another non-array value):
$config['options'] = 'some string';

$config['options][] = "-e 'ssh $sshArguments'";
would cause:
PHP Fatal error:  Uncaught Error: [] operator not supported for strings

PR #1269, opened 2017-06-19
Per user persistent storage

ccristescu-reol:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

In a multi user environment where multiple projects are located on the same instance(s)
if two or more users are deploying a project on the same environment, the generated hosts files
in the system tmp folder are overwritten or worse, the deploy will fail due to file permissions.

Make the persistent storage folder configurable or use the system tmp  as default if not possible.

### DIFF
--- a/src/Host/Storage.php
+++ b/src/Host/Storage.php
@@ -31,7 +31,7 @@ class Storage
 		// default persistent storage in case we can't use the configured value
 		// or we can't create the default location
 		// use the system temporary folder and the current pid to make the path unique
-		$tmp = sys_get_temp_dir() . '/' . uuid(posix_getpid());
+		$tmp = sys_get_temp_dir() . '/' . uniqid(posix_getpid());
 		
 		// get the location for the deployer configuration
 		$deployerConfig = $config->has('deployer_config') ? $config->get('deployer_config') : null;
--- a/src/Host/Storage.php
+++ b/src/Host/Storage.php
@@ -14,6 +14,72 @@ use function Deployer\Support\array_flatten;
 
 class Storage
 {
+	/**
+	 * In a multi user environment where multiple projects are located on the same instance(s)
+	 *	if two or more users are deploying a project on the same environment, the generated hosts files
+	 *	are overwritten or worse, the deploy will fail due to file permissions
+	 *
+	 * Make the persistent storage folder configurable or use the system tmp 
+	 *	as default if not possible
+	 * 
+	 * @return string
+	 * @throws Exception
+	 */
+	private static function _getPersistentStorageLocation() {
+		$config = \Deployer\Deployer::get()->config;
+
+		// default persistent storage in case we can't use the configured value
+		// or we can't create the default location
+		// use the system temporary folder and the current pid to make the path unique
+		$tmp = sys_get_temp_dir() . '/' . uuid(posix_getpid());
+		
+		// get the location for the deployer configuration
+		$deployerConfig = $config->has('deployer_config') ? $config->get('deployer_config') : null;
+
+		if ( !is_null($deployerConfig) ) {
+			// check if the folder exists and it's writable
+			if ( !(is_dir($deployerConfig) && is_writable($deployerConfig)) ) {
+				throw new Exception("Deployer folder `$deployerConfig` doesn't exists or doesn't writable.");
+			}
+		} else {
+			// not configured, generate deployer folder name
+
+			// use the home dir of the current user 
+			// and the repository name
+			$userInfo = posix_getpwuid(posix_getuid());
+			
+			// if for some reason we couldn't find a valid home folder
+			// or if it's not writable, use the default location
+			if ( !isset($userInfo['dir']) || !(is_dir($userInfo['dir']) && is_writable($userInfo['dir'])) ) {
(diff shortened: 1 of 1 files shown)


src/Host/Storage.php
```diff
@@ -14,6 +14,73 @@
...
+	 * @return string
+	 * @throws Exception
+	 */
+	private static function _getPersistentStorageLocation() {
+		$config = \Deployer\Deployer::get()->config;
+
+		// use the system temporary folder and the current pid as default 
+		// persistent storage in case we can't use the configured value
+		// or we can't create the default location
+		$tmp = sys_get_temp_dir() . '/' . posix_getpid();
```
antonmedv: This extension is not available on Windows platforms.
http://php.net/manual/en/intro.posix.php

src/Host/Storage.php
```diff
@@ -14,6 +14,73 @@
...
+	 *	the generated hosts files are overwritten or worse, the deploy will fail 
+	 *	due to file permissions
+	 *
+	 * Make the persistent storage folder configurable or use the system tmp 
+	 *	as default if not possible
+	 * 
+	 * @return string
+	 * @throws Exception
+	 */
+	private static function _getPersistentStorageLocation() {
```
antonmedv: Rename `_getPersistentStorageLocation` to `getPersistentStorageLocation` and move this method to end of class.

src/Host/Storage.php
```diff
@@ -14,6 +14,73 @@
...
+		$config = \Deployer\Deployer::get()->config;
+
+		// use the system temporary folder and the current pid as default 
+		// persistent storage in case we can't use the configured value
+		// or we can't create the default location
+		$tmp = sys_get_temp_dir() . '/' . posix_getpid();
+
+		// get the location for the deployer configuration
+		$deployerConfig = $config->has('deployer_config') ? $config->get('deployer_config') : null;
```
antonmedv: It's over this PR, I think such functionality should done in separate PR.

Anyway I don't see way it's needed. Please remove everything related to deployer_config.

src/Host/Storage.php
```diff
@@ -14,6 +14,73 @@
...
+			// check if the folder exists and it's writable
+			if ( !(is_dir($deployerConfig) && is_writable($deployerConfig)) ) {
+				throw new Exception("Deployer folder `$deployerConfig` doesn't exists or doesn't writable.");
+			}
+		} else {
+			// not configured, generate deployer folder name
+
+			// use the home dir of the current user 
+			// and the repository name
+			$userInfo = posix_getpwuid(posix_getuid());
```
antonmedv: Same here, what about windows?

src/Host/Storage.php
```diff
@@ -14,6 +14,73 @@
...
+		}
+		
+		// we will store the persistent data per repository
+		$configRepository = $config->has('repository') ? $config->get('repository') : null;
+		if (empty($configRepository)) {
+			return $tmp;
+		} 
+		
+		// we now have the repository name
+		$repository = str_replace('/', '_', substr($configRepository, (strrpos($configRepository, ':') + 1)));
```
antonmedv: What is it for?

src/Host/Storage.php
ccristescu-reol: For non posix compatible i'm going to use getenv.
getenv('HOMEDRIVE') . getenv('HOMEPATH') are available on windows 8+ 

It's this enough?

src/Host/Storage.php
ccristescu-reol: Long story short: i have five webapps on the same instances (my staging). If i deploy two apps at the same time (using two terminals for example) the second one i launch will overwrite the content of the .dep files so, the first deployer will start running tasks in the locations set by the second one and everything ends up in a mess.

The .dep files should be stored per webapp so, the easiest way was to use the repository to generate a separate folder name.

src/Host/Storage.php
ccristescu-reol: For non posix compatible i'm going to use getenv.
getenv('HOMEDRIVE'), getenv('HOMEPATH') are available on windows 8+ 
getenv('HOME') for *nix

It's Windows 8+ enough?

ccristescu-reol:
Windows 8+ and non posix systems compatibility, added alternatives to the posix_* calls:
- getmypid() for posix_getpid
- getenv()for posix_getpwuid

Method _getPersistentStorageLocation renamed to getPersistentStorageLocation and moved at then end.

Dropped the option to have a configurable path for the persistent storage folder.

src/Host/Storage.php
antonmedv: I think it's easier to create rundom files instead.

ccristescu-reol: Easier yes... but then you will need an internal task to clean up temporary files older than N days or something simmilar.
This way it's cleaner... i would say :)

antonmedv: But tmp cleaned by system itself

ccristescu-reol:
Removed everything and replaced it with randomly generate file names in system tmp.

PR #1286, opened 2017-07-03
Simpler way of handling optional parameters

taka-oyama:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

I think this is a better way of doing it


taka-oyama:
The use of equal sign for optional parameter was not needed.
This is a better solution to the problem I think.

Please check and let me know what you think.

antonmedv:
Can you add a few test for it? Will be cool to see if we wont break anything in future.

taka-oyama:
Will do.

PR #1300, opened 2017-07-14
Add TYPO3 project type to dep init

stucki:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

PR #1301, opened 2017-07-15
[TASK] Add TYPO3 recipe template to the init command

NeoBlack:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

### DIFF
--- a/src/Initializer/Template/TYPO3Template.php
+++ b/src/Initializer/Template/TYPO3Template.php
@@ -1,0 +1,56 @@
+<?php
+/* (c) Anton Medvedev <anton@medv.io>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Deployer\Initializer\Template;
+
+/**
+ * Generate a common (base) deployer configuration
+ *
+ * @author Vitaliy Zhuk <zhuk2205@gmail.com>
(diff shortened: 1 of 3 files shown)


src/Initializer/Template/TYPO3Template.php
```diff
@@ -0,0 +1,56 @@
...
+class TYPO3Template extends Template
+{
+    /**
+     * Get content of template.
+     *
+     * @param array $params
+     *
+     * @return string
+     */
+    protected function getTemplateContent($params)
```
antonmedv: Please use this instead:
```
    protected function getRecipe()
    {
        return 'typo3';
    }
```

NeoBlack: can you explain why?
My idea is to provide a deploy.php file with information about how to use the recipe.

antonmedv: All framework recipes uses it. So it can be more easily maintained. I don't see any notable changes in your template.

antonmedv:
Merged #1300 instead.

src/Initializer/Template/TYPO3Template.php
NeoBlack: ok, thx

PR #1303, opened 2017-07-17
Remove useless else

lucasmichot:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

---

Remove useless else

### DIFF
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -1,8 +1,11 @@
 preset: psr2
 
 enabled:
-  - short_array_syntax
+  - alpha_ordered_imports
+  - no_unused_imports
+  - no_useless_else
   - no_whitespace_in_blank_line
+  - short_array_syntax
 
 finder:
   not-name:


antonmedv:
?)

lucasmichot:
@antonmedv - yes StyleCI does not take in consideration my change
I believe this is because you've define the fixers in the StyleCI dashboards: thus this file is never taken into consideration

antonmedv:
Nope, it's emply

lucasmichot:
Then StyleCI is not working properly then....

lucasmichot:
See by yourself, the code is full of useless else, but StyleCI does not catch them....

antonmedv:
I see only one changes file in this PR 
image

lucasmichot:
Indeed this is the problem: I've modified StyleCI so that it checks the useless else, but StyleCI could not find anything

PR #1323, opened 2017-08-10
Introduce git_recursive option

rskuipers:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

This PR introduces an option to enable or disable the `--recursive` flag while cloning. The use case is that we only have submodules for development related stuff, so dropping the `--recursive` flag would be more performant and it would keep us from having to include deploy keys for those git repositories as well.

```php
set('git_recursive', true);
```

```bash
➤ Executing task deploy:update_code
...
[example.com] > /usr/bin/git clone  --depth 1 --recursive -q git@github.com:deployphp/deployer.git /srv/http/foobar/releases/9 2>&1
```

```php
set('git_recursive', false);
```

```bash
➤ Executing task deploy:update_code
...
[example.com] > /usr/bin/git clone  --depth 1  -q git@github.com:deployphp/deployer.git /srv/http/foobar/releases/8 2>&1
```

I didn't see any tests I could extend with this new option, let me know if I missed something.

I've also created a PR to update the documentation: https://github.com/deployphp/docs/pull/102

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 ### Added
 - Added possibility to use callable when setting 'default_stage'.
 - Added console init template for TYPO3 CMS [#1300]
+- Added `git_recursive` option
 
 ## v5.1.3
 [v5.1.2...v5.1.3](https://github.com/deployphp/deployer/compare/v5.1.2...v5.1.3)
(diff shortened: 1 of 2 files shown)


antonmedv:
Nice!

Issue #1351, opened 2017-08-31
--version option is not working properly


matyx:
| Q                 | A
| ----------------- | ---
| Issue Type        | Bug
| Deployer Version  | ^4, ^5 and ^6 (composer notation)
| Local Machine OS  | Linux
| Remote Machine OS | N/A

### Description 
`dep --version` always returns master only, and it should returns version of installed deployer

### Steps to reproduce
```bash
composer require deployer/deployer:^5
vendor/bin/dep --version #outputs Deployer master, should be v5.1.3

composer require deployer/deployer:^6
vendor/bin/dep --version #outputs Deployer master, be v6.0.2

composer require deployer/deployer:^4
vendor/bin/dep --version #outputs Deployer master, should be v4.3.1
```

antonmedv:
Actually it should. Take a look at https://github.com/deployphp/deployer/blob/master/bin/dep#L91
Maybe a bug there.

matyx:
I think that it would be better to hardcode version into deployer files, because now I am not able to check version from source, and deployer could be used without composer. I can send PRs, if you agree with this.

antonmedv:
I don't think what hardcoded version will be good. Maintaining versions will be harder as we need to create separate commits with version for releases in order to save master varsion for master branch.

chellem:
The problem is that when installing deployer globally, the version is found in ~/composer.lock but `dep` is trying the read composer.lock from inside the deployer vendor directory which does not contains deployer/deployer version as it is found.

PR #1354, opened 2017-09-01
Fix  --version option is not working properly #1351

chellem:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes 
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | 1351

### DIFF
--- a/bin/dep
+++ b/bin/dep
@@ -86,18 +86,25 @@ set_include_path($includePath . PATH_SEPARATOR . get_include_path());
 
 // Detect version
 $version = 'master';
-
-$composerLockFile = $deployFilePath . '/composer.lock';
-if (is_readable($composerLockFile)) {
-    $lock = json_decode(file_get_contents($composerLockFile), true);
-    foreach ($lock['packages'] as $package) {
-        if ($package['name'] === 'deployer/deployer') {
-            $version = $package['version'];
-        }
+$composerLockFiles = array(
+        realpath($deployFilePath . '/../composer.lock'),
+        realpath(__DIR__ . '/../../../../composer.lock')
+);
+foreach($composerLockFiles as $composerLockFile){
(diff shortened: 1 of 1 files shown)


antonmedv:
Please add comments about implementation,

bin/dep
```diff
@@ -86,18 +86,25 @@ set_include_path($includePath . PATH_SEPARATOR . get_include_path());
...
-
-$composerLockFile = $deployFilePath . '/composer.lock';
-if (is_readable($composerLockFile)) {
-    $lock = json_decode(file_get_contents($composerLockFile), true);
-    foreach ($lock['packages'] as $package) {
-        if ($package['name'] === 'deployer/deployer') {
-            $version = $package['version'];
-        }
+$composerLockFiles = array(
+        realpath($deployFilePath . '/../composer.lock'),
```
antonmedv: Why ..?

bin/dep
```diff
@@ -86,18 +86,25 @@ set_include_path($includePath . PATH_SEPARATOR . get_include_path());
...
+        realpath(__DIR__ . '/../../../../composer.lock')
```
antonmedv: ../../../..?

chellem:
@antonmedv the changes doesn't work?

bin/dep
chellem: getting the global composer.lock

bin/dep
chellem: because I don't find composer.lock in the same directory as $deployFilePath

antonmedv:
But it does working now: 
```
php vendor/bin/dep --version
```

antonmedv:
After a lot fo thinking, I decided to go away from using version detection. Now version will be set manually every time before release.

PR #1355, opened 2017-09-01
Controlled recursion

ps-mattstuart:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

This introduces a new configuration option:

```
/**
 * Accepts TRUE, FALSE, ['module1', 'module2',..], or 'module1,module2,..'
 * Note: this will override git_recursive in some instances
 * TRUE - default behaviour, include all modules if git_recursive is TRUE
 * FALSEy - Include no modules, will override git_recursive (FALSE, [], '')
 * [a, b, c] - Include only these modules, overrides git_recursive
 * 'a,b,c' - Include only these modules, overrides git_recursive
 */
set('git_submodules', true);
```

The purpose is to allow control of the modules to be included during a deploy. 

For example, unit tests are a submodule in my repository and I only want them included in a CI build but not during production builds to keep them lighter. 

This also adds support for updating submodules on revision checkout. Includes what I think are an acceptable (though just barely) number of tests.

Next iterations:
* Support wildcards in the module name to allow for better recursive control over submodule deployment. Note: newer versions of git include options to support recursive submodule init

Feedback welcome!

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -64,25 +64,29 @@
         },
         {
             "name": "pimple/pimple",
-            "version": "v3.0.2",
+            "version": "v3.2.2",
             "source": {
                 "type": "git",
                 "url": "https://github.com/silexphp/Pimple.git",
-                "reference": "a30f7d6e57565a2e1a316e1baf2a483f788b258a"
+                "reference": "4d45fb62d96418396ec58ba76e6f065bca16e10a"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/silexphp/Pimple/zipball/a30f7d6e57565a2e1a316e1baf2a483f788b258a",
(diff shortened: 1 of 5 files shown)


ps-mattstuart:
TravisCI `--prefer-lowest` is going to fail on PHP7.0 w/ the Trusty image until they update it or we specify a specific version of PHP. It's currently using PHP7.0.7 & PHPUNIT6.0.0 which has a documented issue. The other tests pass fine.

antonmedv:
Sorry for delay. 

Nice work! But I think we shouldn't put in common recipe . It will complicate it a lot and confuse newcomers. Better is to create a separate recipe this this implementation.

PR #1363, opened 2017-09-15
Composer optimization level 2

piotrantosik:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

Use classmap-authoritative as default composer optimization.
For details see - https://getcomposer.org/doc/articles/autoloader-optimization.md#optimization-level-2-a-authoritative-class-maps

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -82,7 +82,7 @@ set('use_atomic_symlink', function () {
 });
 
 set('composer_action', 'install');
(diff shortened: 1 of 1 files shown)


barryvdh:
> On the flipside it also means that in case a class is generated at runtime for some reason, it will not be allowed to be autoloaded. If your project or any of your dependencies does that then you might experience "class not found" issues in production. Enable this with care.

That doesn't seem like a good default. I suspect that, for example, Laravel facades fall in that category.

PR #1365, opened 2017-09-18
Run autoloader after compilation

barryvdh:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | No
| BC breaks?    | No?
| Deprecations? | No
| Fixed tickets | N/A

According to the Magento opmization guide, the dump-autoload command should be called after the compiling of classes:
http://devdocs.magento.com/guides/v2.1/config-guide/prod/prod_perf-optimize.html

> After running setup:di:compile to generate classes, use composer to update the autoloader.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## master
 [v6.0.3...master](https://github.com/deployphp/deployer/compare/v6.0.3...master)
 
+### Changed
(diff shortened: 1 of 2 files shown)


barryvdh:
No, but it's recommended as performance improvement: http://devdocs.magento.com/guides/v2.1/config-guide/prod/prod_perf-optimize.html#server---php-configuration

PR #1389, opened 2017-10-18
Check if permissions are set if using sudo

rhwilr:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A 

When using sudo, setting the permissions for writable directories should work the same as without sudo.
I now check if the acl is already set and abort if so.

Background: Setting ACLs took over 5min on our production server due to the large amount of small files inside subfolders. With this change we now are back to a comfortable 5 seconds.


antonmedv:
Sorry for delay. Tests are fallen. Let me investigate your solution. Can you explain a little bit more?

rhwilr:
I saw that the tests failed, but from the travis log there seems to be en error with installing the dependencies. Maybe you can rerun them.

PR #1392, opened 2017-10-20
[New recipe] Copy files

fabkow:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | 1317


antonmedv:
Okay, now we endup with `copy_dirs` and `copy_files`. :) 
Let's think about how can we improve this situation. Can we leave only one option? What about BC? Don't use rsync?

PR #1431, opened 2017-11-27
permission denied for the git working direcotry

Erkens:
When using the "become" option in a host configuration, the work directory for git needs to be accesible for that user.
For example, the homedirectory from the user I use to SSH isn't readable for the user I use to deploy, so it fails with a "permission denied".
(also make sure to set the HOME enviroment variable so the correct git configuation can be read if needed)

| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

### DIFF
--- a/recipe/deploy/update_code.php
+++ b/recipe/deploy/update_code.php
@@ -76,7 +76,8 @@ task('deploy:update_code', function () {
             $depth = '';
         }
     }
-
+    
+    cd('{{deploy_path}}');
     if ($gitCache && has('previous_release')) {
         try {
             run("$git clone $at $recursive -q --reference {{previous_release}} --dissociate $repository  {{release_path}} 2>&1", $options);


antonmedv:
Can you explain a little bit more?

Erkens:
On my server I have an user "deploy" which is configured with the "become" option:
`
host("somehost")
    ->become("deploy");
`

I connect to the server with my own credentials (user: erkens, with ssh key) and my user is allowed to "sudo -u deploy". But the "deploy" user is not allowed inside my home-directory.

The git client can't access the current workdirectory (/home/erkens) and exits with the "Permission denied" result. When you first change the current workdirectory to a place where it can read (like the deploy_path) git will work fine. (it can however give a warning because of the "HOME" enviroment variable, which is still set to "/home/erkens", but this can be fixed by:

`
host("somehost")
    ->become("deploy")
    ->set('env', ['HOME' => '~deploy']);
`

PR #1432, opened 2017-11-27
Allow for naming the symbolic link

Repox:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

This is a feature I needed which made me think I'm not the only one.
It's backwards compatible and allows for changing the symbolic link name (`current`) to a user defined value, like this:

    set('symlink_name', 'latest');

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 [v6.0.3...master](https://github.com/deployphp/deployer/compare/v6.0.3...master)
 
 ### Changed
+- Added support for naming the symbolic link
 - Added support for GroupTask in invoke() [#1364]
 - Magento2 recipe optimizes the autoloader after the DI compilation [#1365]
 - Host's `roles()` API now can accept arrays too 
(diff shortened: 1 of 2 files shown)


dkorpar:
`recipe/common.php` already has a variable defined and most of the recipes are built around that, but changing it won't do much since the path is hardcoded in several places.

```
set('current_path', function () {
    $link = run("readlink {{deploy_path}}/current");
    return substr($link, 0, 1) === '/' ? $link : get('deploy_path') . '/' . $link;
});
```

`recipe/deploy/prepare.php` and `recipe/deploy/rollback.php` are one of them that should be taken care of. There may be more that I haven't run into.

cafferata:
>There may be more that I haven't run into.

That is indeed true, see the links below

- https://github.com/deployphp/deployer/blob/master/recipe/laravel.php#L53
- https://github.com/deployphp/deployer/blob/master/recipe/laravel.php#L59
- https://github.com/deployphp/deployer/blob/master/recipe/magento2.php#L45

PR #1434, opened 2017-11-30
Add a variable for "realhostname"

syphernl:
In a setup like this:
```
host('production')
    ->hostname('server.example.com)
    ->user('example')
    ->forwardAgent(true)
    ->set('environment', 'production')
    ->set('deploy_path', '/var/www');
```

the `hostname`  variable is actually the stage name (`production`) rather than the hostname (`server.example.com`). 

This PR adds a `realhostname` variable that can be used to refer to .. well, the real hostname :smile:

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -35,6 +35,10 @@ set('hostname', function () {
     return Context::get()->getHost()->getHostname();
 });
 
+set('realhostname', function () {
+    return Context::get()->getHost()->getRealHostname();
+});
+
 set('user', function () {
     try {
         return runLocally('git config --get user.name');


antonmedv:
What about renaming `hostname` -> `$host->getRealHostname()` and `hostalias` -> `$host->getHostname()`?

For next major v7?

syphernl:
That is probably even a better idea since right now the content doesn't match what it is set to.

staabm:
IMO deployer way too often releases major version after major version (becaus eof BC breaking changes).

This makes the lib unattractive for users/consumers, because most of the time you alos stop supporting older versions.

Just my thoughts so.

antonmedv:
I don't like BC too. But it good to have new major with BC. Also upgrade.md is updated. 

I think to do same other do - introduce LTS version. And I think first LTS version will be next v7.

PS A lot of new stuff what I wanted to do for a log time will ship with v7.

Issue #1437, opened 2017-11-30
Add support for Symfony 4.0


tehplague:
| Q                 | A
| ----------------- | ---
| Issue Type        | Feature Request
| Deployer Version  | 6.0
| Local Machine OS  | N/A
| Remote Machine OS | N/A

### Description 
Now that Symfony 4.0 is released, let's try to make Deployer compatible with its Console component.

Cryde:
We could already do some research like : 

```
// Clear paths
set('clear_paths', ['web/app_*.php', 'web/config.php']); // <= no needed anymore

// Symfony shared files
set('shared_files', ['app/config/parameters.yml']); // <= no needed anymore / .env now !

// Assets
set('assets', ['web/css', 'web/images', 'web/js']); // <= old
set('assets', ['public/css', 'public/images', 'public/js']); // <= new 
```

For me the https://github.com/deployphp/deployer/blob/master/recipe/symfony3.php is ok ! :)
Some stuff to adjust about symfony env variable I guess in : https://github.com/deployphp/deployer/blob/master/recipe/symfony.php

antonmedv:
#1420

zorn-v:
@Cryde 
`deploy:assets:install` task not compatible with flex, as here https://github.com/deployphp/deployer/blob/6708d2640f9d8b68b539b3929d9b09cf0d4fec1e/recipe/symfony.php#L95 `web` is hardcoded

erik-ropez:
Larave 5.6 is out and it depends on Symfony 4.

solofeed:
We have the problem with Symfony 4 support.

antonmedv:
Done! 🎉

Shaked:
@antonmedv is this really done? 

AFAIK symfony now uses `public/` instead of `web/` and `app/console` is `bin/console`.

zorn-v:
Use `require/symfony4.php`

PR #1440, opened 2017-12-08
Symfony flex support

zorn-v:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #1437

### DIFF
--- a/recipe/symfony-flex.php
+++ b/recipe/symfony-flex.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Deployer;
+
+require_once __DIR__ . '/symfony3.php';
+
+/**
+ * Symfony Flex Configuration
+ */
+
+//No need to clear anything
+set('clear_paths', []);
+
+// Symfony shared dirs
+set('shared_dirs', ['var/log', 'var/sessions']);
+
+// Symfony writable dirs
+set('writable_dirs', ['var/cache', 'var/log', 'var/sessions']);
+
+//File for DotEnv if not using env vars
+set('shared_files', ['.env']);
+
+// Symfony web dir
+set('web_dir', 'public');
+
+// Assets
+set('assets', ['public/css', 'public/images', 'public/js']);
+
+// Environment vars
+set('env', function () {
+    return [
+        'APP_ENV' => get('symfony_env')
+    ];
+});
--- a/recipe/symfony-flex.php
+++ b/recipe/symfony-flex.php
@@ -26,9 +26,5 @@ set('web_dir', 'public');
 // Assets
 set('assets', ['public/css', 'public/images', 'public/js']);
 
-// Environment vars
-set('env', function () {
-    return [
-        'APP_ENV' => get('symfony_env')
-    ];
-});
+// No need environment vars if using DotEnv
+set('env', []);
--- a/src/Initializer/Template/SymfonyFlexTemplate.php
+++ b/src/Initializer/Template/SymfonyFlexTemplate.php
@@ -13,32 +13,10 @@ namespace Deployer\Initializer\Template;
  * @author zorn-v
  * @codeCoverageIgnore
  */
-class SymfonyFlexTemplate extends FrameworkTemplate
+class SymfonyFlexTemplate extends SymfonyTemplate
 {
     protected function getRecipe()
     {
         return 'symfony-flex';
-    }
-
-    /**
-     * {@inheritDoc}
-     */
-    protected function getExtraContent()
-    {
-        return <<<PHP
-
-// Migrate database before symlink new release.
-
-before('deploy:symlink', 'database:migrate');
-
-//If you don't use DotEnv in prod just wipe it out
-task('clear:env', function () {
-    //DotEnt does not execute if APP_ENV is set, but it required for `composer install`
-    \$env = get('env');
-    unset(\$env['APP_ENV']);
-    set('env', \$env);
-})->setPrivate();
-after('deploy:vendors', 'clear:env');
-PHP;
     }
 }
--- a/src/Initializer/Template/SymfonyFlexTemplate.php
+++ b/src/Initializer/Template/SymfonyFlexTemplate.php
@@ -1,0 +1,44 @@
+<?php
+/* (c) github.com/zorn-v
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Deployer\Initializer\Template;
+
+/**
+ * Generate a Symfony flex deployer configuration.
+ *
+ * @author zorn-v
+ * @codeCoverageIgnore
+ */
+class SymfonyFlexTemplate extends FrameworkTemplate
+{
+    protected function getRecipe()
+    {
+        return 'symfony-flex';
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    protected function getExtraContent()
+    {
+        return <<<PHP
+
+// Migrate database before symlink new release.
+
+before('deploy:symlink', 'database:migrate');
+
+//If you don't use DotEnv in prod just wipe it out
+task('clear:env', function () {
+    //DotEnt does not execute if APP_ENV is set, but it required for `composer install`
+    \$env = get('env');
+    unset(\$env['APP_ENV']);
+    set('env', \$env);
+})->setPrivate();
+after('deploy:vendors', 'clear:env');
+PHP;
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,10 @@
 [v6.0.3...master](https://github.com/deployphp/deployer/compare/v6.0.3...master)
 
 ### Changed
+- Added recipe for Symfony flex
 - Added support for GroupTask in invoke() [#1364]
 - Magento2 recipe optimizes the autoloader after the DI compilation [#1365]
-- Host's `roles()` API now can accept arrays too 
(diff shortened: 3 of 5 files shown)


recipe/symfony-flex.php
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Symfony Flex Configuration
+ */
+
+//No need to clear anything
+set('clear_paths', []);
+
+// Symfony shared dirs
+set('shared_dirs', ['var/log', 'var/sessions']);
```
Cryde: same here

recipe/symfony-flex.php
```diff
@@ -0,0 +1,34 @@
...
+
+// Symfony writable dirs
+set('writable_dirs', ['var/cache', 'var/log', 'var/sessions']);
```
Cryde: same here

recipe/symfony-flex.php
```diff
@@ -0,0 +1,34 @@
...
+// Symfony web dir
+set('web_dir', 'public');
+
+// Assets
+set('assets', ['public/css', 'public/images', 'public/js']);
+
+// Environment vars
+set('env', function () {
+    return [
+        'APP_ENV' => get('symfony_env')
```
Cryde: We should probably add documentation to tell the user to put "symfony_env" in it's host.yml

recipe/symfony-flex.php
```diff
@@ -0,0 +1,34 @@
...
+namespace Deployer;
+
+require_once __DIR__ . '/symfony3.php';
+
+/**
+ * Symfony Flex Configuration
+ */
+
+//No need to clear anything
+set('clear_paths', []);
```
Cryde: It's already define in symfony3 recipe :)

recipe/symfony-flex.php
```diff
@@ -0,0 +1,34 @@
...
+
+// Symfony shared dirs
+set('shared_dirs', ['var/log', 'var/sessions']);
+
+// Symfony writable dirs
+set('writable_dirs', ['var/cache', 'var/log', 'var/sessions']);
+
+//File for DotEnv if not using env vars
+set('shared_files', ['.env']);
```
Cryde: And btw the dotenv component is in require-dev when we do a fresh install ... I'm not sure if it's a good idea

Cryde:
Another tought : what about commands launched that require DB access (in prod mode) ?  🤔    
Like migration for instance

recipe/symfony-flex.php
antonmedv: Actually there is a conversation on moving dotenv to require https://github.com/symfony/symfony/issues/25643

recipe/symfony-flex.php
zorn-v: Nope

recipe/symfony-flex.php
zorn-v: Also nope. And 'var/logs', not 'var/log'

recipe/symfony-flex.php
zorn-v: Also `var/log`

recipe/symfony-flex.php
zorn-v: Before it was `SYMFONY_ENV` https://github.com/deployphp/deployer/blob/master/recipe/symfony.php#L41
But it not for flex

zorn-v:
And now think about is it ENV vars really good for deploy ?

antonmedv:
@zorn-v no, then don't. 😄 

I really want to combine all symfony recipes into one in next v7. Let's merge it now for v6 although. 

Can you add init script for symfony flex?

recipe/symfony-flex.php
Cryde: My bad ! Sorry !

zorn-v:
What do you mean by 'init script' ?

antonmedv:
https://github.com/deployphp/deployer/tree/master/src/Initializer/Template

zorn-v:
Yes, I already understand what do you mean 😄 
How do you think, do it with accounting of DotEnv in prod ?

"Best practice" (env vars) not so conveniently... You need to set params in 3 different places. 
1. In deployer
2. For console commands on remote
3. For web server on remote

antonmedv:
Let's leave it for user? 😁 

Trim every thing related of setting params from recipe and add comment to init template about for to start using `.env`, if user want of cause :))

I like `.env` file.

zorn-v:
What do you think ?
Yes there need some workaround for `composer install` (coz scripts in composer.json)

src/Initializer/Template/SymfonyFlexTemplate.php
```diff
@@ -0,0 +1,44 @@
...
+    protected function getExtraContent()
+    {
+        return <<<PHP
+
+// Migrate database before symlink new release.
+
+before('deploy:symlink', 'database:migrate');
+
+//If you don't use DotEnv in prod just wipe it out
+task('clear:env', function () {
```
antonmedv: What purpose of this task?

zorn-v: DotEnv does not run at all if APP_ENV is set https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/3.3/public/index.php#L11

antonmedv: I think it's better not to hack APP_ENV but unstead write comment what user need to move dotenv to require section.

zorn-v: But this task needed exactly in order if DotEnv in require ;)
One more time - DotEnv does not run if APP_ENV is set.
If you don't set it in env, `composer install` will fail.

Yes I know that is hack. But I think that I can not convince community to change `!isset($_SERVER['APP_ENV'])` to `!file_exists(__DIR__.'/../.env')`

recipe/symfony-flex.php
```diff
@@ -0,0 +1,34 @@
...
+
+// Symfony web dir
+set('web_dir', 'public');
+
+// Assets
+set('assets', ['public/css', 'public/images', 'public/js']);
```
gander:
```php
set('assets', ['{{web_dir}}/css', '{{web_dir}}/images', '{{web_dir}}/js']);
```

recipe/symfony-flex.php
gander: Maybe:
```php
// Environment vars
set('env', function () {
    return [
        'SYMFONY_ENV' => get('symfony_env'),
        'APP_ENV'     => get('symfony_env'),
    ];
});
```

recipe/symfony-flex.php
LouTerrailloune: The default directory for logs is `var/log`  (since 3.4)

recipe/symfony-flex.php
gander: I can not set environment variables on my server so `Dotenv` is also included in the _production_ environment. There were complications. I noticed that setting `APP_ENV` disables reading the _.env_ file, therefore I set only `symfony_env` to have all commands with `--env = {{symfony_env}}` option

zorn-v: Hmm. Maybe you are right and your solution is much better than my hacky workaround https://github.com/deployphp/deployer/pull/1440#discussion_r159818610
AFAIR composer does not account `--env=`
Will test, and change it if all works fine.

antonmedv:
Okay, problem now with:
```
if (!isset($_SERVER['APP_ENV'])) {
    if (!class_exists(Dotenv::class)) {
        throw new \RuntimeException('APP_ENV environment variable is not defined. You need to define environment variables for configuration or add "symfony/dotenv" as a Composer dependency to load variables from a .env file.');
    }
    (new Dotenv())->load(__DIR__.'/../.env');
}
```

I think we just need mention not to refactor to:

```
(new Dotenv())->load(__DIR__.'/../.env');
```

Cryde:
Any news on this ?

antonmedv:
Need to figure out how to properly handle env. Soon will create my recipe.

zorn-v:
Any news, suggestions. Maybe need some suggestions ?

zorn-v:
It last project I just installed apache-pack and put SetEnv in .htaccess
ENV for deployer only needed for doctrine migrations (except APP_ENV=prod). Maybe think in this direction ?

antonmedv:
I’m thinking on suggesting of using doyens for prod. This will require APP_ENV rrmoval

antonmedv:
Done in 443de26e87d1c1f3185ebf445eb625bacaec23d6

zorn-v:
Maybe `var/sessions` also in shared ?

antonmedv:
`var` stands for variable?

antonmedv:
`var/sessions` yes, my bad. Should go to shared.

zorn-v:
About `writable_mode` - I get errors with `acl` one. `chown` is my bet )

antonmedv:
But you should set `writable_mode` in deploy.php

zorn-v:
Yep, but I talk about default one.

antonmedv:
No, but chown may require sudo. acl no. And acl more flexible. 

But I'm thinking to refactor it in v7.

Cryde:
Will there be a 6.0.1 ? for the missing ```var/sessions``` ? :p

antonmedv:
Yes :) 
Sorry for delay.

Cryde:
Don't be sorry ! You do an amazing work ! We can still add it ourself :D

PR #1443, opened 2017-12-08
more docs

staabm:
in the process of setting up a deployer setup I added this notes as this information where not available beforehand and we had to reverse engineer the sources to find out whats correct..


antonmedv:
I'm also unhappy with docs. And have big plans on improving it (from self documenting code to tutorials and casts.)

Issue #1449, opened 2017-12-08
ssh multiplexing initialization error: Bad configuration option


staabm:
| Q                 | A
| ----------------- | ---
| Issue Type        | Bug
| Deployer Version  | 6.0.3
| Local Machine OS  | Ubuntu16 lts
| Remote Machine OS | Ubuntu16 lts

```
 php -v
PHP 7.0.26-1+ubuntu16.04.1+deb.sury.org+1 (cli) (built: Nov 29 2017 09:51:01) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.0.26-1+ubuntu16.04.1+deb.sury.org+1, Copyright (c) 1999-2017, by Zend Technologies
    with blackfire v1.18.0~linux-x64-non_zts70, https://blackfire.io, by SensioLabs
```

### Description 
I am running into a `Bad configuration option` error when trying to use ssh with multiplexing

### Output log
```
➤ Executing task deploy:lock
[46.22.41.116] > if [ -f /www/www/test/relaunch/testdeployer/.dep/deploy.lock ]; then echo 'true'; fi
[46.22.41.116] < ssh multiplexing initialization
[46.22.41.116] < /home/mstaab/.ssh/known_hosts: line 1: Bad configuration option: |1|h/xxxx/yyyyy
[46.22.41.116] < /home/mstaab/.ssh/known_hosts: line 2: Bad configuration option: |1|xxxxx+yyyy//r08
[46.22.41.116] < /home/mstaab/.ssh/known_hosts: terminating, 2 bad configuration options


  [Deployer\Exception\RuntimeException (-1)]
  The command "if [ -f /www/www/test/relaunch/testdeployer/.dep/deploy.lock ]; then echo 'true'; fi" failed.
  Exit Code: -1 (Unknown error)
  Host Name: 46.22.yy.xx
  ================
  /home/mstaab/.ssh/known_hosts: line 1: Bad configuration option: |1|h/fxxxxxxtn/l6xxxxx1wq
  /home/mstaab/.ssh/known_hosts: line 2: Bad configuration option: |1|ayyyyyj+xyyyyy//r08
  /home/mstaab/.ssh/known_hosts: terminating, 2 bad configuration options


Exception trace:
 () at phar:///usr/local/bin/dep/src/Ssh/Client.php:96
[...]

PR #1461, opened 2017-12-19
print ssh commands when running in debug-mode (-vvv)

staabm:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

added more debug output in ssh-client to ease debugging of possible host-configuration problems.

hopefully it gets easier to find the root-cause for https://github.com/deployphp/deployer/issues/1449 with this change applied.

### DIFF
--- a/src/Ssh/Client.php
+++ b/src/Ssh/Client.php
@@ -70,7 +70,7 @@ class Client
             if ($this->output->isDebug()) {
                 $this->pop->writeln(Process::OUT, $host->getHostname(), $ssh);
             }
-            
+
             $process = new Process($ssh);
             $process
                 ->setTimeout($config['timeout'])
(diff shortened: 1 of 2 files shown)


src/Ssh/Client.php
```diff
@@ -81,6 +85,9 @@ public function run(Host $host, string $command, array $config = [])
         }
 
         $ssh = "ssh $sshArguments $host $become 'bash -s; printf \"[exit_code:%s]\" $?;'";
+        if ($this->output->isDebug()) {
```
staabm: not sure whether this is required in this case, as the process below uses this `pop->callback()`...

antonmedv:
I wanted to do the same. But it must go in next major v7 (not BC in v6).
I'm closing this for master branch and will manually merge into v7.

staabm:
Hmm where does it break BC?

antonmedv:
New output. Some Deployer users depends on parsing of `dep deploy -vvv` still.

PR #1465, opened 2017-12-20
simplify startTask()

staabm:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A 

use the same structure we use in endTask(), with a early return.

this simplifies the IF condition.

### DIFF
--- a/src/Console/Output/Informer.php
+++ b/src/Console/Output/Informer.php
@@ -32,11 +32,12 @@ class Informer
 
     public function startTask(Task $task)
     {
+        if ($task->isShallow()) {
+            return;
+        }
+        
         $this->startTime = round(microtime(true) * 1000);
-        if (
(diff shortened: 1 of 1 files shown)


src/Console/Output/Informer.php
```diff
@@ -32,11 +32,12 @@ public function __construct(OutputWatcher $output)
 
     public function startTask(Task $task)
     {
+        if ($task->isShallow()) {
+            return;
+        }
+
         $this->startTime = round(microtime(true) * 1000);
```
antonmedv: But we loose startTime here.

staabm: we use startTime only in one place inside this file and this case has the same IF guard

staabm: How to proceed?

staabm:
the few lines change is not worth discussing.

Issue #1445, opened 2017-12-08
commandSupportsOption: `[[: not found`


staabm:
| Q                 | A
| ----------------- | ---
| Issue Type        | Bug
| Deployer Version  | 6.0.3
| Local Machine OS  | Ubuntu16 lts
| Remote Machine OS | N/A

### Description 
when running my recipe I can see a `[[: not found` error, which is triggered by a call to `commandSupportsOption()`. I guess there is some kind of incompatibility with bash or similar, which makes this simple command trigger this error

### Output log
```sh
[localhost] > cd /cluster/www/www/www/xxx_relaunch/.build && (if [[ $(man ln 2>&1 || ln -h 2>&1 || ln --help 2>&1) =~ '--relative' ]]; then echo 'true'; fi)
[localhost] < sh: 1: [[: not found
```

### Env
```
php -v
PHP 7.0.26-1+ubuntu16.04.1+deb.sury.org+1 (cli) (built: Nov 29 2017 09:51:01) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.0.26-1+ubuntu16.04.1+deb.sury.org+1, Copyright (c) 1999-2017, by Zend Technologies
    with blackfire v1.18.0~linux-x64-non_zts70, https://blackfire.io, by SensioLabs
```

antonmedv:
Yes, maybe we should use `test` here as well.

JavierPons:
Hi there!
I would like to contribute in this project. What I can do? It will be my first contributing place.

antonmedv:
Yes we need! You can start fron any issue with label good for beginners

JavierPons:
@antonmedv thanks. I going to looking for something. Do you recommend any special?

JavierPons:
@antonmedv I have notice that good for beginners there are quite old issues. (correct me if I'm wrong). How can I go forward?

staabm:
You can start with this issue. Provide a PR and note the change in the Changelog

JavierPons:
@staabm hi! do you mean with PR (private repository) Im quite new in this .... :p Can you explain if I'm wrong?

JavierPons:
@staabm ok I will try

staabm:
PR = Pull Request. Github itself provides some great tutorials on howto provide a patch toma codebase, see https://help.github.com/

staabm:
[...]

PR #1467, opened 2017-12-20
use a single square-brocket in commandsSupportsOption() like we do in…

staabm:
… all other calls to test()

| Q             | A
| ------------- | ---
| Bug fix?      | Yes 
| New feature?  |  No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #1445

should fix "[[: not found" warning in `commandSupportsOption()`

> * Add description under added/changed/fixed section.
> * Add reference to closed issues `[#000]`.
> * Add link to issue in the end of document.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@
 [v6.0.5...master](https://github.com/deployphp/deployer/compare/v6.0.5...master)
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


staabm:
I guess it will fix #1445  not sure it will, as I cannot really reproduce this IF-case error on bash-only.

staabm:
this patch is wrong. I have no clue how to solve this warning.

PR #1468, opened 2017-12-21
print the task description when in debug mode

staabm:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes 
| BC breaks?    |  No
| Deprecations? | No
| Fixed tickets | N/A 

DX: print the task description when in debug mode, to get a human friendly message what the task is supposed todo.

especially for the builtin tasks this is very helpful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 ## master
 [v6.0.5...master](https://github.com/deployphp/deployer/compare/v6.0.5...master)
 
+### Added
+
+- DX: print the task description when in debug mode, to get a human friendly message what the task is supposed todo [#1468]
+
 ### Fixed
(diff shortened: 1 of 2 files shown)


antonmedv:
I don't see any need for this in debug mode. Description is helpful in cli.

staabm:
as a "consumer" of this library I dont know what the "internal" or "common" tasks are doing without looking into them.

I frequently need the source of this project open to be able to write down my recipe.

antonmedv:
As a maintainer, I don't see use of this feature. 
Maybe there other solution, like better docs or more verbosity levels, or better extension system which will allow this as plugin. 
Let's find out better, what kind of problem you are trying to solve.
For now, I'm closing this PR.

PR #1469, opened 2017-12-21
added deploy:islocked task

staabm:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes 
| BC breaks?    | No
| Deprecations? |  No
| Fixed tickets | N/A 

added `deploy:islocked` task which allows to determine if a lock was already acquired by another process/task. this can be usefull to fast fail before starting a extensive build-process and finally run into a "already locked" error after waiting mintues for a build.

### DIFF
--- a/recipe/deploy/lock.php
+++ b/recipe/deploy/lock.php
@@ -29,3 +29,8 @@ desc('Unlock deploy');
 task('deploy:unlock', function () {
     run("rm -f {{deploy_path}}/.dep/deploy.lock");//always success
 });
+
+desc('Checks whether deploy is locked');
+task('deploy:islocked', function () {
+    return test("[ -f {{deploy_path}}/.dep/deploy.lock ]");
+});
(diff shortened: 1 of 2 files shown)


recipe/deploy/lock.php
```diff
@@ -29,3 +29,8 @@
 task('deploy:unlock', function () {
     run("rm -f {{deploy_path}}/.dep/deploy.lock");//always success
 });
+
+desc('Checks whether deploy is locked');
+task('deploy:islocked', function () {
```
staabm: hmm tasks cannot return a value it seems. how to provide this api in a correct way?

staabm:
hmm this doesnt make sense. to safely call islocked you need all the "remote prepare" things beforehand, so all the heavy stuff was already done at this step.

antonmedv:
There is no need to extra task.
You can call `deploy:lock` before build step.

staabm:
but this will error in case the lock can't be acquired. (and it only works after `deploy:prepare` succeeded once.

staabm:
ahh I see... you can lock repeatedly. great.

staabm:
actually it doesnt.. with https://github.com/deployphp/deployer/pull/1470 it will allow to invoke deploy:lock repeatedly

PR #1470, opened 2017-12-21
make sure `deploy:lock` can be called several times, when the lock wa…

staabm:
…s acquired successfully once

| Q             | A
| ------------- | ---
| Bug fix?      | Kindof
| New feature?  | Yes 
| BC breaks?    | Kindof
| Deprecations? | No
| Fixed tickets | N/A

Make sure we can use `invoke('deploy:lock');` several times within the recipe without errors

this changes allows recepies like
```php
task('build', function() {
    //...
    
    invoke('deploy:prepare'); // create basic folder structure
    invoke('deploy:lock');
    //...
    // runLocally('{{composer_command}}');
    // runLocally('npm install');
    // runLocally('{{grunt_command}}');
});

task('release', function() {
    //...
    
    invoke('deploy:prepare'); // create basic folder structure
    invoke('deploy:lock');
    //...
    // invoke('deploy:release');
    // upload($releaseTar, '{{release_path}}');
    // invoke('deploy:symlink');
    //...   
});

task('deploy', [
    'build',
    'release',
])->shallow();
```

without this change I can not do `dep deploy` because it would throw a error on the 2nd `deploy:lock` invocation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@
 [v6.0.5...master](https://github.com/deployphp/deployer/compare/v6.0.5...master)
 
 ### Fixed
-- fix within() to also restore the working-path when the given callback throws a Exception [#1463]
+- Make sure we can use invoke('deploy:lock'); several times within the recipe without errors [#1470]
+- Fix within() to also restore the working-path when the given callback throws a Exception [#1463]
 
 ## v6.0.5
 [v6.0.4...v6.0.5](https://github.com/deployphp/deployer/compare/v6.0.4...v6.0.5)
@@ -338,7 +339,7 @@
 - Fixed typo3 recipe
 - Fixed remove of shared dir on first deploy
 
-
+[#1470]: https://github.com/deployphp/deployer/pull/1470
 [#1463]: https://github.com/deployphp/deployer/pull/1463
 [#1455]: https://github.com/deployphp/deployer/pull/1455
 [#1452]: https://github.com/deployphp/deployer/pull/1452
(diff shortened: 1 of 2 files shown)


antonmedv:
This is bad solution. You should use `deploy:lock` only once.
```php
task('deploy', [
    'deploy:lock',
    'build',
    'release',
]);
```

staabm:
but then I cannot invoke build or deploy on its own

antonmedv:
For what?

staabm:
In dev I need only build.

 For debugging its easy to do the separate steps one by one

antonmedv:
Okay, let me think about solution for it.

antonmedv:
But if build fails, you can run only release task.

staabm:
not sure what you mean. I can ran each step indivually. I need todo those 2 steps in correct order, so a realease without a build beforehand doesnt work.

most of the time I will use "deploy" which is an alias for build+release.

antonmedv:
Can you tell me again, why it's not possible to run `deploy:lock` first?

staabm:
I can call "deploy:lock" only once per request, because calling it a second time within the same process makes it throw.

I have 3 use cases:
A) `dep deploy` (doing both at once)
```
task('deploy', [
    'build',
    'release',
]);
```

B: `dep build; /* some times goes by... */ dep release;` 
```
task('deploy', [
    'build',
    'release',
]);
```

C: `dep build` (without a release)

antonmedv:
Can you run dep release if deploy locked?

staabm:
I can only run dep:release if I not make it acquire the lokc, but this kinda makes no sense

antonmedv:
You can run unlock

staabm:
Ooohh ofcourse. I can lock+unlock immediately in build. That will solve it, thx

antonmedv:
It’s like mutex. Should be occurred only once

PR #1471, opened 2017-12-21
update vendors

staabm:
| Q             | A
| ------------- | ---
| Bug fix?      | No 
| New feature?  | Yes 
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A 

Updated vendor libraries. set php 7.0.8 as min version, because its required for symfony components 3.4.* versions. 

there is no symfony component for >=7.0 & <=7.0.8 because php7 had some nasty bugs in those early 7.0.x versions.

I know this change could lead to some discussion, therefore I will not yet add CHANGELOG and similar.
In case we agree on this change, I can finish it up.

Motivation for this change is some strange behaviour with `uoload()`, which seems to only work in 40-50%. there could be some kind of race condition in the php impl.  without changes in the recipe it hangs rather often for me

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,7 @@
     "config": {
         "sort-packages": true,
         "platform": {
-            "php": "7.0.0"
+            "php": "7.0.8"
         }
     }
 }
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "content-hash": "d765474b620c917395370a0d39fdc911",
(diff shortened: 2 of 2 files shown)


staabm:
per changelog there is no documented BC break in any of our referenced components
https://github.com/symfony/console/blob/3.4/CHANGELOG.md
https://github.com/symfony/process/blob/3.4/CHANGELOG.md
https://github.com/symfony/yaml/blob/3.4/CHANGELOG.md

antonmedv:
Dropped config at all.

staabm:
thx.

FWIW, I created a symfony process upstream issue for the hanging rsync+multiplex problem
https://github.com/symfony/symfony/issues/25580

as the update of the vendors did not fix the issue for me :-/

PR #1476, opened 2017-12-27
prevent warning when pcntl_fork() is on disable_functions php.ini bla…

staabm:
…cklist

| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A 

`pcntl_fork` is blacklisted per default on ubuntu lts boxes. make sure deployer doesnt emit a warning in this case.

Without this fix you get
```
PHP Warning:  pcntl_fork() has been disabled for security reasons in phar:///usr/local/bin/dep/src/Utility/Reporter.php on line 22
```

### DIFF
--- a/src/Utility/Reporter.php
+++ b/src/Utility/Reporter.php
@@ -17,7 +17,7 @@ class Reporter
     public static function report(array $stats)
     {
         $pid = null;
-        if (extension_loaded('pcntl')) {
+        if (extension_loaded('pcntl') && strpos(ini_get('disable_functions'), 'pcntl_fork') === false) {
             declare(ticks = 1);
             $pid = pcntl_fork();
         }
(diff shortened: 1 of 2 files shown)


src/Utility/Reporter.php
```diff
@@ -17,7 +17,7 @@ class Reporter
     public static function report(array $stats)
     {
         $pid = null;
-        if (extension_loaded('pcntl')) {
```
antonmedv: Maybe it's better to use `function_exists`? If it available?

staabm: `function_exists` also works. I feel it is less expressive, but nevertheless changed it.

staabm:
Friendly ping @antonmedv

staabm:
Fixed build

Issue #1481, opened 2017-12-28
logfile does not contain exception messages


staabm:
| Q                 | A
| ----------------- | ---
| Issue Type        | Feature Request
| Deployer Version  | 6.0.3
| Local Machine OS  | N/A
| Remote Machine OS | N/A

### Description 
when using the `--log` option I see the progress and messages of the deployment in the given logfile.
when the deployment runs into a exception and stopps, the message of the exception is missing from the logfile.

it would be very helpful to have at least the message of the occured exception in the logfile, at best even the stacktrace next to it.

### Steps to reproduce
`dep errors --log deployer.log`
### Content of `deploy.php`
```php
namespace Deployer;

require 'recipe/common.php';

task('errors', function() {
    run('echo "test"'); // create some output, works correctly
    throw new \Exception("boooom!"); // this exception message is missing from the logfile
});
```

### Output log
*With enabled option for verbose output `-vvv`.*
```
[localhost] > echo "test"
[localhost] < test
```

staabm:
Implemented with https://github.com/deployphp/deployer/pull/1482

PR #1482, opened 2017-12-28
Log unhandled Exceptions into logfile

staabm:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes 
| BC breaks?    |  No
| Deprecations? | No
| Fixed tickets | #1481

Log unhandled Exceptions into logfile

### DIFF
--- a/src/Console/Application.php
+++ b/src/Console/Application.php
@@ -9,13 +9,13 @@ namespace Deployer\Console;
 
 use Deployer\Component\PharUpdate\Console\Command as PharUpdateCommand;
 use Deployer\Component\PharUpdate\Console\Helper as PharUpdateHelper;
+use Deployer\Deployer;
 use Symfony\Component\Console\Application as Console;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputDefinition;
 use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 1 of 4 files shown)


src/Console/Application.php
```diff
@@ -162,4 +163,15 @@ public function afterRun($callable)
     {
         $this->after = $callable;
     }
+
+    public function renderException($e, $output)
```
staabm: hmm depeding on the version of symfony console we use (prefer-stable vs. prefer-lowest) php expects a different signature for this method... not sure how to handle this case...

staabm: looking at it from a different angle. does it actually make sense that we CI build deployer with php 7.0/7.1/7.2 and prefer-stable vs. prefer-lowser, as the deployer binary/executable itself will finally be bundled with whatever the build process will decide?

it will always contain what prefer-latest will suggest and noone will use this repos in a prefer-lowest fashion, right?

antonmedv: Actually a lot of people use it as dev dependency of composer. And we need to support all of them. Mybe there is a better place to handle this logging. See CommandTask.

staabm:
very good advice, thx for your feedback. adjusted and it seems test are passing now.

PR #1483, opened 2017-12-28
reduce number of run() calls in cleanup-task

staabm:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | No
| BC breaks?    | I guess No
| Deprecations? | No
| Fixed tickets | N/A

optimize `cleanup` task for speed by saving a single `run()` call.
in my setup this saves up to 10% of deployment time (measured with blackfire).

`run()` is a rather expensive operation because of the network, ssh and x-process overhead. every call counts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ## master
 [v6.0.5...master](https://github.com/deployphp/deployer/compare/v6.0.5...master)
 
+### Changed
+- reduce number of run() calls in cleanup-task to optimize performance [#1483]
+
 ### Added
 - Added a description to the autocomplete command [#1472]
 
@@ -22,7 +25,7 @@
 ### Changed
 - Added support for GroupTask in invoke() [#1364]
 - Magento2 recipe optimizes the autoloader after the DI compilation [#1365]
-- Host's `roles()` API now can accept arrays too 
+- Host's `roles()` API now can accept arrays too
 - Fixed bug where wrong time format is passed to touch when deploying assets [#1390]
 - Added artisan:migrate:fresh task for laravel recipe
 - Added platform config to composer.json [#1426]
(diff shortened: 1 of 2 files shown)


antonmedv:
This is really bad for clean code and readability. 
Performance in this two runs insignificant. 
In v7 i'm plannig to add batched command support. This is what you want.

staabm:
In my setup this reduced bulit time of 10 seconds (I guess since no multiplexing).

I am looking forward for batch support and how this will be implemented.

Thanks

antonmedv:
Will something like this:
```
begin();
run('...');
run('...');
// ...
commit(); // run it here
```
But API may change.

staabm:
awesome.

maybe it could also batch automatically, when switching between different hosts like run() and runLocally() (so e.g. deployer auto-batches everything without additional code since deployer already knows which commands will have to be executed on which host)

antonmedv:
This is an implicit behavior. And it's bad. Better use explicit syntax.

staabm:
Wouldnt batch mode technically just `&&` all commands?

I dont like begin/commit as I guess it will imply some kind atomic command, what I guess we wouldnt achieve.

Proposal
```
runBatch(function() {
  run(...);
  test(...);
  supportsCommand(...);
  runLocally(...);
  // ...
});

**Comment by antonmedv** [comment]
How to achieve this?

```
runBatch(function() {
  run(...);
  if(test(...)) {
     run(...);
  } else {
     run(...);
  }
  // ...
});
```

Which cond to run?

**Comment by staabm** [comment]
Hmm good question. Maybe we would name the commands by string instead like

```
runBatch([
  'rm -rf ..',
  'If [test ...] cp ...',
]);
```

Or maybe forbidd branches in a batch?

**Comment by antonmedv** [comment]
Sooo, what the difference with?
```
run('
    rm -rf ..
    If [test ...] cp ...
');
```

antonmedv:
begin and commit allow you to inject it in already working code without if's in it.

staabm:
So lets go one step back: what are the goals of this new batch api?

What I thouht is the goal: performance. Batch several commands together into a single one to reduce ssh/network/bashßcreation overhead

antonmedv:
Split complex command creations (commands containing loops, etc)

PR #1512, opened 2018-01-21
Use POSIX style if

### LINKED ISSUE #1445: commandSupportsOption: `[[: not found`

staabm:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes or No
| New feature?  | Yes or No
| BC breaks?    | Yes or No
| Deprecations? | Yes or No
| Fixed tickets | #1445

Will hopefully fix #1445 . not tested yet.

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -727,7 +727,7 @@ function commandExist($command)
 
 function commandSupportsOption($command, $option)
 {
-    return test("[[ $(man $command 2>&1 || $command -h 2>&1 || $command --help 2>&1) =~ '$option' ]]");
+    return test("[ $(man $command 2>&1 || $command -h 2>&1 || $command --help 2>&1) =~ '$option' ]");
 }
 
 /**


staabm:
before this change
```
[localhost] > if [[ $(man svn 2>&1 || svn -h 2>&1 || svn --help 2>&1) =~ 'status' ]]; then echo 'true'; fi
[localhost] < sh: 1: [[: not found
```

after this change
```
[localhost] > if [ $(man svn 2>&1 || svn -h 2>&1 || svn --help 2>&1) =~ 'status' ]; then echo 'true'; fi
[localhost] < sh: 1: [: svn(1): unexpected operator
```

so it changed one warning for another.. not sure what the correct patch should look like

antonmedv:
Well, `[[` is much easier to use. Try to install bash on your system.

staabm:
I have a regular ubuntu where bash is installed per default.

antonmedv:
So check what login with deployer user has bash by default. `[[: not found` bash has `[[`.

staabm:
```
dep run --hosts=devmst16 -- 'echo $0'
bash
```

I guess I use bash..

zorn-v:
On ubuntu `/bin/sh` is symlink to `dash` not `bash`

staabm:
@zorn-v  thx for your input.

both suggested commands do not work.
the mentioned commands do not return "TRUE" when the given option is supported by the command

1st: 
runs withour error but does not provide the correct result

2nd: 
contains a syntax error

I tried to fix the second to
```sh
test("[ \"\$(case `man $command 2>&1 || $command -h 2>&1 || $command --help 2>&1` in *$option*)  echo '1'; esac)\" = '1' ]");
```

but this doesnt provide the correct results either.

any insights?

staabm:
Ping @zorn-v

zorn-v:
Maybe need some escaping. I test those command in shell.

zorn-v:
I think it need to do all things via /bin/bash (fuck FreeBSD) or introduce some var (default to /bin/bash, not /bin/sh - important for ubuntu) that tells thru wich shell commands should execute (/usr/local/bin/bash for FreeBSD for ex.) than torment with such things...

antonmedv:
Yes, I think car for bash is ok. I think we can relay on bash3 as it can be found anywhere

staabm:
I dont know exactly what you guys are talking about, as I am not a bash guy.

In case you want me to adjust a bit in this PR please give me a more concrete hint and I am happily adjust whatever is required.

antonmedv:
Problem here is more wide: it's nice to have `[[` `]]` as it more flexible. Maybe instead we need to always use bash.

zorn-v:
I just try to replace "bashism" like `=~` in "POSIX". I don't know what excatly "not working" at you. Then I think about not breaking mind, and use `bash` as shell for ALL commands (I suspect that in deployer `/bin/sh` is used in `test` php function)

PR #1521, opened 2018-01-24
Missing {{bin/php}} in deploy:vendors

gander:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A


mintao:
The {{bin/php}} part is now defined twice. In the common.php file line 119 and in the vendors.php in line 15.
Now the call looks like `/usr/bin/php /usr/bin/php ./composer.phar` which, of course, fails.

PR #1532, opened 2018-02-05
Redact sensitive information provided by the user via askHiddenRespon…

### LINKED ISSUE #1527: allow to mask sensitive output in verbose modes
| Q                 | A
| ----------------- | ---
| Issue Type        | Feature Request
| Deployer Version  | 6.0.5
| Local Machine OS  | N/A
| Remote Machine OS | N/A

### Description 
in our recipe we ask the user for login credentials which we later on need to pass to some commands.
in our case we use `svn .... --username plainUserName --password plainPassword`.

while we ask the user for the password the input is hidden, but the credentials are rendered in plain text in the stdout while commands run.

it would be great if we could somehow blacklist things, so the plain passwords arent visible while running or even logged into file.

staabm:
…se()

| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes 
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #1527

strings which are provided by the end-user with `askHiddenResponse()` will be redacted in the outputstream and also in the logfile.

closes #1527

used in action

```
dep secret -vvv --log=deployer.log
➤ Executing task secret
tell me your secret
[localhost] > echo **SECRECT**
[localhost] < **SECRECT**
• done on [localhost]
✔ Ok [1s 919ms]

mstaab@mst16:/cluster/www/www/www/tools/dev/deploytest$ cat deployer.log
[localhost] > echo **SECRECT**
[localhost] < **SECRECT**

```


antonmedv:
This is a nice addition. Will look more closely on weekend. But I think of getting more common solution.

antonmedv:
Also check CI.

staabm:
fixed the build.

I had the need for this feature therefore I implemented it in this simple form.
IMO we could land it in this simple form and do a more abstracted/common one in a later PR.

staabm:
rebased. ping @antonmedv

antonmedv:
I think this cool feature can go to v7. Motivating people to upgrade (a lot still uses v3 :( )

PR #1533, opened 2018-02-05
added phpdoc to ProcessRunnter->run()

staabm:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A 

just adding docs, escpecially for the Exception case which you might want to handle on invalid run() commands.

intentionally without a CHANGELOG.md as it seems not required.


staabm:
ping @antonmedv ... should be a no-brainer.

PR #1542, opened 2018-02-13
Don't touch files

### LINKED ISSUE #1540: Required to touch existing files?
| Q                 | A
| ----------------- | ---
| Issue Type        | Bug
| Deployer Version  | 6
| Local Machine OS  | N/A
| Remote Machine OS | Ubuntu

### Description 
When a shared file does not yet exist, an empty file is created. This doesn't always make sense, because the file is usually not correct. This is usually the case with environment files, eg Laravel `.env` or Magento `app/etc/env.php`. 
For Laravel this is not a problem, but Magento expects an array for example. See https://github.com/magento/magento2/pull/13090

### Steps to reproduce
Deploy for the first time with that file.

### Solution?

Do we need to touch the file? If it doesn't exist, we can still create the symlink, which will be activated once the file is created, right? If so, should we just check if the file actually exists before touching, or do we not need `touch` at all?

barryvdh:
Fixes #1540

| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | Yes or No
| Deprecations? | No
| Fixed tickets | #1540

Removes the `touch` so that non-existing files are not created, possibly in an invalid format.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -349,7 +349,7 @@
 - Fixed typo3 recipe
 - Fixed remove of shared dir on first deploy
 
-
+[#1540]: https://github.com/deployphp/deployer/issues/1540
 [#1521]: https://github.com/deployphp/deployer/pull/1521
 [#1513]: https://github.com/deployphp/deployer/pull/1513
 [#1481]: https://github.com/deployphp/deployer/issues/1481
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Throw meaningfull exception on errors in cd() [#1480]
 - Make sure Context::pop() is called when Callback errors in on(...) function [#1513]
 - Update silverstripe recipe to support silverstripe 4
+- Don't create shared files automatically when they do not exist yet [#1540]
 
 ## v6.0.5
 [v6.0.4...v6.0.5](https://github.com/deployphp/deployer/compare/v6.0.4...v6.0.5)
--- a/recipe/deploy/shared.php
+++ b/recipe/deploy/shared.php
@@ -64,9 +64,6 @@ task('deploy:shared', function () {
         // Ensure dir is available in release
         run("if [ ! -d $(echo {{release_path}}/$dirname) ]; then mkdir -p {{release_path}}/$dirname;fi");
 
-        // Touch shared
-        run("touch $sharedPath/$file");
-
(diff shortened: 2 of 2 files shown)


antonmedv:
I'm afraid what this is breaking change. How can we figure this out?

staabm:
I have the same feeling... this might be a breaking change... maybe this should be feature-flagged instead of beeing deleted, so you can opt-out from this touch call in your recipe.

antonmedv:
I think we need to see history of this addition and why it was introduced in first place.

barryvdh:
Yeah I was trying to, but at some time there was a large refactor so the history was hard to track down.

The symlink would still be created right? Just as a broken link, which (imho) is easier to notice then an empty file.

gharlan:
> I think we need to see history of this addition and why it was introduced in first place.

Apparently it was introduced along with whole `deploy:shared` task in 2014:
https://github.com/deployphp/deployer/commit/339baedd143ac7f7b393c9fd43959b729d0d64f5#diff-17bfdc9b1a5d2375e8f0241c0bee0619R84

barryvdh:
I guess we could also show a warning message when this doesn't exist, or fail.

But imho it seems to be more dangerous/unexpected to create an empty file, because some frameworks might check for the existence of a file and then require it or change behavior.

antonmedv:
@gharlan your are right. Actually shared_dirs also created. What about them? 

I think this is kind of breaking changes. I think to add this to v7 milestone.

barryvdh:
So another use-case about this:
Magento has a 'maintenance' flag in 'var/.maintenance.flag'. If that exists, it's in maintenance mode, otherwise not. Currently you can't share that file, because if it doesn't exist you create it, so Magento thinks it's in maintenance mode..

antonmedv:
Agree. Will be better to not create file. What about shared dirs?

cafferata:
If you ask me, I would answer; no shared directory creation and an showing an warning message is enough. Because this situation will only occurs when you bootstrap the project. In our situation the shared directories contains an `.gitkeep` file to prevent from not existing.

jordisala1991:
What about a flag to disable/enable this feature... and disabled by default, because otherwise it would be BC break

antonmedv:
I think one more flag is redundant. Will merge in into v7 major release.

PR #1545, opened 2018-02-13
parse input before using dirname()

gharlan:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

In my case `$dir` is something like `{{media_dir}}`, and this is set to `public/media`.
`dirname('{{media_dir}}')` returns `'.'`, but `'public'` is needed.


antonmedv:
Please, add tests and update changelog.

staabm:
with all those minor changes it would be great we could get a new release

Issue #1475, opened 2017-12-27
Support after/before/fail($it, callable);


staabm:
| Q                 | A
| ----------------- | ---
| Issue Type        | Feature Request
| Deployer Version  | N/A
| Local Machine OS  | N/A
| Remote Machine OS | N/A

### Description 
with the current version I need to define a private task which then need to be wired with after/before/fail.
it would be easier if I could just register a callable, like already possible with `task()`, e.g.

```php
after('success', function() {
  $chatBot->msg('successfully deployed... yay');
});
fail('deploy', function() {
  $chatBot->msg('deploy task fu**ed up');
});
```

antonmedv:
Added support in master!

PR #1553, opened 2018-02-16
PoC [WIP]

TheLuckyOne:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #1475 

Hi. What do you think about this implementation?

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -171,40 +171,71 @@ function task($name, $body = null)
  * Call that task before specified task runs.
  *
  * @param string $it The task before $that should be run.
- * @param string $that The task to be run.
+ * @param string|callable $that  Callable task or the task to be run.
  */
 function before($it, $that)
 {
     $deployer = Deployer::get();
     $beforeTask = $deployer->tasks->get($it);
 
-    $beforeTask->addBefore($that);
+    if (is_callable($that)) {
+        $taskName = 'before_' . $it;
+        $task = new T($taskName, $that);
+        $deployer->tasks->set($taskName, $task);
+    } elseif (is_string($that)) {
+        $taskName = $that;
+    } else {
+        throw new \InvalidArgumentException('Task should be an closure or other task.');
+    }
+
+    $beforeTask->addBefore($taskName);
 }
 
 /**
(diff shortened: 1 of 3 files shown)


antonmedv:
What is a few tasks was added?

TheLuckyOne:
I think for multiple tasks in before/after we need group task 
`$task = new GroupTask($name, $body);`

antonmedv:
I think it should not must be unique, otherwise it will rise a lot of questions.

src/functions.php
```diff
@@ -171,40 +171,71 @@ function task($name, $body = null)
...
 
-    $beforeTask->addBefore($that);
+    if (is_callable($that)) {
+        $taskName = 'before_' . $it . '_' . (count($beforeTask->getBefore()) + 1);
+        $task = new T($taskName, $that);
+        $deployer->tasks->set($taskName, $task);
+    } elseif (is_string($that)) {
+        $taskName = $that;
+    } else {
+        throw new \InvalidArgumentException('Task should be an closure or other task.');
```
staabm: other task name

TheLuckyOne:
Guys, tell me how can i improve my PR

antonmedv:
Again, what about adding a few callbacks this way? Really want this feature.

TheLuckyOne:
@antonmedv Feature added in "Added dynamic task name in before/after".

staabm:
sorry I cannot follow your conversation? what exactly is missing?

marvinhinz:
This is actually a very useful feature, what needs to be done to merge this?

antonmedv:
I think it should be rebased.

test/src/FunctionsTest.php
```diff
@@ -131,6 +131,28 @@ public function testBefore()
         self::assertEquals(['before', 'main'], $names);
     }
 
+    public function testCallableBefore()
+    {
+        task('main', 'pwd');
+        before('main', function () {
```
staabm: you should assert that this callables are actually invoked

PR #1555, opened 2018-02-18
Remove composer lock file

bestlong:
Not suggest include composer.lock

| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A


antonmedv:
composer.lock used for phar creation. Why to remove?

bestlong:
@antonmedv composer.lock will created when run `composer install`.

antonmedv:
I know, This file lock specified versions.

adriandmitroca:
.lock files should always be in every repository, without any exceptions really.

antonmedv:
@adriandmitroca not for libraries :) 
But in case of phar creation, it's ok.

adriandmitroca:
@antonmedv 

After some thinking you are definitely right. Thanks for putting me on right direction! <3

bestlong:
@antonmedv please ref https://travis-ci.org/laradock/laradock/jobs/342684374
have any suggest?

bestlong:
switch to use Phar version.

PR #1556, opened 2018-02-22
added becomeSetHome method to add -H flag when using sudo

gabrielef:
I have added the becomeSetHome method to add the -H flag when become is used. In this way is possible to use home directory of the current user.

### DIFF
--- a/src/Host/Host.php
+++ b/src/Host/Host.php
@@ -24,6 +24,7 @@ class Host
     private $identityFile;
     private $forwardAgent = true;
     private $multiplexing = null;
+    private $becomeSetHome = true;
     private $sshArguments;
 
     /**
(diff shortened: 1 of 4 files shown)


gabrielef:
This feature add the -H flasg to the current implementation of the become() method:
`sudo -H -u <user>`

In this way the ~ directory will be the _/home/user_ directory. This could be useful because not in all systems is the default behaviour as reported in the sudo man:

> -H, --set-home
Request that the security policy set the HOME environment variable to the home directory specified by the target user's password database entry. Depending on the policy, this may be the default behavior.

Ansible is doing this as a default behaviour.

antonmedv:
Maybe we can do it as default too and merge in into v7?

gabrielef:
I think it could be good as a default behaviour too.

My current workaround, using version 6.x is to declare become as below:

`->become('{{become_user}} -H ')`

src/Host/Host.php
```diff
@@ -209,31 +210,49 @@ public function multiplexing(bool $multiplexing = true)
...
+    /**
+     * @return bool
+     */
+    public function isBecomeSetHome()
+    {
+        return $this->becomeSetHome;
+    }
+
+    /**
+     * @param bool $becomeSetHome
```
staabm: would be great to have more docs on this "setter", so people know why/when this switch will be usefull.

antonmedv:
Let's use -H by default and merge it then.

PR #1564, opened 2018-03-09
added blackfire captured metrics for subprocesses and DI calls

staabm:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | No (no deployer enduser feature)
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

utilize [metrics argument capturing](https://blackfire.io/docs/reference-guide/metrics#metrics-capturing-arguments) to get a better understanding of which cli commands are actually slow. also added a profiler node per DI call.

so with this change one can get a better picture of what is going on when running blackfire profiles. 

**before this patch**
(processes)
grafik
grafik

**after this patch**
(processes)
grafik
(magic DI calls)
grafik

### DIFF
--- a/.blackfire.yml
+++ b/.blackfire.yml
@@ -13,4 +13,3 @@ metrics:
         - callee:
             selector: '=Deployer\Deployer::__get' # aggregate the costs of all Deployer::__get calls
             argument: { 1: "^" }  # but create separate nodes by the first argument value
-                        
--- a/.blackfire.yml
+++ b/.blackfire.yml
@@ -1,0 +1,16 @@
+metrics:
+  deployer.process_runner: # metric name
+    label: "ProcessRunner processes"
+    matching_calls:
+      php:
+        - callee:
+            selector: '=Deployer\Utility\ProcessRunner::run' # aggregate the costs of all ProcessRunner::run calls
+            argument: { 2: "^" }  # but create separate nodes by the second argument value
+  deployer.di_container: # metric name
+    label: "DI Container magic calls"
+    matching_calls:
+      php:
+        - callee:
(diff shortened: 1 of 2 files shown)


antonmedv:
Wow, this is cool. But do we been blackfire in main repo? What is purpose of this? What bottleneck did you found this this?

staabm:
this file is required in the main repo, as the configuration inside is very specific on classes/methods used by deployer.

Its especially usefull for the subprocess component, as otherwise you only see that deployer spends 99% of its time in running processes. since we run so many of them and each have different perf characterisitics its very helpful to see which of the proccess takes so much time.

the DI thing is not as usefull, but I thought it could be worth it, as the blackfire.yml file is now created with this PR anyway.

antonmedv:
That's cool. But my question is more about why we need blackfire.io in first place? :)

staabm:
I use it for profilling my builds as those take mintues and I want to know where the time is spent.

antonmedv:
So, what insides?

antonmedv:
Did you find what was slowing build?

staabm:
I found it in my builds. It highly depends on the cli commands one runs. 

With this file everyone using deployer and blackfire can easily profile the process. 

Its not about finding a problem inside deployer but about easing profilling the custom recipes/builds of deployer users

antonmedv:
Maybe implement same log output as composer does with —profile flag?

staabm:
Habing more out of the box timing infos would also be a good alternative... yes

staabm:
@antonmedv any chanche to get this merged.. blackfire.io is a very widely used php profiler nowadays..?

antonmedv:
Hi,

This is old conversation here :) Actually I still don't see points of merging this into main repo. I think blackfire.io is more for end projects.

PR #1574, opened 2018-03-23
Laravel recipe: added maintenance text and retry header options on artisan:down

DaanGeurts:
When putting a Laravel application in maintenance mode you can set a message to display and a value to the Retry-After HTTP header. 
I've added questions when dep artisan:down and also added a check for Laravel > 5.3.
Some input validation is needed.

| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

### DIFF
--- a/recipe/laravel.php
+++ b/recipe/laravel.php
@@ -62,7 +62,6 @@ task('artisan:down', function () {
     $minVersion = 5.3;
     $currentVersion = get('laravel_version');
     $output = '';
-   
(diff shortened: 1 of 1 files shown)


antonmedv:
Maybe create `option` for it too?

DaanGeurts:
Yeah could be useful!

antonmedv:
If you specify - don't ask.

PR #1589, opened 2018-04-11
Fix setfacl operation not permitted bug for symfony4 recipe

vuras:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    |  No
| Deprecations? |  No
| Fixed tickets | N/A


antonmedv:
So, what's the problem?

vuras:
The problem is that when only var directory is set as writable dir and the user on a host machine is not root then setfacl operation not permitted error is thrown.

antonmedv:
You can use sudo with writeable, or use chown iinstead of.

Issue #1496, opened 2018-01-09
Question / Bug shared_dirs (recipe/deploy/shared.php)


ahalimkara:
| Q                 | A
| ----------------- | ---
| Issue Type        | Bug, Question
| Deployer Version  | 6.0.5
| Local Machine OS  | MAC
| Remote Machine OS | Debian (PHP 7.1)

### Description 
I just want to set `/image` folder as shared directory, but I am getting following error:

```
In Client.php line 96:

  [Deployer\Exception\RuntimeException (1)]
  The command "ln -nfs --relative /var/www/deployments/project.com/shared/image/ /var/www/deployments/project.com/releases/1/image/" failed.

  Exit Code: 1 (General error)

  Host Name: project.com

  ================
  ln: target '/var/www/deployments/project.com/releases/1/image/' is not a directory: No such file or directory
```

When I check the code of `recipe/deploy/shared.php`, shared directory is removed from source [here](https://github.com/deployphp/deployer/blob/703292c1e92b4537ca4105f74fcceaf1acc6f589/recipe/deploy/shared.php#L38) and when creating the directory it is using parent directory of `shared_dir` [here](https://github.com/deployphp/deployer/blob/703292c1e92b4537ca4105f74fcceaf1acc6f589/recipe/deploy/shared.php#L33) and [here](https://github.com/deployphp/deployer/blob/703292c1e92b4537ca4105f74fcceaf1acc6f589/recipe/deploy/shared.php#L42). It means, it is removing `/var/www/deployments/project.com/releases/1/image/` and creating `/var/www/deployments/project.com/releases/1/` which is already exists and root folder. So when it tries to link it fails.

I fixed this error by removing `dirname` from [here](https://github.com/deployphp/deployer/blob/703292c1e92b4537ca4105f74fcceaf1acc6f589/recipe/deploy/shared.php#L33) and [here](https://github.com/deployphp/deployer/blob/703292c1e92b4537ca4105f74fcceaf1acc6f589/recipe/deploy/shared.php#L42).

PS: I was using deployer for my Laravel projects successfully with `recipe/laravel.php`, I don't understand how it was working so far...

### Steps to reproduce
It is a simple OpenCart installation. You can test it with a fresh OpenCart installation or any php project should produce this error. 

### Content of `deploy.php`
```php
require 'recipe/common.php';

set('shared_files', ['image/']);
...
```

### Output log
*It is in Description.*

ahalimkara:
[...]

PR #1604, opened 2018-05-02
remove trailing slash on shared dir

ryanpark91:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #1496

### DIFF
--- a/recipe/deploy/shared.php
+++ b/recipe/deploy/shared.php
@@ -23,7 +23,9 @@ task('deploy:shared', function () {
     }
 
     foreach (get('shared_dirs') as $dir) {
+        var_dump($dir);
         $dir = rtrim($dir, '/');
+        var_dump($dir);
         // Check if shared dir does not exist.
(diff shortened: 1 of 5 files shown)


recipe/deploy/shared.php
```diff
@@ -23,7 +23,9 @@
     }
 
     foreach (get('shared_dirs') as $dir) {
+        var_dump($dir);
```
antonmedv: Vad dump?

ryanpark91: this was temporary code to check my code. Now, it was removed.

ryanpark91:
@antonmedv 
I add test for this changes. But, I found another bug and I can't find any solution.

If I set shared_dirs by variable, its impossible to use rtrim() on shared.php
```
set('dist', 'public/dist/');
set('shared_dirs', [
    '{{dist}}'
]);
```

PR #1618, opened 2018-05-17
Update symfony4 recipe

zorn-v:
Make writable after cache warmup


markuspoerschke:
The `cache:clear` and `cache:warmup` will happen on an empty directory. Therefore the `var/cache` directory is writable by the deploy user. Therefore it makes sense to make the cache directory writable after cache was warmed up.

Of course this will not work if you try to share the cache directory among releases (which is not recommended).

@zorn-v can you provide more input about the concrete problem you want to solve?

Otherwise this PR could be closed in my opinion.

zorn-v:
> Of course this will not work if you try to share the cache directory among releases

Why ? Shared dir is symlink. And `writable` task follow symlinks.

After `cache:warmup` all files in cache dir is belongs to deploy user, which can be NOT the same as webserver user (`www-data` for ex.). And webserver can not write in cache dir then. Therefore usually there no problem with this coz all cache warmed up already, but in some cases app need to write in that dir for some custom app cache for example.

antonmedv:
But you can configure right dir perm first time

PR #1634, opened 2018-06-15
locate binary with less subprocesses

staabm:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

to optimize resource consumption.

before this change, the number of subprocesses required to locate a binary was 2 in the best case and 6 in the worst case.

after this change it always requires 1 process, no matter which command will be used at the end (whether `command -v`, `which` or `type -p`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
 - Make used shell configurable via `shellCommand` [#1536]
 - Added `cleanup_tty` option for `deploy:cleanup`
 
+### Changed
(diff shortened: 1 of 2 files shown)


staabm:
@antonmedv any feedback?

antonmedv:
Sorry, for delay. Will test it today and merge it.

staabm:
Ping

PR #1638, opened 2018-06-21
Fix return types for fluent interface

lstrojny:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | Kinda
| Deprecations? | No
| Fixed tickets | N/A

### DIFF
--- a/src/Host/Host.php
+++ b/src/Host/Host.php
@@ -84,11 +84,7 @@ class Host
         return $this->config->parse($this->realHostname);
     }
 
-    /**
-     * @param string $hostname
-     * @return $this
-     */
-    public function hostname(string $hostname)
+    public function hostname(string $hostname): self
     {
         $this->setRealHostname($hostname);
         return $this;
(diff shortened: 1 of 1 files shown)


antonmedv:
What is it?

lstrojny:
Are you asking what the PR contains? It fixes all fluent interfaces in the Host class to return self, instead of `@return $this` (which is wrong) or `Host` as a type hint (which can be lifted to be `self`).

antonmedv:
And why BC break kinda?

lstrojny:
Since `Host` is a non-final class, somebody could have extended Host and overridden a method that doesn't have a return type hint as of now. If we now add one, this could render declaration could become incompatible.

```
<?php
class Host
{
    public function methodReturningSomething()
    {}
}

class MyHost extends Host {
    public function methodReturningSomething(): MyHost
    {}
}
```

This works just fine. This will trigger a fatal error though:
```
<?php
class Host
{
    public function methodReturningSomething(): Host
    {}
}

class MyHost extends Host {
    public function methodReturningSomething(): MyHost
    {}
}
```

```
Fatal error: Declaration of MyHost::methodReturningSomething(): MyHost must be compatible with Host::methodReturningSomething()
```

antonmedv:
I see, I think it's okay.

PR #1656, opened 2018-07-23
Opcode optimizations

lfluvisotto:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

See this article: https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls

### DIFF
--- a/recipe/config/dump.php
+++ b/recipe/config/dump.php
@@ -29,9 +29,9 @@ task('config:dump', function () {
     }
 
     foreach ($config->getCollection() as $name => $value) {
-        if (is_array($value)) {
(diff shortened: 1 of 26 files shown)


staabm:
At best this would be enforced via a php-cs-fixer config

antonmedv:
Thanks for your contribution. 
But performance improvements like this, doesn't add speedup real deployment, but decreases readability. 
Also I'm personally don't like how function's call looks. I prefer to omit `\`.

PR #1657, opened 2018-07-23
Avoid callables in loop conditionals for better performance

lfluvisotto:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

### DIFF
--- a/bin/dep
+++ b/bin/dep
@@ -64,7 +64,7 @@ $includes = [
 $loaded = false;
 $includePath = false;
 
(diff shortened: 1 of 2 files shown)


antonmedv:
Thanks for your contribution. 
But performance improvements like this, doesn't add speedup real deployment, but decreases readability.

PR #1660, opened 2018-07-26
[deploy:writable] chgrp doesn't need http_user

Slamdunk:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

### DIFF
--- a/recipe/deploy/writable.php
+++ b/recipe/deploy/writable.php
@@ -22,7 +22,7 @@ task('deploy:writable', function () {
         return;
     }
 
-    if ($httpUser === false && $mode !== 'chmod') {
+    if ($httpUser === false && ! in_array($mode, ['chgrp', 'chmod'], true)) {
         // Detect http user in process list.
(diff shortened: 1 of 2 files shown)


recipe/deploy/writable.php
```diff
@@ -22,7 +22,7 @@
         return;
     }
 
-    if ($httpUser === false && $mode !== 'chmod') {
+    if ($httpUser === false && ! in_array($mode, ['chgrp', 'chmod'], true)) {
```
antonmedv: But you need http_user: `run("$sudo chgrp -RH $httpGroup $dirs", $runOpts);`

Slamdunk: No you don't: the full code of `chgrp` is:

https://github.com/deployphp/deployer/blob/62002e7a07fbcd4009d5695c85d5d997e6efd208/recipe/deploy/writable.php#L48-L56

And here the only variable used is `http_group`, no mention to `http_user`

jamieburchell:
http_group is not mentioned in any documentation

PR #1667, opened 2018-08-04
Adds POSIX style brackets to commandSupportOption

gpietoso:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes 
| New feature?  |  No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #1445 

Changed BASH only brackets to POSIX brackets on commandSupportOption method.


antonmedv:
I think bash op `[[` `]]` more powerful and we should use it. Also I'm not sure if it is save to delete `[`. Will it work the save way?

PR #1669, opened 2018-08-07
Added horizon:terminate for Laravel recipe

lasselehtinen:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

Laravel has a dashboard package for Redis queue jobs. Most typical configuration is to run php artisan horizon through supervisor and you are supposed to restart the process during deployment by running php artisan horizon:terminate. This PR adds this command.

There is no easy and clean way to determine whether the project is using Horizon or not, so the end user has to add the artisan:horizon:terminate after deploy:symlink.

https://laravel.com/docs/5.6/horizon#deploying-horizon

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Added `cleanup_tty` option for `deploy:cleanup`
 - Added Prestashop 1.6 recipe
 - Set dedicated user variable under CI environments, if not provided by git-config
+- Added artisan:horizon:terminate task for laravel recipe
 
 ### Changed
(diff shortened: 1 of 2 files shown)


antonmedv:
I think such easy task can be implemented in user space code. Also I'm using systemd, for example.
So thank you for contributing, but I think this task shouldn't be merged into common laravel recipe.
Now I'm working on better website with better docs and think such task can be added there as docs.

Issue #1666, opened 2018-08-02
Invalid composer configuration or misleading documentation


Bonscho:
| Q                 | A
| ----------------- | ---
| Issue Type        | Bug, Documentation
| Deployer Version  | 6.2.0
| Local Machine OS  | Mac OS X 10.11
| Remote Machine OS | Ubuntu 16.04LTS

### Description 
Misleading or incorrect documentation information about `composer_action` and `composer_options`. 

### Steps to reproduce
My expected behavior is that the following configuration runs composer skipping dev resources and optimizing autoloading. 

### Content of `deploy.php`
```php
set('composer_action', 'install');
set('composer_options', '--optimize-autoloader --no-dev');
```

### Output log
Result is 
`cd /var/www/xxx/releases/17 && /usr/bin/php /var/www/xxx/releases/17/composer.phar --optimize-autoloader --no-dev` 
which is neither expected nor valid. The task exits with a RuntimeException. 

### Workaround
```php
set('composer_action', '');
set('composer_options', 'install --optimize-autoloader --no-dev');
```

works as expected.

antonmedv:
Agree, we need to refactor this part of docs or code.

syntaxseed:
The problem is that the deploy/vendor.php recipe is missing the composer_action setting when executing composer:

`run('cd {{release_path}} && {{bin/composer}} {{composer_options}}');`

Should be:
`run('cd {{release_path}} && {{bin/composer}} {{composer_action}} {{composer_options}}');`

antonmedv:
Fixed in master. Please verify.

Bonscho:
Could not verify. After installing deployer from dev-master (upgrading from 6.2.x) I am running into general trouble executing the dep task with following result: 

`vendor/bin/dep deploy test -vv`

```
 Error  in deploy.php on line 54:

  Call to undefined method Deployer\Host\Host::stage()

#0 /var/www/awo-bildungscampus/vendor/deployer/deployer/src/Deployer.php(291): require()
#1 [internal function]: Deployer\Deployer::Deployer\{closure}()
#2 /var/www/awo-bildungscampus/vendor/deployer/deployer/src/Deployer.php(300): call_user_func(Object(Closure))
#3 /var/www/awo-bildungscampus/vendor/deployer/deployer/src/Deployer.php(272): Deployer\Deployer::load('/var/www/awo-bi...')
[...]

PR #1673, opened 2018-08-09
Fix 1666

alpipego:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | Maybe
| Deprecations? | No
| Fixed tickets | #1666 

### Possible BC break

I've added a check if `composer_options` contains `composer_action`, here is where I potentially see a  problem: if a user has added the composer action as a string to their `composer_options` variable the detection in https://github.com/deployphp/deployer/compare/master...alpipego:fix-1666#diff-8f884efdf2b864c9ee15854766038c1eR19 will still probably fail.


antonmedv:
Thanks for contributing. But i think to refactor this part completely and in more intuitive way.

Issue #1675, opened 2018-08-09
chmod problem


it-can:
| Q                 | A
| ----------------- | ---
| Issue Type        | Bug?
| Deployer Version  | 6.2.0
| Local Machine OS  | mac os x 10.13.6
| Remote Machine OS | Centos 7

### Description 
I have a couple of Laravel apps that I deploy. Now I use chmod to make dirs writable, but I noticed that all the files in these directories also get a chmod. I expected that all directories were updated, but not the files. Now I set ```writable_chmod_recursive``` to ```false``` and now it doesn't do this, but now it will not update chmod recursively all directories.
I think the writeable script needs to be updated so that it only changes the chmod on directories.  (http://patrickward.com/2012/03/04/recursively-chmod-only-directories-or-only-files/)

### Content of `deploy.php`
```php
namespace Deployer;

require 'recipe/laravel.php';

// Config
set('keep_releases', 3);
set('allow_anonymous_stats', false);
set('application', 'Test');
set('repository', 'gitrepo');
//set('git_tty', true);
set('git_recursive', false);
set('ssh_multiplexing', true);
set('writable_mode', 'chmod');

// add('shared_files', []);
// add('shared_dirs', []);
add('writable_dirs', [
    'storage/app/public/uploads',
    'storage/app/public/uploads/product',
    'storage/app/public/uploads/blog',
    'storage/app/public/uploads/category',
]);

// Hosts
host('server')
    ->user('test')
    ->set('branch', 'master')
    ->set('deploy_path', '/home/test');

// Tasks
desc('Reload PHP-FPM service');
task('php-fpm:reload', function () {
    // The user must have rights for restart service
    // /etc/sudoers: username ALL=NOPASSWD:/bin/systemctl restart php-fpm
    run('sudo systemctl reload php-fpm');
});

// Build assets locally
desc('Build assets');
task('local:npm:build', function () {
    runLocally("npm run production");
});

desc('Commit assets');
task('local:git:assets', function () {
    runLocally("git add . && git diff-index --quiet HEAD || git commit -a -m 'compress assets' && git push");
});

// Run tests
desc('Run PHPUnit');
task('local:phpunit', function () {
    runLocally("phpunit");
});

desc('Deploy your project');
task('deploy', [
    'deploy:info',
    'local:phpunit',
    'local:npm:build',
    'local:git:assets',
    'deploy:prepare',
    'deploy:lock',
    'deploy:release',
    'deploy:update_code',
    'deploy:shared',
[...]

PR #1676, opened 2018-08-09
Fix recursive chmod files issue

it-can:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | Possible
| Deprecations? | No
| Fixed tickets | #1675


antonmedv:
Thanks for contributing.
Why writable_chmod_recursive should be performed only on dirs?

it-can:
@antonmedv it uses the writable_dirs variable? It suggests chmodding directories. I don't think all files should be chmodded...

antonmedv:
Why only dirs and not files?

it-can:
Why would all files need to be chmodded to 755? For my projects it is not needed to chmod all my files within the storage directories to 755...

antonmedv:
But what if users want to chmod files as well? This PR brings some kind of BC. What if someone expect this recipe to chmod files too?

it-can:
@antonmedv agree...

PR #1681, opened 2018-08-27
Laravel deploy task should not call cache:clear

leo108:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? |  No
| Fixed tickets | N/A

`cache:clear` command will clear users' sessions if CACHE_DRIVER is the same as SESSION_DRIVER, I think users should call this task on demand instead of putting it in default deploy task.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ## master
 [v6.3.0...master](https://github.com/deployphp/deployer/compare/v6.3.0...master)
 
+### Changed
+- Laravel recipe should not run `artisan:cache:clear` in `deploy` task
+
 
 ## v6.3.0
 [v6.2.0...v6.3.0](https://github.com/deployphp/deployer/compare/v6.2.0...v6.3.0)
--- a/recipe/laravel.php
+++ b/recipe/laravel.php
@@ -168,7 +168,6 @@ task('deploy', [
     'deploy:writable',
     'artisan:storage:link',
     'artisan:view:clear',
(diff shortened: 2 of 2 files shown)


antonmedv:
ping @lasselehtinen @lucasmezencio

lucasmezencio:
I think you're right @leo108..
LGTM @antonmedv.

Pull the trigger! 🙂

antonmedv:
Please update changelog

leo108:
@antonmedv Updated

antonmedv:
@lasselehtinen sorry for distracting.

barryvdh:
Imho you woud usually want to clear the cache upon deployments. I understand your issue, but you should split your redis cache database from the other redis databases (sessions, queues etc). This is the default for new 5.7 installs, see https://github.com/laravel/laravel/pull/4665

matthew-muscat:
@antonmedv — i think the behaviour here should be that a cache clear is part of the recipe by default — as mentioned by @barryvdh , this is now the default in v5.7 installs, and can be configured by the user in earlier releases.

Flushing the cache during a deployment is what i would consider as expected behaviour — if you have your caches + sessions in the same location (ie: on the same redis db), then this is a bug in the setup or cache clearing logic, rather than an issue with how deployer is operating.

FYI — Laravel's `artisan cache:clear` command describes the action as the following.. 
`Flush the application cache`

PR #1715, opened 2018-09-27
improve symfony recipe to php7

rvitaliy:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

makes symfony recipe more strictness with php7 syntax

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -1,4 +1,4 @@
-<?php
+<?php declare(strict_types=1);
 /* (c) Anton Medvedev <anton@medv.io>
  *
  * For the full copyright and license information, please view the LICENSE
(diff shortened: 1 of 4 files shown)


antonmedv:
I don't see reason for doing this only for symfony recipe. And for recipies at all :) For code -maybe.

rvitaliy:
It's a start to migrate the code to be more strict and avoid unnecessary bugs (🐛🐛🐛) related to different types of data.
I started from symfony recipes because it's the one I use, and it's a test to see if the deployphp community approves these improvements.

IMHO: this kind of work should be done on all the code, where it is possible.

antonmedv:
What about future auto detect type mechanism in php?

rvitaliy:
I'm not aware of this feature. what is it?

antonmedv:
There is no proposal yet, but it's cool feature there return types can be automatically getted from code.

rvitaliy:
It's an interesting feature, but it's not yet reality.
so if we want to have a more strict code today, the only way to do that is to add typehint manually

rvitaliy:
if it's ok for you, sometimes I could contribute by doing some PR related to php7 improvements.
if you want you could prepare an issue with the lineup of priorities, otherwise I do my own thing.

PR #1751, opened 2018-10-25
[Support] upgrade code to php 7.0

rvitaliy:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

Improve code with php 7.0 syntax

### DIFF
--- a/src/Support/Changelog/Changelog.php
+++ b/src/Support/Changelog/Changelog.php
@@ -1,4 +1,4 @@
-<?php
+<?php declare(strict_types=1);
 /* (c) Anton Medvedev <anton@medv.io>
  *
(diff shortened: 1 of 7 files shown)


ahmadmayahi:
The issue says `upgrade code to PHP 7.0` but what you are doing is using the `declare()` function and return types in some of the files.

Could you please explain why only a few files have been changed?

rvitaliy:
@ahmadmayahi is a chain of several small PR to upgrade code to php 7 standards and make it more strict
see #1721, #1733, #1736

Issue #1755, opened 2018-11-06
Avoid unnecessary releases


ahmadmayahi:
| Q                 | A
| ----------------- | ---
| Issue Type        | Feature Request
| Deployer Version  | 6.3.0
| Local Machine OS  | All
| Remote Machine OS | All

### Description 
Each time I run `dep deploy ...`, the deployer will immediately work, and this is how it suppose to be, but sometimes I don't have any changes in my code, and if I run `dep deploy ...` a new unnecessary release will be created.

For example:

I use `at` command to run deployer every day at midnight:
```bash
at midnight dep deploy prod
```

Deployer has to skip the deployment if no changes were made into the git repo.

### Solution
Whenever deployer finishes, a new file contains the last commit id will be created/updated in `.dep` folder.
The next time we deploy, deployer has to check if the current commit id matches the existing saved one, if so then it will skip the deployment.

I will work on this if accepted.

antonmedv:
Nice idea

Chrico:
Howdy!

In versoin 6.4.7, this new "feature" will break deployment, if you don't have set any `repository`-URL because you're doing e.G. a deploy via CI by running composer + rsync to push the code to server.

This command should only run if `repository` is set: https://github.com/deployphp/deployer/blob/v6.4.7/recipe/deploy/prepare.php#L43-L45

![image](https://user-images.githubusercontent.com/3417446/62861187-e9b44300-bd02-11e9-9877-52c85b55d351.png)

ahmadmayahi:
@Chrico You just need to set the `check_remote_head` as follows:
```php
# your ci deploy file.
set('check_remote_head', false)
```

Chrico:
`check_remote_head` is `false` by default. The problem is this line: https://github.com/deployphp/deployer/blob/v6.4.7/recipe/deploy/prepare.php#L45  which runs even if the `repository`  is not set.

Additonally it could be possible, that the server has no access to the repo itself. So you should probably use `runLocally` instead of `run` here as well.

ahmadmayahi:
@Chrico You are right, I'll look into that. Thanks.

antonmedv:
I dot this. :) See 85aae9fd681028c28b27a5e118d4cd3422484c8d

antonmedv:
Usage:
```php
require 'recipe/deploy/check_remote.php';
after('deploy:prepare', 'deploy:check_remote');
```

antonmedv:
[...]

PR #1759, opened 2018-11-07
Add check_remote_head option to avoid unnecessary new releases by che…

ahmadmayahi:
Checking the remote git HEAD without cloning the repo.

| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #1755

A new option `check_remote_head` has been added, when this option sets to `true` deployer will not clone the repository if there are no changes found.

Thanks to `git ls-remote`, this command gets the `HEAD` without cloning the repository.

Revision is also checked, so if we run something like `dep deploy --revision=a69539703d917cafb1c51dad20f4ee1e317c3017` deployer will also check for the revision.

### DIFF
--- a/recipe/deploy/prepare.php
+++ b/recipe/deploy/prepare.php
@@ -42,8 +42,7 @@ task('deploy:prepare', function () {
     // This avoid unnecessary releases when the last commit id matches the existing one (HEAD)
     $repository  = trim(get('repository'));
     $revision    = input()->getOption('revision') ?? null;
-    $remoteHead  = $revision ? $revision :
-        run(sprintf('%s ls-remote %s HEAD | tr -d "HEAD"', get('bin/git'), $repository));
+    $remoteHead  = $revision ?? run(sprintf('%s ls-remote %s HEAD | tr -d "HEAD"', get('bin/git'), $repository));
 
     if (true === get('check_remote_head') && null == input()->getOption('tag')) {
         $headPath = trim(get('deploy_path').'/.dep/HEAD');
--- a/recipe/deploy/prepare.php
+++ b/recipe/deploy/prepare.php
@@ -36,4 +36,26 @@ task('deploy:prepare', function () {
 
     // Create shared dir.
     run("cd {{deploy_path}} && if [ ! -d shared ]; then mkdir shared; fi");
+
+    // Check the remote HEAD
+    if (true === get('check_remote_head') && null == input()->getOption('tag')) {
+        $repository  = trim(get('repository'));
+        $revision    = input()->getOption('revision') ?? null;
+        $remoteHead  = $revision ? $revision :
+            trim(run(sprintf('%s ls-remote %s HEAD | tr -d "HEAD"',
+        get('bin/git'), $repository)));
+
+        $headPath   = trim(get('deploy_path').'/.dep/HEAD');
+        $isRemoteHeadExists = (bool) run('if [ -f '.$headPath.' ]; then echo 1; else echo 0; fi');
+
+        //check if HEAD file is exists and then compare it
+        if (true === $isRemoteHeadExists) {
+            $headContents = run('cat '.$headPath);
+            if ($headContents === $remoteHead) {
+                writeln('<info>Already up-to-date.</info>');
+                exit(0);
+            }
+        }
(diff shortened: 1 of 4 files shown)


antonmedv:
Build failed

ahmadmayahi:
> Build failed

It keeps saying `Deployer\Support\Changelog\ParseException: Error in compare link syntax` I have no idea why?

antonmedv:
Try checkout CHANGELOG and run `php bin/changelog`

ahmadmayahi:
I tried that, but still get the same result:
Image of Yaktocat

antonmedv:
Really strange, will look at this tonight.

ahmadmayahi:
Thank you @cafferata

recipe/deploy/prepare.php
```diff
@@ -36,4 +36,26 @@
...
+    if (true === get('check_remote_head') && null == input()->getOption('tag')) {
+        $headPath = trim(get('deploy_path').'/.dep/HEAD');
+        $isRemoteHeadExists = (bool) run('if [ -f '.$headPath.' ]; then echo 1; else echo 0; fi');
+
+        //check if HEAD file is exists and then compare it
+        if (true === $isRemoteHeadExists) {
+            $headContents = run('cat '.$headPath);
+            if ($headContents === $remoteHead) {
+                writeln('<info>Already up-to-date.</info>');
+                exit(0);
```
antonmedv: I think exiting program like this is dangerous and not flexible. What is some one want's to add hook what will be triggered all the time? Maybe throw GrasefulSutdown excetion?

recipe/deploy/prepare.php
```diff
@@ -36,4 +36,26 @@
...
+    // Check and save the remote HEAD/revision and compare it with the existing saved one (if any)
+    // This avoid unnecessary releases when the last commit id matches the existing one (HEAD)
+    $repository  = trim(get('repository'));
+    $revision    = input()->getOption('revision') ?? null;
+    $remoteHead  = $revision ? $revision :
+        trim(run(sprintf('%s ls-remote %s HEAD | tr -d "HEAD"',
+            get('bin/git'), $repository)));
+    if (true === get('check_remote_head') && null == input()->getOption('tag')) {
+        $headPath = trim(get('deploy_path').'/.dep/HEAD');
+        $isRemoteHeadExists = (bool) run('if [ -f '.$headPath.' ]; then echo 1; else echo 0; fi');
```
antonmedv: You can use `test()` fn here. Also better to use double quotes `"[ -f $headPath.]"` as it more readable.

recipe/deploy/prepare.php
```diff
@@ -36,4 +37,25 @@
...
+    // Check and save the remote HEAD/revision and compare it with the existing saved one (if any)
+    // This avoid unnecessary releases when the last commit id matches the existing one (HEAD)
+    $repository  = trim(get('repository'));
+    $revision    = input()->getOption('revision') ?? null;
+    $remoteHead  = $revision ? $revision :
+        trim(run(sprintf('%s ls-remote %s HEAD | tr -d "HEAD"',
+            get('bin/git'), $repository)));
+    if (true === get('check_remote_head') && null == input()->getOption('tag')) {
+        $headPath = trim(get('deploy_path').'/.dep/HEAD');
+        $isRemoteHeadExists = test("[ -f $headPath ]");
```
staabm: Is this check really required, or could it be achieved in combination with the „cat“ line in a single command?
Every remote command has a big overhead

ahmadmayahi: Well, `cat`ing non-existing file leads to `STDERR`.
I know, we could output the error to the `2>/dev/null` but I think it's a hack, the right solution is to check whether the file exists or not.

staabm: And something along the lines of
`run('test -f '. $headPath ' && cat '.$headPath);`

ahmadmayahi: This is great.

antonmedv: Nope, the https://en.wikipedia.org/wiki/Test_(Unix)

staabm:
ping @antonmedv

PR #1762, opened 2018-11-10
Code style fixes

cafferata:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

Some small code styles fixes:
1. Update missing argument in PHPDoc.
1. Splits the IF-statement/workflow(s) for readability
1. Replace double quotes to single quotes.
1. Replaced alias functions by original functions.
1. Define return types.
1. Type casting can be used
1. Non-optimal regular expression, \d is a subset of \w.

### DIFF
--- a/bin/changelog
+++ b/bin/changelog
@@ -34,7 +34,7 @@ $app->register('update')->setCode(function () use ($io) {
 
     $item = new Item();
 
-    $message = $io->ask("What is " . strtolower($type) . '?');
+    $message = $io->ask('What is ' . strtolower($type) . '?');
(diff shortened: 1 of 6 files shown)


cafferata:
I have removed, on your request, the two specific commits in this pull request.
1. Using fully-qualified function calls is faster.
1. Unnecessary variable overrides

antonmedv:
1. Yes, but I this case it wouldn't increase speed.
2. I think clean code is better anyway 🙃

PR #1771, opened 2018-12-07
Only share the local file

trsteel88:
https://symfony.com/doc/current/configuration/dot-env-changes.html

Recent changes mean that .env should no longer be shared. It is the fallback if no .env.local exists. Instead we should share the .env.local files.

| Q             | A
| ------------- | ---
| Bug fix?      | Yes or No
| New feature?  | Yes or No
| BC breaks?    | Yes or No
| Deprecations? | Yes or No
| Fixed tickets | N/A or xx

>
>
>     php bin/changelog
>

### DIFF
--- a/recipe/symfony4.php
+++ b/recipe/symfony4.php
@@ -10,7 +10,7 @@ namespace Deployer;
 require_once __DIR__ . '/common.php';
 
 set('shared_dirs', ['var/log', 'var/sessions']);
-set('shared_files', ['.env']);
+set('shared_files', ['.env.local']);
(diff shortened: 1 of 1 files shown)


antonmedv:
What about old user what relay on this recipe?

ahmadmayahi:
These changes are only for `Symfony 4.2`+ but deployer has to support all the Symfony versions.

adrienfr:
Fixed by https://github.com/deployphp/deployer/pull/1881

Issue #1778, opened 2018-12-20
http_user detection command can return wrong user


krukru:
| Q                 | A
| ----------------- | ---
| Issue Type        | Bug/Feature Request
| Deployer Version  | v6.3.0
| Local Machine OS  | Tried on macOS 10.14.1 (but shouldn't be relevant in this case)
| Remote Machine OS | Ubuntu server 16.04, Running apache2

### Description 
The command for detecting http_user (`ps axo comm,user | grep -E '[a]pache|[h]ttpd|[_]www|[w]ww-data|[n]ginx' | grep -v root | head -1 | awk '{print $2}'`) returned a user who is not the true webserver user, in my case I am running apache webserver.
In turn this prevents writable_dirs to work properly.

Why does this happen and who was the user returned? 

On my server, I have a service running called "apache_exporter", owned by user "foo".
When running the first part of the command `ps axo comm,user | grep -E '[a]pache|[h]ttpd|[_]www|[w]ww-data|[n]ginx'` I can see that "apache_exporter" service survives grep and pops up on top of the list and in turn the owner of this services (foo) gets returned as `http_user` in deployer.

Maybe a better strategy would be to throw an exception when there are multiple unique http_user candidates, and request the user to manually set this variable. This would of course break BC in a very bad way, but would never result in a bad build.

How to mitigate: Manually set http_user. Also note in docs that this detection strategy produces fuzzy results, and suggest manually setting whenever possible.

### Steps to reproduce

Create new daemon service called "apache_exporter", run it under some user who is not your webserver user.
Run deployer, but do not manually set http_user

Expected: http_user will be that of the webserver
Actual: http_user equals the user who owns apache_exporter service

antonmedv:
> Maybe a better strategy would be to throw an exception when there are multiple unique http_user candidates, and request the user to manually set this variable.

This will be awesome. Can you create it as PR?

krukru:
Yea, no problem :)

andrey-helldar:
@antonmedv, In the latest version of the deplorer, the problem has reappeared:

v.6.4.3 (working):
```
$httpUserCandidates = explode("\n", run("ps axo comm,user | grep -E '[a]pache|[h]ttpd|[_]www|[w]ww-data|[n]ginx' | grep -v root | sort | awk '{print $2}' | uniq"));
```

v6.4.4 (not working):
```
[...]

PR #1779, opened 2018-12-20
http_user detection command could return wrong user

krukru:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | Yes 
| Deprecations? | No
| Fixed tickets | #1778 

My Bash is really bad, so I would highlight the fact that this needs reviewing. I tried deploying on three servers (with 0/1/3 http_user candidates) and it worked as expected - only the server with a single unique candidate detected http_user. Of the three servers, two were ubuntu servers and one was my local machine which is macOS.

Another important note is that this will break BC for users who were lucky to have multiple http_user candidates, and the first one from the list was the correct one - so consider maybe delaying this for the next major release.

If there is anything else I can help please let me know :)

### DIFF
--- a/recipe/deploy/writable.php
+++ b/recipe/deploy/writable.php
@@ -23,8 +23,10 @@ task('deploy:writable', function () {
     }
 
     if ($httpUser === false && ! in_array($mode, ['chgrp', 'chmod'], true)) {
-        // Detect http user in process list.
-        $httpUser = run("ps axo comm,user | grep -E '[a]pache|[h]ttpd|[_]www|[w]ww-data|[n]ginx' | grep -v root | head -1 | awk '{print $2}'");
+        // Attempt to detect http user in process list.
+        if (test("[ $(ps axo comm,user | grep -E '[a]pache|[h]ttpd|[_]www|[w]ww-data|[n]ginx' | grep -v root | sort | uniq | wc -l) -eq 1 ]")) {
+            $httpUser = run("ps axo comm,user | grep -E '[a]pache|[h]ttpd|[_]www|[w]ww-data|[n]ginx' | grep -v root | sort | uniq | head -1 | awk '{print $2}'");
+        }
 
         if (empty($httpUser)) {
             throw new \RuntimeException(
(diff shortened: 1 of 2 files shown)


recipe/deploy/writable.php
```diff
@@ -23,8 +23,10 @@
     }
 
     if ($httpUser === false && ! in_array($mode, ['chgrp', 'chmod'], true)) {
-        // Detect http user in process list.
-        $httpUser = run("ps axo comm,user | grep -E '[a]pache|[h]ttpd|[_]www|[w]ww-data|[n]ginx' | grep -v root | head -1 | awk '{print $2}'");
+        // Attempt to detect http user in process list.
+        if (test("[ $(ps axo comm,user | grep -E '[a]pache|[h]ttpd|[_]www|[w]ww-data|[n]ginx' | grep -v root | sort | uniq | wc -l) -eq 1 ]")) {
+            $httpUser = run("ps axo comm,user | grep -E '[a]pache|[h]ttpd|[_]www|[w]ww-data|[n]ginx' | grep -v root | sort | uniq | head -1 | awk '{print $2}'");
```
antonmedv: What about using php here? Remove `head -1` get all lines and check on php side.

krukru: In that case, I have two suggestions
The first is 
```php
$httpUserCandidates = explode(',', run("ps axo comm,user | grep -E '[a]pache|[h]ttpd|[_]www|[w]ww-data|[n]ginx' | grep -v root | sort | uniq | awk '{print $2}' | paste -sd ',' -"));
if (count($httpUserCandidates) === 1) {
    $httpUser = $httpUserCandidates[0];
}
```
This way I don't really need to know what character is the line delimiter (\n, \r or \n\r) on the remote machine, since `paste` will solve this for me and produce comma delimited values.

But maybe this is overthinking things and the simpler approach is better
```php
$httpUserCandidates = explode("\n", run("ps axo comm,user | grep -E '[a]pache|[h]ttpd|[_]www|[w]ww-data|[n]ginx' | grep -v root | sort | uniq | awk '{print $2}'"));
if (count($httpUserCandidates) === 1) {
    $httpUser = $httpUserCandidates[0];
}
```

What would you suggest to go with?

krukru: Hey @antonmedv, I decided to go with the second snippet. If you think using `paste` is a better option, I can revert the commit and go with that option.
It would also be nice if there was some way to automatically check this recipe on multiple environments, and not just the ones I have ready. How do you in general test these recipes files?

PR #1792, opened 2019-01-14
improve ParallelExecutor::generateOptions to manage all types of InputOption

rvitaliy:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | https://github.com/deployphp/deployer/pull/1721#issuecomment-453936737

### DIFF
--- a/src/Console/Input/Argument.php
+++ b/src/Console/Input/Argument.php
@@ -1,0 +1,21 @@
+<?php declare(strict_types=1);
+/* (c) Anton Medvedev <anton@medv.io>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Deployer\Console\Input;
+
+use Symfony\Component\Console\Input\InputArgument;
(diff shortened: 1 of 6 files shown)


kick-the-bucket:
Just by looking at the code it seems like it now handle NO_VALUE options wrong by always adding them no matter if they are true or false (set or not).

I'll try to test it more thoroughly later today.

kick-the-bucket:
The previous fatals would've been cought just by adding something like this to any recipe test:
```php
option('value-none', null, InputOption::VALUE_NONE);
option('value-optional', null, InputOption::VALUE_OPTIONAL, 'blah');
option('value-required', null, InputOption::VALUE_REQUIRED);
option('value-array-required', null, InputOption::VALUE_IS_ARRAY | InputOption::VALUE_REQUIRED);
option('value-array-optional', null, InputOption::VALUE_IS_ARRAY | InputOption::VALUE_OPTIONAL, ['a' => 'b']);```

PR #1793, opened 2019-01-15
Properly handle input options

kick-the-bucket:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #1792 comment

### DIFF
--- a/src/Executor/ParallelExecutor.php
+++ b/src/Executor/ParallelExecutor.php
@@ -20,7 +20,6 @@ use Deployer\Host\Storage;
 use Deployer\Task\Context;
 use Deployer\Task\Task;
 use Symfony\Component\Console\Input\InputInterface;
-use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Process\Process;
 
@@ -269,7 +268,9 @@ class ParallelExecutor implements ExecutorInterface
             $inputs[] = Option::toString($this->input, $option);
         }
 
-        return implode(' ', $inputs);
+        return implode(' ', array_filter($inputs, function (string $item) {
+            return $item !== '';
+        }));
     }
 
     private function generateArguments(): string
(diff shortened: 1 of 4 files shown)


antonmedv:
Awesome!

kick-the-bucket:
Looks like it's still handling the array options wrong in some edge cases, but I'll try to fix those as well.

antonmedv:
@kick-the-bucket so? How it's going? Maybe delete `declare(strict_types=1);` if it helps?

kick-the-bucket:
@rvitaliy I did find a way to pass `null` to both `VALUE_OPTIONAL` and `VALUE_IS_ARRAY | VALUE_OPTIONAL`.

> @kick-the-bucket so? How it's going? Maybe delete `declare(strict_types=1);` if it helps?
The option handling issues are logical, rather than type-related, so sadly this won't help.
Option handling might have never been fully correct (it certainly never worked with array options)...

antonmedv:
So, everything looks ok to me. @kick-the-bucket @rvitaliy merging? 

And please update changelog by `php bin/changelog` command.

kick-the-bucket:
Looks like all possible options values are now being handled correctly which was probably never the case since the feature was implemented almost 3 years ago :sweat_smile:

rvitaliy:
LGTM 🍻

src/Executor/ParallelExecutor.php
```diff
@@ -269,7 +268,9 @@ private function generateOptions(): string
             $inputs[] = Option::toString($this->input, $option);
         }
 
-        return implode(' ', $inputs);
+        return implode(' ', array_filter($inputs, function (string $item): bool {
```
rvitaliy: in this case `Lambda function` can be `static`. [for more info see](http://php.net/manual/en/functions.anonymous.php#functions.anonymous-functions.static)

kick-the-bucket: TIL!

PR #1811, opened 2019-02-17
Changes to .env for symfony4 recipe

samuel-queniart:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | No
| BC breaks?    | No
| Deprecations? | Yes
| Fixed tickets | N/A

Why ? https://symfony.com/doc/current/configuration/dot-env-changes.html

`.env` is commited and now `.env.local` override `.env`

### DIFF
--- a/recipe/symfony4.php
+++ b/recipe/symfony4.php
@@ -10,7 +10,7 @@ namespace Deployer;
 require_once __DIR__ . '/common.php';
 
 set('shared_dirs', ['var/log', 'var/sessions']);
-set('shared_files', ['.env']);
+set('shared_files', ['.env.local']);
 set('writable_dirs', ['var']);
 set('migrations_config', '');
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### Fixed
 - Fix rsync upload honor become option for host [#1796]
 - Fixed bug to execute ssh command on windows [#1775]
+- Changes to .env for symfony4 recipe
 
 
 ## v6.4.3


alexander-schranz:
If this is changed it should be only released in the a new major version else it could break exist projects which didn't upgrade to the new dotenv structure. Alternative we maybe keep both `.env` and `.env.local` als shared file but not sure if this make problem if in the new structured projects the .env is also in the git repository.

recipe/symfony4.php
```diff
@@ -10,7 +10,7 @@
 require_once __DIR__ . '/common.php';
 
 set('shared_dirs', ['var/log', 'var/sessions']);
-set('shared_files', ['.env']);
+set('shared_files', ['.env.local']);
```
endelwar: Should `env.prod.local` be added here to?

roberto910907:
Not sure why we should share this file between releases. This file is meant to be used just in your local environment. Unless you're proposing this change for local, IMHO you're only encouraging people to use bad practices(using .env files in prod).

From [here](https://symfony.com/blog/improvements-to-the-handling-of-env-files-for-all-symfony-versions#what-changed-exactly):
> C) A .env.local file can now be created to override environment variables for your machine. This file is ignored in the new .gitignore.

[How to make sure you're not allowing .env files to be processed in prod environment.](https://symfonycasts.com/screencast/symfony-fundamentals/env-vars-tricks-production)

Best Practice:
https://symfony.com/doc/current/configuration/environment_variables.html#configuring-environment-variables-in-production

plashenkov:
@roberto910907

On each machine—developer's machine or production server—.env files are loaded in the following order:

* `.env`
* `.env.local`
* `.env.$APP_ENV`
* `.env.$APP_ENV.local`

where rules from later file override rules from the former.
You can "compile" all these into php file using this command:

`composer dump-env prod`

This command will generate `.env.local.php` file. If Symfony founds it, it (and only it) will be loaded instead of all `.env` files.

And this is the recommended practice. In this way you define default values, which can be overridden with real environment variables (real environment variables will take precedence over those you define in `.env` files or compiled `.env.local.php`).

So, `.env`, `.env.local`, `.env.$APP_ENV`, `.env.$APP_ENV.local` and `.env.local.php` files should be shared.

Fix me if I am mistaken.

alexander-schranz:
@plashenkov 

> So, .env, .env.local, .env.$APP_ENV, .env.$APP_ENV.local and .env.local.php files should be shared.

Not correct as `.env` and `.env.$APP_ENV` are files which are commited so they should not be shared as they are under the version control.

So only `.env.local` and `.env.$APP_ENV.local` would be the one which are not under version control and so need to be shared. As `.env.test.local` or `.env.dev.local` does for me make no sense to have on a deploy server I would only go for only `.env.local` and maybe `.env.local.prod`.

For BC promise it maybe would be better to go with a new `symfony4.2.php`  recipe which extends from symfony4.php.

alexander-schranz:
@plashenkov That need to be regenerated else it would not contain changes of the files (.env, ...) which are in the version control. But the dump command need to be optional as this is a symfony/flex only command and does not work with regular composer.

adrienfr:
Fixed by https://github.com/deployphp/deployer/pull/1881 right ?

samuel-queniart:
@adrienfr yes ;)

PR #1837, opened 2019-05-05
Add artisan:optimize:clear to Laravel recipe

adevade:
| Q | A
| - | -
| Bug fix? | No
| New feature? | Yes
| BC breaks? | No
| Deprecations? | No
| Fixed tickets | N/A

This PR adds `artisan:optimize:clear` task to the Laravel recipe. It's available for versions 5.7 and up. 
I did not add it to the deploy script. 

I've also readded support for `artisan:optimize` for versions 5.7 and up. This command was deprecated in 5.5, removed in 5.6 and readded in 5.7 with different functionality. 

The previous version (<5.5) optimized composers autoload files.
The new version (>5.7) caches the app config and routes. It's the same as running `config:cache` and `route:cache`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 ### Changed
 - Add lock and unlock task to flow_framework receipe
+- Updated `artisan:optimize` to run for Laravel 5.7 and above, since [it got added back](https://github.com/laravel/framework/commit/fe1cbdf3b51ce1235b8c91f5e603f1e9306e4f6f) last year. It still doesn't run for 5.5 and below.
 
 ### Fixed
 - Fix rsync upload honor become option for host [#1796]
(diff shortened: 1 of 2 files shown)


kyranb:
This fails if the routes file uses route closures.

Any ideas on how to check for this or make route caching optional?

antonmedv:
What did you mean? By route caching?

mpdx:
@antonmedv he meant that with out-of-the box routes that don't use controllers' method but closure, ie. https://github.com/laravel/laravel/blob/master/routes/api.php#L16 that will always result in failed deployment. Mohamed Said who maintains Laravel weaseled out the responsibility of fixing this bug https://github.com/laravel/framework/issues/22034 and in this PR #1837 you assumed that you can safely use optimize command, which you can't.

neamtua:
If this fails for you in the latest version just add this to your deploy script:

`
task('artisan:optimize', function () {});
`

PR #1858, opened 2019-06-06
Prevent dir creation when it is existing

kanevbg:
If "shared" directory is already existing just do not call mkdir, to prevent pointless update during processing of the "shared" files just like it's checked during "shared" directories processing.


kanevbg:
This saves some pointless update of the inode

kanevbg:
10x!

PR #1684, opened 2018-09-01
Added view:cache and auth:clear-resets commands 

DevDavido:
Added the view:cache command as added in Laravel [v5.6.13](https://github.com/laravel/framework/blob/5.6/CHANGELOG-5.6.md#v5613-2018-03-26) and the auth:clear-resets artisan command too.

| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

PR #1860, opened 2019-06-10
Laravel view cache task

alexsoft:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

Instead of only calling view:clear command, starting from 5.6 laravel version it is possible to cache views


matthew-muscat:
This should be considered a breaking change, as the "artisan:view:clear" command is now no longer available as a command for this recipe.

@antonmedv — what are your thoughts for restoring the `artisan:view:clear` command?

antonmedv:
`artisan:view:clear` maybe restore it with version detection?

matthew-muscat:
@antonmedv — i've restored this and modified the existing `artisan:view:cache` command to only run the `view:cache` command, as the `view:clear` is already completed by the framework

matthew-muscat:
see https://github.com/deployphp/deployer/pull/1889

DevDavido:
See also #1684

PR #1881, opened 2019-07-10
Update symfony4.php

Benjamin-Adrea:
On SF 4.1+, the .env file is versionned, only the .env.local is not versionned anymore

| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

### DIFF
--- a/recipe/symfony4.php
+++ b/recipe/symfony4.php
@@ -10,7 +10,7 @@ namespace Deployer;
 require_once __DIR__ . '/common.php';
 
 set('shared_dirs', ['var/log', 'var/sessions']);
-set('shared_files', ['.env']);
+set('shared_files', ['.env.local']);
 set('writable_dirs', ['var']);
 set('migrations_config', '');
 


recipe/symfony4.php
```diff
@@ -10,7 +10,7 @@
 require_once __DIR__ . '/common.php';
 
 set('shared_dirs', ['var/log', 'var/sessions']);
-set('shared_files', ['.env']);
+set('shared_files', ['.env.local']);
```
antonmedv: Will it break existing Symfony installations which are using this recipe?

Benjamin-Adrea: In facts, .env is versionned since Symfony 4.1.

Your configuration is ONLY for symfony 4.0 (https://symfony.com/doc/4.1/configuration/dot-env-changes.html) , and this version is not maintained anymore (https://symfony.com/roadmap)

So... yes it will maybe break existing symfony 4.0 apps, but symfony 4.0 isn't maintained so... I think you should adapt your config on the maintained version of symfony

As you want :)

antonmedv:
Can you also update CHANGELOG?

Run `php bin/changelog` script

Issue #374, opened 2015-06-23
Use * wildcard in shared files


marcosh:
This would be a nice feature to have. For example in Zend Framework 2 you have a lot of local (meaning git ignored) configuration files. It would be nice to be able to share them using a wildcard like `*.local.php`.

I realize that this goes againts the fact that the actual implementation creates the declared shared files if they are not there. Using a wildcard like the above mentioned one would create an actual `*.local.php` file.

Maybe we could have some optional configuration setting to say that the shared files needs to be created or not, or we could just simply search the file name and avoid the creation of the file if we find a * wildcard.

What do you think?

elfet:
This is a way of bugs %) 
Why you need this functionality?

marcosh:
As I said in Zend Framework applications you could have a lot of configuration files, all ending in `.local.php`. I would like to avoid the need to modify my `deploy.php` file everytime I add a new configuration file. It would be easier if all the `*.local.php` files were automatically shared every time.

On the other hand, I would have to create my config files in the `shared` directory, but anyhow I would have to put some content into them even I created them automatically.

Why do you think it could generate bugs?

hummer2k:
@marcosh maybe this will help: [environment-specific application config of zf2](http://framework.zend.com/manual/current/en/tutorials/config.advanced.html#environment-specific-application-configuration)

tautelis:
Our team is also in need of this wildcard feature. I would like to contribute to have it.

tforesti:
Hi, I think I need the same feature : I work on a multi-sites app and each site's admin can upload images for example. And these images are to be shared between releases.

Do you confirm that rigth now there is not feature which does this job and that this issue can does what I want ?

If yes do you have an idea on when this feature will be available ?

Thank you for your work.

PR #1895, opened 2019-08-06
Handle asterisk wildcard in shared files

tautelis:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | https://github.com/deployphp/deployer/issues/374

There are cases where sharing directory does not work and many dynamic files must be shared across releases. For example symfony configuration does not load properly from symlinked folders (here `config/dynamic` is a shared dir):
```
In FileLoader.php line 168:

The file "../config.yml" does not exist (in: /var/www/project/releases/10/src/../config/dynamic) in ../config.yml (which is being imported from "/var/www/project/releases/10/src/../config/dynamic/config_a6gr2.yml"). 


In FileLocator.php line 71:

The file "../config.yml" does not exist (in: /var/www/project/releases/10/src/../config/dynamic).
```

`/var/www/project/releases/10/src/../config/dynamic/../config.yml` - not found
`/var/www/project/releases/10/src/../config/config.yml` - exists

We are already using a custom task like that in our deployment and it works like a charm. I guess it might be useful for many fellow developers out there ;)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 ### Added
 - A task to cache the event listeners manifest in Laravel [#1893]
 - Added `check_remote_head` option, by setting this to true, deployer will avoid unnecessary new releases by checking the remote git HEAD without cloning the repo [#1755]
+- Handling asterisk wildcard in shared files [#374]
+
 
 ## v6.4.6
 [v6.4.5...v6.4.6](https://github.com/deployphp/deployer/compare/v6.4.5...v6.4.6)
@@ -608,4 +610,5 @@
 [#914]: https://github.com/deployphp/deployer/pull/914
 [#911]: https://github.com/deployphp/deployer/pull/911
 [#381]: https://github.com/deployphp/deployer/pull/381
+[#374]: https://github.com/deployphp/deployer/issues/374
 [#330]: https://github.com/deployphp/deployer/pull/330
--- a/recipe/deploy/shared.php
+++ b/recipe/deploy/shared.php
@@ -48,6 +48,9 @@ task('deploy:shared', function () {
     foreach (get('shared_files') as $file) {
         $dirname = dirname(parse($file));
 
+        // Check if file name has a wildcard * (e.g. config/parameters_*.yml)
+        $hasWildcard = strpos(str_replace($dirname, '', parse($file)), '*') !== false;
+
         // Create dir of shared file if not existing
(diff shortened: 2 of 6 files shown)


antonmedv:
Awesome!

antonmedv:
Please rebase.

antonmedv:
I got test error on my macos.
```
cp -nrv ./parameters_*.yml ./tmp-dist
./tmp-dist/parameters_dev.yml not overwritten
./tmp-dist/parameters_prod.yml not overwritten
```

I think we need to find a better interface for dealing with a wildcard. I'm going to revert this PR as it breaks my local development process. Will try to investigate it a little more.

tautelis:
Weird, does it have non-zero exit code? Since -n flag in cp is meant not to overwrite exiting files 🤔

antonmedv:
Yes, echo $0 == 1

tautelis:
Bloody hell, my bad :) I totally forgot that `-n` does not overwrite original but still returns non-zero response... we don't keep those files in repo so did not test this case. Anyhow there are two possible solutions here:

a) avoid non-zero response
```
run("cp -nrv {{release_path}}/$file $sharedPath/$dirname || true");
```
b) fetch matching files an do a check on each
```
$files = explode("\n", run("for f in {{release_path}}/$file ; do echo \$f ; done"));
foreach ($files as $f) {
    $sharedFile = $sharedPath . '/' .str_replace(parse('{{release_path}}'),'', $f);
    run("if [ ! -f $(echo $sharedFile) ]; then cp -rv $f $sharedFile; fi");
}
```

antonmedv:
What is faster?

tautelis:
First option would be faster for sure.

synio-wesley:
So, this has been reverted and won't be re-integrated? I'm missing this functionality. In one project with legacy code they are creating *.json and *.csv files in a subdirectory, but there's also versioned files in there, so I can't make the whole directory shared.

synio-wesley:
@tautelis @antonmedv Would this be an acceptable workaround?

```
tasks:
  share_wildcard_files:
    - run: cp -rf ~/admin/csv/*.csv {{release_path}}/admin/csv/
    - run: cp -rf ~/hooks/*.json {{release_path}}/hooks/

after:
  deploy:shared: share_wildcard_files
```

I'm trying to copy those files from the current release to the new release after the normal shared files have been created.

PR #1899, opened 2019-08-09
fixed invalid phpdoc

staabm:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  |  No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A 

>
>
>     php bin/changelog
>

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -602,7 +602,7 @@ function ask($message, $default = null, $suggestedChoices = null)
  * @param string[] $availableChoices
  * @param string|null $default
  * @param bool|false $multiselect
- * @return array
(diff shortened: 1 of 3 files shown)


src/functions.php
```diff
@@ -602,7 +602,7 @@ function ask($message, $default = null, $suggestedChoices = null)
  * @param string[] $availableChoices
  * @param string|null $default
  * @param bool|false $multiselect
- * @return array
+ * @return string|string[]
```
staabm: it can return string[] on multiselect questions but also string on non-multi

PR #1907, opened 2019-08-15
Update symfony4.php

Fahani:
The .env and .env.<environment> files should be committed to the shared repository because they are the same for all developers and machines. However, the env files ending in .local (.env.local and .env.<environment>.local) should not be committed because only you will use them. In fact, the .gitignore file that comes with Symfony prevents them from being committed.

https://symfony.com/doc/current/configuration.html#managing-multiple-env-files

| Q             | A
| ------------- | ---
| Bug fix?      | Yes

### DIFF
--- a/recipe/symfony4.php
+++ b/recipe/symfony4.php
@@ -10,7 +10,7 @@ namespace Deployer;
 require_once __DIR__ . '/common.php';
 
 set('shared_dirs', ['var/log', 'var/sessions']);
-set('shared_files', ['.env.local']);
+set('shared_files', ['.env']);
 set('writable_dirs', ['var']);
 set('migrations_config', '');
 


antonmedv:
Please, update changelog.

CharlyPoppins:
Hum are you sure of this.

.env contains the default values, it is versioned and should be in each release.

.env.local and .env.local.php should be in the shared_files, because it is NOT versioned and is an infrastructure related file.

TomaszGasior:
@antonmedv 

This PR merge should be reverted. Original author of PR does not understand how `shared_files` option works it seems to me.

Yep, it's true that ` .env`, `.env.test`, `.env.prod`, `.env.dev` files are "shared" and should be commited to the repository while `.env.local` and `.env.local.*` files should not (because they are user-specific) but the point is: to the repository, where application is developed. It's completely different from "shared" file in deployment meaning. In both "beta", "demo" or "prod" stages `.env` files without `local` in its names should be separate for each release, while `.env.local`, `.env.local.*` and `.env.local.*.php` should be shared because `*local*` files are specific to environment, where app is deployed, not to the release itself.

antonmedv:
Can you create pr with revert changes?

PR #1917, opened 2019-09-01
Dev/add custom key to fileloader

bambamboole:
| Q             | A
| ------------- | ---
| Bug fix?      |  No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No

This PR adds an optional second parameter to the inventory function. This parameter gets piped to the FileLoader's load method (optional as well). It is to specify a custom key under which the hosts definition can be found in the yaml file. 

Currently this is just for one level. Because this enough for my use case.


antonmedv:

Please, recreate your PR.

PR #1918, opened 2019-09-03
Symfony4: add a task to clear doctrine caches

CatoTH:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

`bin/console cache:clear` in Symfony does not automatically clear the caches of Doctrine. That is problematic if there is a migration, adding or removing fields to entities. If doctrine caches the relations between the tables and the entities, calling `clear-metadata` is necessary, as otherwise the app might break.

I only added the task and did not add it to the deploy chain (as `database:migrate` is also not); however, every time that `database:migrate` is called, it should be safe (and probably advisable) to call this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### Added
 - Added doc page with sample "real-world" Deployer script
+- Added `doctrine:cache:clear` task to symfony4 recipe, clearing the doctrine caches [#1918]
 
 ### Fixed
 - Parameters -f or --file now are accepted also without the equal sign [#1479]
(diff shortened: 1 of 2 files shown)


u-mulder:
`database:cache:clear` is too vague. What if in symfony project there's no `doctrine`? Therefore there're no cache.
I presume task should be renamed to `doctrine:cache:clear`.

CatoTH:
Ok, I changed the name as suggested.

fliespl:
I think it's worth including in all symfony recipes. But also worth mentioning that it should be run (i.e. when redis is involved) after symlink.

antonmedv:

Please, recreate your PR.

PR #1958, opened 2019-11-25
Fix the use of a branch or tag name containing a single quote

MoogyG:
If you are using a branch named with a single quote, you will get the next error `bash: line 1: unexpected EOF while looking for matching `''`

This pull request fixes it.


antonmedv:
Why do you add single quote in a first place?))

MoogyG:
In gitlab branch names are built from issues titles.

PR #1971, opened 2019-12-16
More Symfony 5 compatibility fixes

aboks:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  |  No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

I hate to bring the bad news here after your previous efforts to support Symfony 5, but there are still incompatibilities with symfony/process` 5.x in the Rsync- and SSH-client. This PR should fix them.

>
>
>     php bin/changelog
>

### DIFF
--- a/src/Ssh/Client.php
+++ b/src/Ssh/Client.php
@@ -68,7 +68,7 @@ class Client
             $command = escapeshellarg($command);
 
             $ssh = "ssh $sshArguments $host $command";
-            $process = new Process($ssh);
+            $process = $this->createProcess($ssh);
             $process
                 ->setTimeout($config['timeout'])
                 ->setTty(true)
@@ -89,7 +89,7 @@ class Client
             $ssh = "ssh $sshArguments $host $become '$shellCommand; printf \"[exit_code:%s]\" $?;'";
         }
 
-        $process = new Process($ssh);
+        $process = $this->createProcess($ssh);
         $process
             ->setInput($command)
             ->setTimeout($config['timeout']);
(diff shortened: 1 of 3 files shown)


loevgaard:
Could this be merged and released? :)

src/Utility/Rsync.php
```diff
@@ -43,7 +43,11 @@ public function call($hostname, $source, $destination, array $config = [])
 
         $this->pop->command($hostname, $rsync);
 
-        $process = new Process($rsync);
+        if (method_exists('Symfony\Component\Process\Process', 'fromShellCommandline')) {
```
sirian: Maybe replace with `if (method_exists(Process::class, 'fromShellCommandline'))`?

src/Ssh/Client.php
```diff
@@ -174,4 +174,13 @@ private function exec($command, &$exitCode = null)
         }
         return $output;
     }
+
+    private function createProcess($command)
+    {
+        if (method_exists('Symfony\Component\Process\Process', 'fromShellCommandline')) {
```
sirian: Maybe replace with `if (method_exists(Process::class, 'fromShellCommandline'))`?

src/Utility/Rsync.php
aboks: I agree that would be slightly cleaner. For now I used the exact same fix that was already present in e.g. `Deployer\Utility\ProcessRunner` for consistency. The best long-term solution would probably be to have a utility function somewhere in the deployer codebase to contain this compatibility layer.

PR #1975, opened 2019-12-18
[MOD] Constructor argument types of Process

alcoig:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes or No
| New feature?  | Yes or No
| BC breaks?    | **Yes** or No
| Deprecations? | Yes or No
| Fixed tickets | N/A or xx

First argument on Symfony\Component\Process\Process constructor must be of type array
>
>
>     php bin/changelog
>

### DIFF
--- a/src/Ssh/Client.php
+++ b/src/Ssh/Client.php
@@ -180,7 +180,7 @@ class Client
         if (method_exists('Symfony\Component\Process\Process', 'fromShellCommandline')) {
             return Process::fromShellCommandline($command);
         } else {
-            return new Process($command);
+            return new Process(explode(' ', $command));
         }
     }
 }


src/Ssh/Client.php
```diff
@@ -180,7 +180,7 @@ private function createProcess($command)
         if (method_exists('Symfony\Component\Process\Process', 'fromShellCommandline')) {
             return Process::fromShellCommandline($command);
         } else {
-            return new Process($command);
+            return new Process(explode(' ', $command));
```
antonmedv: I’m not sure this is proper way. There is a fromSgellCommand method.

maxlyaptsev: Are you sure there is only one compatibility issue? Could it be better to change the version for Process in the composer?

alcoig:
You are right, but if you do that, you don’t need the *else *bloc statement.

src/Ssh/Client.php
alcoig: You are right, but if you do that, you don´t need else bloc statement

Issue #1990, opened 2020-01-28
Deploy:check_remote task can't succeed


jonom:
1. The check_remote task checks for the existence of a `.dep/HEAD` file and throws an exception if not found, halting the deployment. That file doesn't get created/updated unless this check succeeds, so if you want to use this task you have to manually initialise that file currently.
2. Fetching the remote head is done from the target server, which will fail if you're using agent forwarding for auth (as noted previously [here](https://github.com/deployphp/deployer/issues/1755#issuecomment-520385551)).

Fix incoming.

Issue #1989, opened 2020-01-28
Silverstripe recipe asset path broken


jonom:
[This commit](https://github.com/deployphp/deployer/commit/d3a6ea7353ed014511c1ec3c821b3c977871764e) broke Silverstripe deployment asset path for me, and I think probably for anyone who doesn't commit their asset folder to their git repo (which is convention).

If you don't have an assets folder in your repo then when `{{shared_assets}}` is being evaluated [here](https://github.com/deployphp/deployer/blob/5095b5de166498aedb6ef7ffff98d7243fa036ea/recipe/deploy/shared.php#L38), the shared_assets callback returns null. This means that this command effectively runs `rm -rf {{release_path}}` which makes the deployment fail because the active release directory has been nuked.

Fix incoming.

PR #1991, opened 2020-01-28
Fixes for check_remote task and SS recipe

jonom:
I'd normally submit two separate PRs but noted the request to update the changelog so have bundled them together and pre-emptively added a 6.7.4 entry. Hope that's okay!

I documented `check_remote` task usage in deploy-strategies.md but wasn't sure if this was the right place. Let me know if you want that moved.

| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No (as far as I can tell)
| Deprecations? | No
| Fixed tickets | #1989 & #1990

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 Thank you for considering contributing to Deployer. Please make sure to read the following sections if you plan on submitting new issues or pull requests.
 
-## Bug
+## Reporting bugs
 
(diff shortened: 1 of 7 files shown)


jonom:
I got a bit carried away and worked on some contribution meta stuff too. Looks like the important tests are passing and Scrutinizer is failing because of a configuration problem?

antonmedv:
Yes, you are right. Looks like configuration issue.

PR #1998, opened 2020-02-04
Allow configuring binaries per host 

rvanlaak:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

Right now, it is not possible to modify the php / git / composer binary without having to adjust PATH or add an alias at the target host.

This PR adds functionality to override binaries per host for both `yaml` inventories as `php` configured hosts.

## Use cases

We want to test PHP7.4 on one of our edge / remote hosts without having to modify the defaults.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Added
 - Documented check_remote task usage
 - Speedup deploy:clear_paths
+- Allow configuring binaries per host (thanks @rvanlaak)
 
 ### Fixed
 - Fixed Silverstripe CMS recipe assets path [#1989]
--- a/docs/hosts.md
+++ b/docs/hosts.md
@@ -1,7 +1,7 @@
 # Hosts
 
 Defining a host in Deployer is necessary to deploy your application. It can be a remote machine, a local machine or Amazon EC2 instances.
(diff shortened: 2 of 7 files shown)


antonmedv:
Why not just set(‘bin/php’, ‘...’)? 
This is a common way of using configuration among all recipes. And common recipes already using bin/* prefix for binaries.

rvanlaak:
Because in our use case the binary has to be different per host

antonmedv:
Configuration can be different per host too.

rvanlaak:
Could you share a snippet on how to nicely do that so we can understand how to approach that?

antonmedv:
```php
host('...')
    ->set('bin/php', '/usr/local/bin/php7.4');
```

rvanlaak:
Thanks! How would this work in yaml?

antonmedv:
```yaml
  bin/php: /usr/local/bin/php7.4
```

rvanlaak:
Perfect, thank you!. It did exactly as expected. 

For some reason I did not think about this based on the documentation, so I'll file another PR to add this as example on the documentation.

PR #2000, opened 2020-02-07
Feature: added {{current}} variable

webdevium:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

Variable for current release folder name. For example, if you want **app** for current release folder you can run:
```php
set('current', 'app');
```

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -94,8 +94,9 @@ set('env', []); // Run command environment (for example, SYMFONY_ENV=prod)
 /**
  * Return current release path.
  */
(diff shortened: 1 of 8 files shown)


staabm:
Do you have a use-case, why/when you need to change this path?

I guess there will be a lot of 3rd party recipies which will take a lot of time to support this variable, and therefore it might not be really useful

Issue #1944, opened 2019-10-25
Symfony 4 bin/console not compatible with custom tasks for symfony 3


rabauss:
Unfortunately, the symfony 4 Beta is not fully compatible with symfony 3 custom tasks.
Before, `{{bin/php}}` was not included in `{{bin/console}}`.

This is is problem, if you've overwritten the `{{bin/console}}` and if you use `{{bin/php}} {{bin/console}}` in a custom task!

PR #2005, opened 2020-02-19
fix backward compatibility of bin/console for symfony4 recipe

rabauss:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | https://github.com/deployphp/deployer/issues/1944

Unfortunately, the symfony4 recipe is not compatible with tasks for symfony3 recipe, because the `bin/console` includes the `bin/php` and there are no `console_options` defined.
Therefore we changed the definition of `bin/console`!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Fixed Silverstripe CMS recipe assets path [#1989]
 - Fixed check_remote task errors [#1990]
 - Fixed check_remote task revision resolution [#1994]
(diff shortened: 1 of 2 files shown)


alexander-schranz:
@antonmedv this das also currently crash the sulu2 recipes. Is it possible to release this fix?

antonmedv:
Will release it soon!

PR #2030, opened 2020-04-22
Revert "Output with parallel task execution was garbled"

antonmedv:
Reverts deployphp/deployer#2016

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,10 +10,6 @@
 - Documented default_timeout config usage
 - Recipe for Joomla!
 
-### Changed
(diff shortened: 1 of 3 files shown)


antonmedv:
Nope

ebimmel:
Hi @antonmedv 
Just curious why this was reverted?

antonmedv:
I decided to do different thing what I think is much better for production and pipeline workflows as well as CLI developer experience. Also, this requires some work, also currently everything in the master branch. I hope to finish it someday (hope to find some free time for this). 

Actually you can try master branch right now.

PR #2031, opened 2020-04-23
Added support for composer v2 installed.json format + composer.phar fix

it-can:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A or xx

- Added support for composer v2 installed.json format 
- Only use bin/php on composer.phar


it-can:
@antonmedv can you take a look?

Issue #2035, opened 2020-04-26
When `symfony_env` is set to `dev`, composer `require-dev` dependencies aren't installed.


oleg-andreyev:
**Describe the bug**
When `symfony_env` is set to `dev`, composer `require-dev` dependencies aren't installed.

**To reproduce**
- Have something in require-dev in composer.json
- set symfony_env to dev
- run deployer, require-dev won't be installed

**Expected behavior**
require-dev dependencies are installed

brettins:
Hi, I'm on `6.8.0` with `symfony_dev` set to `dev` and  deployer is still trying to use no dev: 

```
/usr/local/bin/composer install --verbose --prefer-dist --no-progress --no-interaction --no-dev --optimize-autoloader --no-suggest
```

```
set('symfony_env', 'dev');
set('env', [
    'SYMFONY_ENV' => get('symfony_env'),
]);

```

Is there something else needed to enable this feature?

antonmedv:
Yes, update composer_option config.

brettins:
Wouldn't that just do the same thing as the fix itself and/or override it? The fix itself in the recipe is a `set('composer_options', ...)` set of code.

PR #2036, opened 2020-04-26
Composer options debug based on sf env

oleg-andreyev:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #2035

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 
 
+## master
+[v6.8.0...master](https://github.com/deployphp/deployer/compare/v6.8.0...master)
+
+### Fixed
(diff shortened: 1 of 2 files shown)


antonmedv:
Awesome! Also I'm thinking what else can be improved in Symfony recipe? As now its time to break change!

oleg-andreyev:
One of the things I was thinking recently is to utilize `git archive` instead of `git clone` not sure if it will improve anything yet but probably it could at least {{.git}} directory won't be created.

antonmedv:
Actually current `deploy:update_code` task is using `--dissociate` option which will reuse previously cloned release (!)
But you always can test. There is a new `--profile` option too.

PR #2047, opened 2020-05-07
fix composer to use defined php version if found in /usr/bin/composer

scuben:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

if compoesr is found /usr/bin/composer it wont add the defined php binary `{{bin/php}}` in front of it as it would if composer need to be downloaded.


scuben:
BTW: This PR originates from the 6.8.0 TAG. master seems to clearly be 7.0.0 and the branch 6.x is (way) behind.

antonmedv:
Yes, now work activity going on. Maybe you can rebased you pr to master?

scuben:
Would it be possible to release that fix within 6.8.x? I can then also port it to 7.0.0. WDYT?

PR #2062, opened 2020-05-19
Fix Property "seriesExecutor" does not exist.

RobinHoutevelts:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

I've started to create a POC for our deploying needs and noticed that the follow snippet threw an error in the master branch:

```php
task('build', function() {
    run('echo "hello world" > build.txt');
})->local();

task('deploy', function() {
    invoke('build');
});
```

### DIFF
--- a/src/Task/Task.php
+++ b/src/Task/Task.php
@@ -231,9 +231,14 @@ class Task
         return $this;
     }
 
+    /**
+     * @param string $selector
+     * @return Task
+     */
     public function select(string $selector)
     {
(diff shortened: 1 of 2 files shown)


RobinHoutevelts:
Seems I also inadvertently pushed some other changes as well.

If needed I can separate it into two PR's

PR #2072, opened 2020-05-28
Shopware6 installation routine

Webmasterei:
Due to the lack of ANY shopware6 installation routine, here's my little beginning of it. If you think it can be made better, I am very happy to have your ideas on board of this.

| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A 

> Do not forget to add notes about your changes to CHANGELOG.md. A command line tool has been included to make this easy, see [CONTRIBUTING.md] for details.

[CONTRIBUTING.md]: https://github.com/deployphp/deployer/blob/master/.github/CONTRIBUTING.md#update-the-changelog

### DIFF
--- a/recipe/shopware6.php
+++ b/recipe/shopware6.php
@@ -1,0 +1,59 @@
+<?php
+namespace Deployer;
+
+set('repository', 'git@github.com:shopware/production.git');
+
+set('shared_files', [
+    '.env'
+]);
+set('shared_dirs', [
+    'var',
+    'public/media',
+    'public/thumbnail'
+]);
+set('writable_dirs', [
+    'var',
+    'public/media',
+    'public/thumbnail'
+]);
+set('static_folders', []);
+set('writable_dirs', [
+    'public'
+]);
+
+task('sw:update_code', function(){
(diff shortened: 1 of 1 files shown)


recipe/shopware6.php
```diff
@@ -0,0 +1,59 @@
+<?php
+namespace Deployer;
+
+set('repository', 'git@github.com:shopware/production.git');
```
antonmedv, line 4: I believe this should be per project-based?

Webmasterei: The URL given is the standard-production template for shopware. It is of course required to fork that repository and use an custom URL within deploy.php
I thought it would be handy if it works even without it....

antonmedv: But it will be more likely overrided by init scriot.

Webmasterei: > But it will be more likely overrided by init scriot.

Right. And that would be totally ok. Anyway - I got you. So should I rather remove that?

PR #2081, opened 2020-06-08
Update symfony dependencies to latest long term support release

loevgaard:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

By only requiring the latest versions of these packages you are making it more difficult to require deployer because there will be conflicts.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,9 +32,9 @@
     "require": {
         "php": "^7.3",
         "pimple/pimple": "^3",
-        "symfony/console": "^5",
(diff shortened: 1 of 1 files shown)


antonmedv:
There is possibly to use deployer/dist or phar download instead. This will be only for development.

antonmedv:
I will add dist as default preferred usage installation method.

Issue #2012, opened 2020-03-11
Feature: Fix composer version between deploys


downace:
Currently `deploy:vendors` uses `bin/composer` option which downloads the latest Composer binary on every deploy. This can lead to unexpected breaking changes. For example, 1.10.0 started to output a lot of deprecation warnings in my case

There are lots of easy workarounds (override `bin/composer` option, use pre-installed Composer, etc.). But it would be nice to just add something like `set('composer_version', '1.9.3')`. Or just save composer binary somewhere between deploys (maybe in `.dep` dir)

asafov:
As i see - master branch first try to find composer binary.
```
set('bin/composer', function () {
    if (commandExist('composer')) {
        $composer = '{{bin/php}} ' . locateBinaryPath('composer');
    }

    if (empty($composer)) {
        run("cd {{release_path}} && curl -sS https://getcomposer.org/installer | {{bin/php}}");
        $composer = '{{bin/php}} {{release_path}}/composer.phar';
    }

    return $composer;
});
```
Try to set it:
`set('bin/composer', 'your_composer_path');`

asafov:
PR #2110 ready

antonmedv:
Fixed! Now composer will be saved!

PR #2110, opened 2020-06-24
Ability to select composer version

asafov:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #2012

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -114,13 +114,24 @@ set('bin/git', function () {
     return locateBinaryPath('git');
 });
 
+// Composer custom version
+set('composer_version', false);
+
 set('bin/composer', function () {
     if (commandExist('composer')) {
         $composer = '{{bin/php}} ' . locateBinaryPath('composer');
     }
 
     if (empty($composer)) {
-        run("cd {{release_path}} && curl -sS https://getcomposer.org/installer | {{bin/php}}");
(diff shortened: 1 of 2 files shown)


recipe/common.php
```diff
@@ -114,13 +114,24 @@
...
 set('bin/composer', function () {
     if (commandExist('composer')) {
         $composer = '{{bin/php}} ' . locateBinaryPath('composer');
     }
 
     if (empty($composer)) {
-        run("cd {{release_path}} && curl -sS https://getcomposer.org/installer | {{bin/php}}");
+        $installCommand = "cd {{release_path}} && curl -sS https://getcomposer.org/installer | {{bin/php}}";
+        
+        $composerVersion = get('composer_version', false);
```
antonmedv, line 128: What if instead of specifying version we will put composer in `.dep` dir? This way user will not need this hack to fix it's version.

asafov: .dep dir? release_path used. I only add version arg for pipeline.

vandres:
@antonmedv What is the solution for this atm? Have some failed builds, because Deployer is pulling Composer 2, but our projects still using Composer 1

antonmedv:
You can try to migrate to v7 (migration docs are coming). 

Simple mitigation is: install composer of desired version on host in PATH.

PR #2131, opened 2020-08-10
Feature/v6 exit code config option

edvordo:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

Recently I changed to a new server and in the process of setting it up I discovered the default shell on there is [fish-shell](https://github.com/fish-shell/fish-shell).
This one does not support `$?` as an exit code getter and [docs](https://fishshell.com/docs/current/tutorial.html#exit-status) suggest to use `$status`. When `$?` is used, it throws error, that stops the deploy script.

This setting allows to set the exit code getter to whatever I need via `set('ssh/exit_code_token', '$status')` and everything seems to work again.

---

Wasn't sure which branch to target this into. I made the branch from v6.8.0 tag/commit, so `6.x` seemed like a logical target.
Let me know if I should to change anything.


antonmedv:
There is no plans to support this feature.

edvordo:
Can you elaborate as to why not?
It's not a big change, it is fully backward compatible, doesn't add any overhead.

This is effectively forcing me to stop using your great tool I've enjoyed working with for over two years.

edvordo:
Okay, never mind, just found out I can change the shell do good old `/bin/bash`, everything works again.

Thanks for taking the time to look at my PR :)

PR #2143, opened 2020-08-29
Updated Shopware 6 recipe to reflect production template

peterjaap:
I was confused by production/development template terminology so I where I previously changed the recipe to work with the development template (which was wrong), I now fixed it to work with the production template

More info; https://www.p16r.nl/2020-08-28-shopware-6-development-versus-production-template/

### DIFF
--- a/recipe/shopware6.php
+++ b/recipe/shopware6.php
@@ -3,9 +3,12 @@ namespace Deployer;
 
 set('repository', 'git@github.com:shopware/production.git');
 
+set('release_name', function () {
+    return date('YmdHis');
+});
+
 set('shared_files', [
-    '.env',
-    '.psh.yaml.override'
+    '.env'
 ]);
 set('shared_dirs', [
     'custom/plugins',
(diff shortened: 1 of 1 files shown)


recipe/shopware6.php
```diff
@@ -3,9 +3,12 @@
 
 set('repository', 'git@github.com:shopware/production.git');
 
+set('release_name', function () {
+    return date('YmdHis');
```
antonmedv, line 7: Maybe this more user decided?

peterjaap: Might be, but I've never seen somebody *not* change it to the timestamp.

jonnitto: That's right, @peterjaap, but I wouldn't set this on an official recipe, as all the others don't change this also… BTW:  I've set it to this:
```php
set('release_name', static function (): string {
    return run('date +"%Y-%m-%d__%H-%M-%S"');
});
```

The benefit of that is that the release name is based on the server time and not on the local time.

PR #2148, opened 2020-09-02
Fix: Do not bother updating composer

localheinz:
This PR

* [x] stops updating `composer` twice

💁‍♂️ It's already updated on Travis. For reference, see

- https://travis-ci.org/github/deployphp/deployer/jobs/722653436#L192
- https://travis-ci.org/github/deployphp/deployer/jobs/722653436#L215

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,6 @@ addons:
       - acl
 
 install:
-  - composer self-update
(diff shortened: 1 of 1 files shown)


localheinz:
Thank you, @antonmedv!

PR #2156, opened 2020-09-04
fixes composer not using specified {{bin/php}}

riconeitzel:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

fixes missing {{bin/php}} in vendor task for composer calls.

forces composer calls to use the specified php binary

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
 - Recipe for Magento now supports locale configuration for `setup:static-content:deploy`. [#2040]
 - When symfony_env is set to dev, require-dev are not installed. [#2035]
 - Fixed exit status of rollback command when there are no releases to rollback to. [#2052]
-
+- composer install now respects {{bin/php}} config
 
 ## v6.8.0
 [v6.7.3...v6.8.0](https://github.com/deployphp/deployer/compare/v6.7.3...v6.8.0)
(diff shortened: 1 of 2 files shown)


antonmedv:
Please, update change log

riconeitzel:
@antonmedv done :)

antonmedv:
Somehow travis is now working.

riconeitzel:
"did you try to turn it off and on again?" 🤣 it's green 💚

loevgaard:
I think this is wrong. See https://github.com/deployphp/deployer/blob/8293eb931ed761fd8cb64f1c0ab20b510ed0c817/recipe/common.php#L111-L122

antonmedv:
Yes, @loevgaard is right

riconeitzel:
Then there’s another bug. In my environment composer ran into issues without my patch and was used with php instead of my set bin/php call. I can look into it again.

loevgaard:
Because you are running v6 probably :)

riconeitzel:
Downloaded it 5 days ago - no version constraint configured. I’ll check it!

riconeitzel:
Hi there …

I finally looked into it again.

I'm using 6.8.0 via composer - latest stable …

https://github.com/deployphp/deployer/blob/4e243a64ed61e779fbb31c5a74e258a8e52fdaff/recipe/common.php#L114-L125

so - the code @loevgaard mentions isn't part of the stable, yet 🤷🏼‍♂️ so in that case the only way is to use the beta … Am I right?

Rico

loevgaard:
Or copy the code and overwrite the `bin/composer` param in your local `deploy.php` file

PR #2159, opened 2020-09-09
Add sw:plugin:activate:all and sw:plugin:migrate:all to Shopware 6 recipe

peterjaap:
Do not update upgradeable plugins, it updates code too. What we actually want is to run the migration. But the database:migrate --all command only runs the migrations for the core. Added a task that runs migrations for all plugins too.

| Q             | A
| ------------- | ---
| Bug fix?      | Yes or No
| New feature?  | Yes or No
| BC breaks?    | Yes or No
| Deprecations? | Yes or No
| Fixed tickets | N/A or xx

> Do not forget to add notes about your changes to CHANGELOG.md. A command line tool has been included to make this easy, see [CONTRIBUTING.md] for details.

[CONTRIBUTING.md]: https://github.com/deployphp/deployer/blob/master/.github/CONTRIBUTING.md#update-the-changelog

### DIFF
--- a/recipe/shopware6.php
+++ b/recipe/shopware6.php
@@ -56,17 +56,19 @@ task('sw:cache:warmup', static function () {
 task('sw:database:migrate', static function () {
     run('cd {{release_path}} && bin/console database:migrate --all');
 });
-task('sw:plugins:installActivateUpgrade', static function () {
+task('sw:plugin:refresh', function (){
+    run('cd {{release_path}} && bin/console plugin:refresh');
+});
+task('sw:plugin:activate:all', static function () {
+    task('sw:plugin:refresh');
     $plugins = explode("\n", run('cd {{release_path}} && bin/console plugin:list'));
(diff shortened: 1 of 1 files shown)


recipe/shopware6.php
```diff
@@ -88,9 +90,37 @@
         if ($installed === 'No' || $active === 'No') {
             run("cd {{release_path}} && bin/console plugin:install --activate $plugin");
         }
+    }
+});
+task('sw:plugin:migrate:all', static function(){
+    $plugins = explode("\n", run('cd {{release_path}} && bin/console plugin:list'));
```
Schrank, line 96: We should extract the listing of the plugins and maybe even make a json out of it, so we can reuse it later?

peterjaap: Yeah I thought of that, but it needs to update in the meantime so we now which plugins are actually installed & active.

PR #2161, opened 2020-09-10
Make deploy:check_remote command less vague in its output

### LINKED ISSUE #2160: [Enhancement] Make deploy:check_remote more specific as to which host is up-to-date
When using the `deploy:check_remote` step, to prevent multiple releases of the same revision, the current output simply says:
```
➤ Executing task deploy:check_remote

In check_remote.php line 57:
                       
  Already up-to-date.  
```
After showing that message, the deploy procedure is being stopped.

This can be too vague in a situation where stage has multiple hosts - deployer does not convey any information about the host that is up-to-date. I can imagine a scenario, where for some reason, one of the hosts is on newest deployment and the rest isn't. Having some sort of information about which host to skip would be beneficial here.

I would propose to include host information in the "is up-to-date" message, for instance:
```
   YOUR-HOST (your.host.domain) is already up-to-date
```
where `YOUR-HOST` is a _host_ and `your.host.domain` is a _hostname_.

Alternatively, the `deploy:check_remote` could mark specific host as being up-to-date and would skip next steps for this configuration. Deploy for other hosts would still continue. I don't know however if this is currently possible to implement, without major overhaul.

bsacharski:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #2160 

The `deploy:check_remote` command would return very vague information about the target machine being up-to-date:

```
➤ Executing task deploy:check_remote

In check_remote.php line 57:

  Already up-to-date.
```

When deploying to multiple hosts in a single stage, this message can be too vague, because it doesn't specify which host is preventing the deploy.

This commit changes that behavior, to pull the host from the Context and present that information in the message:

```
➤ Executing task deploy:check_remote

In deploy_steps.php line 63:

  HOSTNAME (real.hostname.domain) is already up-to-date.
```

Fixes #2160

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@
 - Symfony 5 recipe.
 - Command for checking if a deploy is unlocked. [#2150] [#2150]
 
+### Changed
+- Make deploy:check_remote command less vague in its output. [#2160]
+
 ### Fixed
 - Normalize CRLF to LF new line endings. [#2111]
 - Recipe for Magento now supports locale configuration for `setup:static-content:deploy`. [#2040]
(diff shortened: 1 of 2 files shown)


antonmedv:
Now it looks like this:
```
dep deploy:check_remote prod
task deploy:check_remote
[prod]  Deployer\Exception\GracefulShutdownException  in check_remote.php on line 56:
[prod]
[prod]   Already up-to-date.
[prod]
```

Issue #2170, opened 2020-09-24
Remote binary paths may overwrite local binary paths (and vice versa)


bsacharski:
I have recently discovered bug where the values like `bin/git` (defined in `common.php` recipe) would point to wrong binary.
This was caused by the local and remote machines having different file structure:
* on remote, the git binary was in `/bin/git`
* on local, the git binary was in `/usr/bin/git`

During the deploy procedure the `get('bin/git')` would've been called twice:
* first time on remote, which caused the standard closure with `return locateBinaryPath('git');` to be called.
* second time on local machine.

Due to existing `get` implementation, the first time that `bin/git` was accessed, the closure was executed and its result has been saved in place of that closure.
As such, the second time the `bin/git` has been accessed, the "cached" value has been returned immediately.

tl;dr; using the `get(),` combined with closure that contains call to `locateBinaryPath()`, will result in the value being set to the binary path of the first host that the method was invoked on.

**To reproduce**
1. Have a local environment with the git executable in the `/usr/bin/git`
2. Have a remote environment with the git executable in the `/bin/git`
3. Have a deploy.php that uses:
    ```php
    set('bin/git', function () {
        return locateBinaryPath('git');
    });
4. Run `get('bin/git')` against remote host
5. Try to exec git using executable from `get('bin/git')` on local machine

**Expected behavior**
The paths to binaries returned by the `get()`should always reflect real path to the binary on the machine that the `get` is being invoked on.

IMO this could be done by either:
* having `get` know its current context (so it would differentiate between different hosts)
* having an option to mark the closure passed as the second argument of `set` method as "protected", meaning it would never be replaced with the result of invocation (simpler, but would impact the performance).

**Environment**
* Deployer version: 6.8
* PHP version: 7.3
* Deployment target(s) OS:
  remote was running on CentOS
  local was running in alpine-based docker container

antonmedv:
> tl;dr; using the get(), combined with closure that contains call to locateBinaryPath(), will result in the value being set to the binary path of the first host that the method was invoked on.

This is not true. This closure/memorization happens per host. So different hosts and localhost will store different results. 

Please provide a minimal example to show your case. Maybe something else is happening there.

bsacharski:
> This is not true. This closure/memorization happens per host. So different hosts and localhost will store different results.

You're right and I feel stupid now.

I've been using the `check_remote` task provided by the version `6.8` - here's the relevant snippet:

```php
...
[...]

PR #2172, opened 2020-09-26
[issue-2170] Modified runLocally function to accept closure

bsacharski:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | Yes
| BC breaks?    |  No
| Deprecations? |  No
| Fixed tickets | #2170 

As noted in the issue #2170, the logic in the `recipe/check_remote.php` would try to get the binary path of local and remote git executable.

However, due to specific implementation, the `get('bin/git')` would have been called twice in the context of remote host, instead of invoking it once on remote and once on local machine.

The `runLocally` function has thus been reworked, to accept Closure, which will be invoked in the context of the localhost machine.
Once a Closure has been passed, the function will push the Localhost to the Context, invoke the closure, store the stdout (if available), pop the Localhost from the Context and finally return the stdout.

For compatibility reasons, the function also accepts the string argument, however it is wrapped in a closure and once again passed to the `runLocally` function.

The `recipe/check_remote.php` file has been reworked, to utilize the Closure, to make sure that the `get('bin/git')` is being called in the Localhost context.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - ISPManager recipe and docs.
 - Symfony 5 recipe.
 - Command for checking if a deploy is unlocked. [#2150] [#2150]
+- `runLocally` function now accepts string or a Closure as first parameter. [#2170]
 
 ### Fixed
 - Normalize CRLF to LF new line endings. [#2111]
@@ -17,6 +18,7 @@
 - When symfony_env is set to dev, require-dev are not installed. [#2035]
 - Fixed exit status of rollback command when there are no releases to rollback to. [#2052]
 - When the second parameter $options passed to run() and runLocally(), use it to overwrite default env config. [#2165]
+- Fixed `recipe/check_remote.php` calling `get('bin/php)` in remote host context twice. [#2170]
 
 
 ## v6.8.0
@@ -583,6 +585,7 @@
 - Fixed remove of shared dir on first deploy.
 
 
+[#2170]: https://github.com/deployphp/deployer/issues/2170
 [#2165]: https://github.com/deployphp/deployer/issues/2165
 [#2150]: https://github.com/deployphp/deployer/issues/2150
 [#2111]: https://github.com/deployphp/deployer/pull/2111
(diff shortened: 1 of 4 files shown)


antonmedv:
This is an interesting idea, but it repeats task functionality. Also if _check_remote_ will be running on a few hosts run part of code will be executed multiple times.
Task can be executed in localhost context:
```php
task('deploy:check_local', function () {
        $remoteLs = run(sprintf("%s ls-remote $opt $repository $ref", get('bin/git')));
})->local();

**Comment by bsacharski** [comment]
> This is an interesting idea, but it repeats task functionality.
I'm not sure whether it repeats the functionality. It can achieve the same thing, but IMO represents different intentions:
- `task()->local()` represents a task that is to be executed on a local machine
- `runLocally()` represents a command that should explicitly be executed on a local machine

With `runLocally` we can mix contexts when needed, to have a task executed mostly on a remote host, with only one or two commands that need to be executed on a localhost.

> Also if _check_remote_ will be running on a few hosts run part of code will be executed multiple times.
Technically, we could add a check that would prevent multiple calls, but I would concentrate on that once we discover that performance really is an issue.

> Task can be executed in localhost context:
> 
> ```
> task('deploy:check_local', function () {
>         $remoteLs = run(sprintf("%s ls-remote $opt $repository $ref", get('bin/git')));
> })->local();
> ```

To me it looks like it splits the `deploy:check_remote` step too much - it would probably end up with:

```php
task('deploy:check_version', [
  'deploy:check_local', // this is executed only once on local machine, sets the 'targetRevision` value
  'deploy:check_remote', // this has to grab a value set by the `check_local` and compare it against remote version
]);
```
Not sure if such separation is desired.

antonmedv:
I like the idea in general, but for me, it seems a more complex thing to add. We have local tasks and _on host_ functionality. 
What about thinking in this direction (it should work from my perspective):
```php
on(localhost(), function () {
    $remoteLs = run(sprintf("%s ls-remote $opt $repository $ref", get('bin/git')));
});
```

bsacharski:
Looks nice. Do you want me to create a new PR with `check_remote` reworked to utilize `on()`?

antonmedv:
Yes, it will be cool. Also, check if this functionality actually work))

PR #2175, opened 2020-10-01
[issue-2170] Use on(localhost(), ...) instead of runLocally in check_remote.php

bsacharski:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #2170 

Reworked the `deploy:check_remote` step, to use the `on(localhost(), ...)`, instead of relying on `runLocally`.

This has been changed, to make sure that the all functions that are being executed and are related to "local" part of the logic, are ran on the local host.

Previously in `runLocally(sprintf("%s ls-remote $opt $repository $ref", get('bin/git')));` the `get('bin/git')` would've been executed on the remote host, which could have resulted in an unexpected behaviour.

By closing all of the instructions, in the closure that is executed by the `on()` function, we are sure that the context is set to the right host.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - When symfony_env is set to dev, require-dev are not installed. [#2035]
 - Fixed exit status of rollback command when there are no releases to rollback to. [#2052]
 - When the second parameter $options passed to run() and runLocally(), use it to overwrite default env config. [#2165]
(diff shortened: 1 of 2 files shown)


bsacharski:
I've tested copied this task to a separate file, then required the local copy in one of my projects and it was working as expected with the `6.8` version of the Deployer.

Would appreciate someone to crosscheck though.

PR #2180, opened 2020-10-02
[issue-2176] Added `escape`/`unescape` functions

### LINKED ISSUE #2176: `escape` method for `run` command
**Is your feature request related to a problem? Please describe.**
we use the `mysql` command line, invoked via deployer `run()` to pass sql data between servers thru deployers ssh tunnel.

the sql command used for the commands can contain any text, e.g. even a `{{abc}}` which deployer tries to parse as a variable.

**Describe the solution you'd like**
I would like some kind of `escape` function which allows me to mask a string which may contain `{{xx}}` and returns me a string which can be passed thru `run()` without beeing interpreted as deployer variables.

in the end my php api looks like 
```
$data = run('mysql --silent --raw --skip-column-names --execute '.AsyncProcess::escapeArgument($query).' '.$dbCredentials->asCliArgs());
```

in the recipe

especially suprising is the error which gets reported, because it mentions a configuration variable not beeing defined and it took me a while to find out that my data contains stuff deployer is trying to interpret

```
In Configuration.php line 89:

  [Deployer\Exception\ConfigurationException]
  Configuration parameter `xx` does not exist.

Exception trace:
  at /home/complexit/.composer/vendor/deployer/deployer/src/Configuration/Configuration.php:89
 Deployer\Configuration\Configuration->get() at /home/complexit/.composer/vendor/deployer/deployer/src/Configuration/Configuration.php:130
 Deployer\Configuration\Configuration->parseCallback() at n/a:n/a
 preg_replace_callback() at /home/complexit/.composer/vendor/deployer/deployer/src/Configuration/Configuration.php:116
 Deployer\Configuration\Configuration->parse() at /home/complexit/.composer/vendor/deployer/deployer/src/functions.php:761
 Deployer\parse() at /home/complexit/.composer/vendor/deployer/deployer/src/functions.php:494
```

bsacharski:
- [x] New feature: functions for escaping/unescaping values wrapped in `{{}}`
- [ ] BC breaks
- [ ] Deprecations
- [x] Tests added
- [x] Changelog updated

This commit adds two new functions:
* `escape(string): string`
* `unescape(string): string`

Once the value has been processed by the `escape` function, then the values wrapped in `{{}}` like `{{foo}}` will not be parsed by the `parse` function.

The `escape` function works by replacing the `{{` with `\\{\\{` and `}}` with `\\}\\}`.
The `unescape` function reverses the process.

The `parse` function has been updated to run the `unescape` afterparsing has been completed, to ensure that the commands are being executed normally.

Fixes #2176

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - ISPManager recipe and docs.
 - Symfony 5 recipe.
 - Command for checking if a deploy is unlocked. [#2150] [#2150]
+- Added `escape` and `unescape` method to ignore certain values from being parsing. [#2176]
 
 ### Changed
 - `within` passess through `$callback` return value. [#2178]
@@ -590,6 +591,7 @@
 
 [#2181]: https://github.com/deployphp/deployer/issues/2181
 [#2178]: https://github.com/deployphp/deployer/issues/2178
+[#2176]: https://github.com/deployphp/deployer/issues/2176
 [#2170]: https://github.com/deployphp/deployer/issues/2170
 [#2165]: https://github.com/deployphp/deployer/issues/2165
 [#2150]: https://github.com/deployphp/deployer/issues/2150
--- a/src/Configuration/Configuration.php
+++ b/src/Configuration/Configuration.php
@@ -13,6 +13,7 @@ use function Deployer\get;
 use function Deployer\Support\array_merge_alternate;
 use function Deployer\Support\is_closure;
 use function Deployer\Support\normalize_line_endings;
+use function Deployer\unescape;
 
 class Configuration implements \ArrayAccess
 {
@@ -107,7 +108,8 @@ class Configuration implements \ArrayAccess
     {
         if (is_string($value)) {
             $normalizedValue = normalize_line_endings($value);
-            return preg_replace_callback('/\{\{\s*([\w\.\/-]+)\s*\}\}/', [$this, 'parseCallback'], $normalizedValue);
+            $parsed = preg_replace_callback('/\{\{\s*([\w\.\/-]+)\s*\}\}/', [$this, 'parseCallback'], $normalizedValue);
+            return unescape($parsed);
         }
 
         return $value;
--- a/src/functions.php
+++ b/src/functions.php
@@ -551,6 +551,30 @@ function parse($value)
 }
 
 /**
+ * Escapes value so it won't be parsed by @see parse()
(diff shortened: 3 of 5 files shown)


clxmstaab (approved):
this looks really good, thx

antonmedv:
So `unescape` will be run on each command? Hmmm. What are consecutions? Is there a way it will break something? 
Maybe we should think a little about placeholders? 

Now there is one special placeholder secret (which will not be printed to stdout nor logs):

```php
run("fjoy -p '%secret%' -fk", ['secret' => getenv('CI_SECRET')]);
```

What about adding something similar?

```php
run('mysql --silent --raw --skip-column-names --execute %query%', ['vars' => [
    'query' => $query,
]]);
```

Also, I'm concerned with the name escape. It may confuse people what this is for escaping CLI arguments (deals with quotas, but it does not).

bsacharski:
This might be even more complicated than I've previously thought. 

The `{{foo}}` is being replaced everytime the `parse` method is being called, meaning it'll not only be executed as a part of `run` but also as a part of `get`.

Maybe the best approach would be simply to do nothing in `\Deployer\Configuration\Configuration::parseCallback` when the "match" is not defined in configuration values?

antonmedv:
So what do you think about placeholders?

bsacharski:
Placeholder will work if the value is being set explicitly, like:

```php
$query = 'select * from {{foo}}';
run('mysql --silent --raw --skip-column-names --execute %query%', ['vars' => [
    'query' => $query,
]]);
```

if however the code is like:

```php
set('foo', 'select * from {{foo}}');
run('mysql --silent --raw --skip-column-names --execute %query%', ['vars' => [
    'query' => get('foo'),
]]);
```
then the code will try to parse that `{{foo}}`.

If we assume that all we want to handle here is `run` then I guess I could proceed with the different MR ;)

antonmedv:
```php
set('foo', 'select * from {{foo}}');
```

What is the point if this?) From my understanding case is next: a user has some $variable with `{{}}` in it. What to pass it to run. Placeholder will work in this case.

bsacharski:
There is no point ;) 

The problem is that we do not know where the value that will be used with `run` and placeholders will be coming from. If it comes from `get` then it'll still be parsed. But I guess that's a problem that has not yet happened ;) It can be dealt with once (if ever) it appears ;)

I'll get to the new MR then ;)

PR #2185, opened 2020-10-03
[issue-2178] within() passes through $callback return value

### LINKED ISSUE #2178: within() should pass thru the $callback return
**Is your feature request related to a problem? Please describe.**
Currently it's unable to wrap run('command') call into within() helper and get command output back.

**Describe the solution you'd like**
The return of $callback() must be returned as result of within().

**Describe alternatives you've considered**
Current solution is just copy/paste the within() sources into client code and replace $callback to `$output = run('command');`

bsacharski:
- [ ] Bug fix
- [x] New feature #2178 
- [ ] BC breaks
- [ ] Deprecations
- [x] Tests added
- [x] Changelog updated

The `within()` function has been modified to pass-through the return value of the `$callback` function.

It is now possible to write code like:

```php
$out = within('/foo', function () {
  return 'bar';
});
```

The `$out` variable will now contain the value `'bar'`.

If the callback function returns nothing, then the return of `within()` call will be `null`.

Fixes #2178

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -262,13 +262,14 @@ function cd($path)
  *
  * @param string $path
  * @param callable $callback
+ * @return mixed|void Return value of the $callback function or void if callback doesn't return anything
  */
 function within($path, $callback)
 {
     $lastWorkingPath = get('working_path', '');
     try {
         set('working_path', parse($path));
-        $callback();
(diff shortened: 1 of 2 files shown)


antonmedv:
Awesome!

Issue #2187, opened 2020-10-05
Add type declarations to the methods


bsacharski:
To improve the quality of code, the methods and functions should use type declarations to define what types of arguments are accepted and what types of values are being returned.

Native type declarations should be used where possible with the fallback to comment-based annotations.

TODO:

- [x] `src/Collection/`
- [x] `src/Component/`
- [x] `src/Configuration/`
- [ ] `src/Console/`
- [x] `src/Documentation/`
- [x] `src/Exception/`
- [x] `src/Executor/`
- [x] `src/Host/`
- [x] `src/Logger/`
- [ ] `src/Selector/`
- [ ] `src/Support/`
- [x] `src/Task/`
- [x] `src/Utility/`
- [ ] `src/Deployer.php`
- [x] `src/functions.php`

antonmedv:
What about adding all files to phpcs lint but with weaker checks? And steady move them to more strict grouping?

This way we can be sure new files will not be missed.

bsacharski:
Type hints in `functions.php` added in https://github.com/deployphp/deployer/pull/2210

antonmedv:
ping @bsacharski 
Should we wait for everything is done bore final release v7?

bsacharski:
@antonmedv nah, I think it's a thing that's nice to have, but I don't think it should impact release ;)
To be honest, I don't have that much time to work on this at the moment (as there were little to no commits from my side) and I don't want to imact the release in any way.

antonmedv:
Only a few left. Will add it myself.

PR #2188, opened 2020-10-05
[issue-2187] Type declarations for src/Configuration

bsacharski:
- [x] Improvement for #2187 (part of work)
- [ ] New feature
- [ ] BC breaks
- [ ] Deprecations
- [ ] Tests added
- [x] Changelog updated

Refactored the code in `src/Configuration` to include type declarations for method arguments and returned values.

To make it easier to track what's left to do and to have a tool that prevents "regressions" in the future, PHP_CodeSniffer and Slevomat Coding Standard have been added.

Next, a `phpcs.xml.dist` file has been created and configured to enforce type hints for methods. For the time being, only files in the src/Configuration directory are now being watched.

To make it simpler to execute the checks, `lint` script has been added to `composer.json` to allow running checks with `composer lint`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,8 @@
         ]
     },
     "scripts": {
-        "test": "phpunit"
+        "test": "phpunit",
+        "lint": "phpcs"
     },
     "bin": [
         "bin/dep"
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -39,7 +40,9 @@
         "symfony/yaml": "^5"
     },
     "require-dev": {
-        "phpunit/phpunit": "^8.3"
+        "phpunit/phpunit": "^8.3",
+        "slevomat/coding-standard": "~6.0",
```
antonmedv, line 44: What about using `^` as in phpunit?

bsacharski: Sure. Both `slevomat/coding-standard` and `squizlabs/php_codesniffer` are now using `^`-based selector.

PR #2193, opened 2020-10-08
[#2192] Support for placeholders in run and runLocally

### LINKED ISSUE #2192: Add support for placeholders in run and runLocally methods
As [mentioned](https://github.com/deployphp/deployer/pull/2180#issuecomment-704116650) by @antonmedv in #2178, it would be good to have support for placeholder that will not be going through `parse` method, so their value will be left intact.

For instance:

```php
run('mysql --silent --raw --skip-column-names --execute %query%', ['vars' => [
    'query' => $query,
]]);
```

In the example above, the `$query` value will replace the `%query%` placeholder. This might be specially useful in the case of values, that contain syntax similar to config-value references, like `{{foo}}`.

bsacharski:
- [x] New feature?
- [ ] BC breaks
- [ ] Deprecations
- [x] Tests added
- [x] Changelog updated

Added support for placeholder in the `run` and `runLocally` functions.
By passing the `vars` key with an associative array to the second
argument of `run` and `runLocally`, it is possible to use placeholders
in the command, that will be used directly in the executed command.

The placeholders should be wrapped in the command with `%NAME%`, for
instance `%foo%`. The `vars` array could be `[ 'foo' => '{{bar}}']`.

Sample:

```php
runLocally("echo '%foo%'", ['vars' => ['foo' => '{{bar}}']]);
```

will output:

```plain
{{bar}}
```

Fixes #2192

### DIFF
--- a/src/Component/ProcessRunner/ProcessRunner.php
+++ b/src/Component/ProcessRunner/ProcessRunner.php
@@ -43,6 +43,7 @@ class ProcessRunner
             'idle_timeout' => null,
             'cwd' => defined('DEPLOYER_ROOT') ? DEPLOYER_ROOT : null,
             'tty' => false,
+            'vars' => [],
         ];
         $config = array_merge($defaults, $config);
 
@@ -53,6 +54,8 @@ class ProcessRunner
             $this->logger->printBuffer($host, $type, $buffer);
             $terminalOutput($type, $buffer);
         };
+
+        $command = $this->replacePlaceholders($command, $config['vars']);
 
         $command = str_replace('%secret%', $config['secret'] ?? '', $command);
         $command = str_replace('%sudo_pass%', $config['sudo_pass'] ?? '', $command);
@@ -78,5 +81,14 @@ class ProcessRunner
                 $process->getErrorOutput()
             );
         }
+    }
+
+    private function replacePlaceholders(string $command, array $variables): string
+    {
+        foreach ($variables as $placeholder => $replacement) {
+            $command = str_replace("%$placeholder%", $replacement, $command);
+        }
+
+        return $command;
     }
 }
--- a/src/Component/Ssh/Client.php
+++ b/src/Component/Ssh/Client.php
@@ -35,6 +35,7 @@ class Client
         $defaults = [
             'timeout' => $host->get('default_timeout', 300),
             'idle_timeout' => null,
+            'vars' => [],
         ];
 
         $config = array_merge($defaults, $config);
@@ -69,6 +70,8 @@ class Client
             $this->logger->printBuffer($host, $type, $buffer);
             $terminalOutput($type, $buffer);
         };
+
+        $command = $this->replacePlaceholders($command, $config['vars']);
 
         $command = str_replace('%secret%', $config['secret'] ?? '', $command);
         $command = str_replace('%sudo_pass%', $config['sudo_pass'] ?? '', $command);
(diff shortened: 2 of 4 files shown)


src/Component/ProcessRunner/ProcessRunner.php
```diff
@@ -54,6 +55,8 @@ public function run(Host $host, string $command, array $config = []): string
             $terminalOutput($type, $buffer);
         };
 
+        $command = $this->replacePlaceholders($command, $config['vars']);
```
antonmedv, line 58: Awesome PR 👏
But there is support in `run()`?

bsacharski: Crap, it's missing. Will follow up with another commit.

I've tried to add unit tests, however it seems that `on(localhost(), function () { run() })` is not working in unit tests at the moment.

bsacharski: Unfortunately, I don't know how to add unit tests for that, as the `on()` will not work, and setting localhost in the context results in the logic to use the `ProcessRunner` instead of `Ssh/Client`

The implementation has been added in the c8657c425c1c1db8ecd12265abeac9e78912836e

src/Component/Ssh/Client.php
```diff
@@ -70,6 +71,8 @@ public function run(Host $host, string $command, array $config = [])
             $terminalOutput($type, $buffer);
         };
 
+        $command = $this->replacePlaceholders($command, $config['vars']);
```
antonmedv, line 74: It will be cool to have tests here as well. Please git pull and see test/joy/OnceTest for example.

src/Component/ProcessRunner/ProcessRunner.php
```diff
@@ -43,6 +43,7 @@ public function run(Host $host, string $command, array $config = []): string
             'idle_timeout' => null,
             'cwd' => defined('DEPLOYER_ROOT') ? DEPLOYER_ROOT : null,
             'tty' => false,
+            'vars' => [],
         ];
```
clxmstaab, line 47: wdyt about adding a phpdoc describing the $config array structure, so phpstan/psalm based workflows can validate it and catch errors eary?

something like
`@psalm-param array{timeout?: int, idle_timeout?: int|null, cwd?: string, tty?: bool, vars?: array<string, string>}`

?

src/Component/Ssh/Client.php
bsacharski: I think the problem is that even with using `ApplicationTester`, the tests are still being executed using `\ProcessRunner` instead of the `Ssh\Client`.

To test it with unit tests I would either:
* have to rework the `Ssh\Client` class, replace the `Process::fromShellCommandline` with some kind of "injectable" factory - this would allow to mock the Process class and inspect what the Client is passing there.
* create a very small environment that the tests would run in (based on docker-compose for instance).

What do you think?

src/Component/ProcessRunner/ProcessRunner.php
bsacharski: That's up to @antonmedv ;) At this point I personally would like to concentrate on this ticket and #2187 

If the project was to use the psalm annotations, then I guess it would be good to also include it in the pipeline and maybe have some kind of check, that makes sure that new commits include them

src/Component/Ssh/Client.php
antonmedv: Yes, you are right.

> onvert it to "end-to-end" test by creating a very small environment that this test would run in (based on docker-compose for instance).

Adding end-to-end will really awesome.

PS Take a look at f429c750bc95dd153228257a3967039661b0bb40

src/Component/ProcessRunner/ProcessRunner.php
clxmstaab: added a phpstan github action with https://github.com/deployphp/deployer/pull/2218

PR #2194, opened 2020-10-08
[issue-2187] Type declarations for src/Collection

bsacharski:
- [X] Improvement for #2187 (part of work)
- [ ] New feature
- [ ] BC breaks
- [ ] Deprecations
- [ ] Tests added
- [X] Changelog updated

Refactored the code in `src/Collection` to include type declarations for method arguments and returned values.
Also updated the code in `src/Configuration/Configuration.php` to add missing type declaration for `ownValues`

### DIFF
--- a/src/Collection/Collection.php
+++ b/src/Collection/Collection.php
@@ -14,11 +14,15 @@ class Collection implements Countable, IteratorAggregate
 {
     protected $values = [];
 
-    public function all()
+    public function all(): array
     {
         return $this->values;
     }
 
+    /**
+     * @return mixed|void
+     * @phpcsSuppress SlevomatCodingStandard.TypeHints.ReturnTypeHint.MissingNativeTypeHint
+     */
     public function get(string $name)
     {
         if ($this->has($name)) {
@@ -33,6 +37,10 @@ class Collection implements Countable, IteratorAggregate
         return array_key_exists($name, $this->values);
     }
 
+    /**
+     * @param mixed $object
+     * @phpcsSuppress SlevomatCodingStandard.TypeHints.ReturnTypeHint.MissingNativeTypeHint
+     */
     public function set(string $name, $object)
     {
         $this->values[$name] = $object;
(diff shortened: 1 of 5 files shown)


src/Collection/Collection.php
```diff
@@ -14,11 +14,15 @@ class Collection implements Countable, IteratorAggregate
...
 
-    public function all()
+    public function all(): array
     {
         return $this->values;
     }
 
+    /**
+     * @return mixed|void
+     * @phpcsSuppress SlevomatCodingStandard.TypeHints.ReturnTypeHint.MissingNativeTypeHint
```
antonmedv, line 24: Is there a way to opt-out from this check? This looks like very verbose stuff.

bsacharski: Given [docs](https://github.com/slevomat/coding-standard#slevomatcodingstandardtypehintsreturntypehint-) none that I am aware of.

antonmedv:
Please, rebase.

src/Collection/Collection.php
antonmedv: I found a way. Please, drop those lines.

bsacharski: Will do. Can you share details how you're doing that?

bsacharski: Ok, I see how you've added configuration setting to phpcs.xml. Didn't know about that one :+1:

src/Task/TaskCollection.php
```diff
@@ -15,7 +15,7 @@
  */
 class TaskCollection extends Collection
 {
-    protected function throwNotFound(string $name)
+    protected function throwNotFound(string $name): void
```
antonmedv, line 18: So, is this `: void` parts also necessary? 
We have lots and lots of anon function in recipes, do we really want to add it there? What benefits will it provide? 
Should we omit recipes/ all together?

bsacharski: I would skip recipes, however IMO the core library should contain as much of native and phpDoc-based type definitions as possible.

As for void being hinted, IMO yes. This having explicit void means that function should never, ever, return anything. This in turn helps with finding bugs even on static analysis level, as PHPStan will report an issue if a piece of code tries to use a void result.

PR #2196, opened 2020-10-10
Validate configurations

### LINKED ISSUE #976: Identify user mistakes in config
| Q                 | A
| ----------------- | ---
| Issue Type        | Feature Request
| Deployer Version  | next

Implement internal bookkeeping to help identify user mistakes like spelling errors or unused variables that may lead to unexpected behavior.

```php
set('keep_relaeses', 5);
```

Shows an warning:

```
Unknown config `keep_relaeses`. Did you mean `keep_releases`?
```

brunofontes:
- [x] Bug fix #976?
- [x] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [x] Changelog updated?

      $ php bin/changelog

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -19,6 +19,7 @@ require __DIR__ . '/deploy/vendors.php';
 require __DIR__ . '/deploy/writable.php';
 
 use Deployer\Exception\RunException;
+use Deployer\Exception\ConfigurationException;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\Output;
 
@@ -43,41 +44,80 @@ set('user', function () {
 });
 
 // Number of releases to preserve in releases folder.
-set('keep_releases', 5);
+set('keep_releases', 5, function ($value) {
+    return is_integer($value) && $value >= -1;
+});
 
 // Repository to deploy.
-set('repository', '');
+set('repository', '', 'is_string');
 
 // List of dirs what will be shared between releases.
 // Each release will have symlink to those dirs stored in {{deploy_path}}/shared dir.
 // ```php
 // set('shared_dirs', ['storage']);
 // ```
-set('shared_dirs', []);
+set('shared_dirs', [], 'is_array');
 
 // List of files what will be shared between releases.
 // Each release will have symlink to those files stored in {{deploy_path}}/shared dir.
 // ```php
 // set('shared_files', ['.env']);
 // ```
-set('shared_files', []);
+set('shared_files', [], 'is_array');
 
 // List of dirs to copy between releases.
 // For example you can copy `node_modules` to speedup npm install.
-set('copy_dirs', []);
+set('copy_dirs', [], 'is_array');
 
 // List of paths to remove from {{release_path}}.
-set('clear_paths', []);
+set('clear_paths', [], 'is_array');
 
 // Use sudo for deploy:clear_path task?
-set('clear_use_sudo', false);
+set('clear_use_sudo', false, 'is_bool');
+
+// NOT IN THE NEW COMMON.PHP FILE
+/* set('writable_dirs', [], 'is_array'); */
+/* set('writable_mode', 'acl', function ($value) { */
+/*     return $value === 'chmod' */
+/*         || $value === 'chown' */
+/*         || $value === 'chgrp' */
+/*         || $value === 'acl'; */ 
+/* }); // chmod, chown, chgrp or acl. */
+/* set('writable_use_sudo', false, 'is_bool'); // Using sudo in writable commands? */
+/* set('writable_recursive', true, 'is_bool'); // Common for all modes */
+/* set('writable_chmod_mode', '0755', function ($value) { */
+/*     $chmodPattern = '/^[ugoa]*([-+=]([rwxXst]*|[ugo]))+|[-+=]?[0-7]+$/'; */
+/*     return preg_match($chmodPattern, $value); */
+/* }); // For chmod mode */
+/* set('writable_chmod_recursive', true, 'is_bool'); // For chmod mode only (if is boolean, it has priority over `writable_recursive`) */
+/* set('http_user', false, 'is_bool'); */
+/* set('http_group', false, 'is_bool'); */
+// END: NOT IN THE NEW COMMON.PHP FILE
 
 set('use_relative_symlink', function () {
     return commandSupportsOption('ln', '--relative');
 });
+
 set('use_atomic_symlink', function () {
     return commandSupportsOption('mv', '--no-target-directory');
 });
+
+// NOT IN THE NEW COMMON.PHP FILE
+/* set('composer_action', 'install', function ($value) { */
+/*     switch ($value) { */
+/*     case 'install': */
+/*     case 'i': */
+/*     case 'require': */
+/*     case 'u': */
+/*     case 'update': */
+/*     case 'upgrade': */
+/*         return true; */
+/*     } */
+/*     return false; */
+/* }); */
+
+/* set('composer_options', '{{composer_action}} --verbose --prefer-dist --no-progress --no-interaction --no-dev --optimize-autoloader --no-suggest', 'is_string'); */
+// END: NOT IN THE NEW COMMON.PHP FILE
 
 /**
  * Remote environment variables.
@@ -93,7 +133,7 @@ set('use_atomic_symlink', function () {
  * run('echo $KEY', ['env' => ['KEY' => 'over']]
  * ```
  */
-set('env', []);
+set('env', [], 'is_array');
 
 /**
  * Path to `.env` file which will be used as environment variables for each command per `run()`.
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -78,50 +78,12 @@ set('clear_paths', [], 'is_array');
 // Use sudo for deploy:clear_path task?
 set('clear_use_sudo', false, 'is_bool');
 
-// NOT IN THE NEW COMMON.PHP FILE
-/* set('cleanup_use_sudo', false, 'is_bool'); // Using sudo in cleanup commands? */
-/* set('writable_dirs', [], 'is_array'); */
-/* set('writable_mode', 'acl', function ($value) { */
-/*     return $value === 'chmod' */
-/*         || $value === 'chown' */
-/*         || $value === 'chgrp' */
-/*         || $value === 'acl'; */ 
-/* }); // chmod, chown, chgrp or acl. */
-/* set('writable_use_sudo', false, 'is_bool'); // Using sudo in writable commands? */
-/* set('writable_recursive', true, 'is_bool'); // Common for all modes */
-/* set('writable_chmod_mode', '0755', function ($value) { */
-/*     $chmodPattern = '/^[ugoa]*([-+=]([rwxXst]*|[ugo]))+|[-+=]?[0-7]+$/'; */
-/*     return preg_match($chmodPattern, $value); */
-/* }); // For chmod mode */
-/* set('writable_chmod_recursive', true, 'is_bool'); // For chmod mode only (if is boolean, it has priority over `writable_recursive`) */
-/* set('http_user', false, 'is_bool'); */
-/* set('http_group', false, 'is_bool'); */
-// END: NOT IN THE NEW COMMON.PHP FILE
-
 set('use_relative_symlink', function () {
     return commandSupportsOption('ln', '--relative');
 });
-
 set('use_atomic_symlink', function () {
     return commandSupportsOption('mv', '--no-target-directory');
 });
-
-// NOT IN THE NEW COMMON.PHP FILE
-/* set('composer_action', 'install', function ($value) { */
-/*     switch ($value) { */
-/*     case 'install': */
-/*     case 'i': */
-/*     case 'require': */
-/*     case 'u': */
-/*     case 'update': */
-/*     case 'upgrade': */
-/*         return true; */
-/*     } */
-/*     return false; */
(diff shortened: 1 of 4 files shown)


recipe/common.php
```diff
@@ -46,41 +47,142 @@
...
+}); // Repository to deploy.
 
-set('shared_dirs', []);
-set('shared_files', []);
+set('shared_dirs', [], function ($value) {
+    if (!is_array($value)) {
+        return false;
+    }
+    foreach ($value as $dir) {
+        if (!is_dir($dir)) {
```
antonmedv, line 63: I think this is not generally true. Some may want to add unexisting dirs.

recipe/common.php
```diff
@@ -46,41 +47,142 @@
...
+    }
+    return true;
+});
+
+set('shared_files', [], function ($value) {
+    if (!is_array($value)) {
+        return false;
+    }
+    foreach ($value as $file) {
+        if (!file_exists($file)) {
```
antonmedv, line 75: Same here.

recipe/common.php
```diff
@@ -46,41 +47,142 @@
...
+    return true;
+});
 
-set('copy_dirs', []);
+set('copy_dirs', [], function ($value) {
+    if (!is_array($value)) {
+        return false;
+    }
+    foreach ($value as $dir) {
+        if (!is_dir($dir)) {
```
antonmedv, line 87: Same here.

recipe/common.php
```diff
@@ -46,41 +47,142 @@
...
-set('writable_use_sudo', false); // Using sudo in writable commands?
-set('writable_recursive', true); // Common for all modes
-set('writable_chmod_mode', '0755'); // For chmod mode
-set('writable_chmod_recursive', true); // For chmod mode only (if is boolean, it has priority over `writable_recursive`)
+set('writable_dirs', [], function ($value) {
+    if (!is_array($value)) {
+        return false;
+    }
+    foreach ($value as $dir) {
+        if (!is_dir($dir)) {
```
antonmedv, line 99: Same here.

recipe/common.php
```diff
@@ -46,41 +47,142 @@
...
+    return $value === 'chmod'
+        || $value === 'chown'
+        || $value === 'chgrp'
+        || $value === 'acl'; 
+}); // chmod, chown, chgrp or acl.
 
-set('clear_paths', []);         // Relative path from release_path
-set('clear_use_sudo', false);    // Using sudo in clean commands?
+set('writable_use_sudo', false, function ($value) {
+    return is_bool($value);
```
antonmedv, line 114: Maybe create some helpers for this? `isBool`

antonmedv: Or just pass it as string `'is_bool'`.

recipe/common.php
```diff
@@ -46,41 +47,142 @@
  * Configuration
  */
 
-set('keep_releases', 5);
+set('keep_releases', 5, function ($value) {
+    return is_integer($value) && $value >= -1;
```
antonmedv, line 51: Also will be cool to give users meaningful error message, not just "Validation failed".

antonmedv (changes requested):

antonmedv:
Good work! 👍

recipe/common.php
brunofontes: Thanks! I am removing this `foreach`.

recipe/common.php
brunofontes: How about keeping the functions, but throwing specific errors as 'Config option "writable_use_sudo" is not a boolean'?

Or yet creating a helper that would receive 2 parameters, `$value` and a string to throw error?
```php
function isBool($value, $component) {
    if (!is_bool($value)) {
        throw new ConfigurationException("$component must have a boolean value");
    }
    return true;
}
```

antonmedv: Yes, with throw it's okay. It will be cool to reduce duplicate code for bool case, as it really common.

brunofontes: I came back on this helper function as it would include some not necessary complexity. I just used your `is_bool` string suggestion.

recipe/common.php
```diff
@@ -46,41 +47,83 @@
...
+    case 'u':
+    case 'update':
+    case 'upgrade':
+        return true;
+    }
+    return false;
+});
+
+set('composer_options', '{{composer_action}} --verbose --prefer-dist --no-progress --no-interaction --no-dev --optimize-autoloader --no-suggest', function ($value) {
+    return is_string($value);
```
antonmedv, line 121: Use string, please.

recipe/common.php
```diff
@@ -46,41 +47,83 @@
...
+
+set('writable_chmod_recursive', true, 'is_bool'); // For chmod mode only (if is boolean, it has priority over `writable_recursive`)
+set('http_user', false, 'is_bool');
+set('http_group', false, 'is_bool');
+
+set('clear_paths', [], function ($value) {
+    if (!is_array($value)) {
+        return false;
+    }
+    foreach ($value as $dir) {
```
antonmedv, line 88: `clear_paths` works the same as other dirs. Not able to use `is_dir`.

src/Configuration/Configuration.php
```diff
@@ -34,8 +49,31 @@ public function bind(Configuration $parent)
...
+                break;
+            }
+            $distance = levenshtein($name, $predefinedParam);
+            if ($distance < $closestDistance) {
+                $closestDistance = $distance;
+                $closestParam = $predefinedParam;
+            }
+        }
+        if (isset($closestParam)) {
+            print("<fg=yellow>Warning:</> \"$name\" not found in parameters. Did you mean \"$closestParam\"?\n");
```
antonmedv, line 75: Use warning func.

antonmedv:
Also, please rebase you PR.

src/Configuration/Configuration.php
```diff
@@ -7,17 +7,32 @@
...
 use function Deployer\get;
+use Deployer\Exception\ConfigurationException;
 use function Deployer\Support\array_merge_alternate;
 use function Deployer\Support\is_closure;
 use function Deployer\Support\normalize_line_endings;
 
 class Configuration implements \ArrayAccess
 {
+
+    const PREDEFINED_PARAMS = [
```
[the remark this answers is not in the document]
brunofontes: Oh, sure! I forgot there are another ones. At the moment, I cannot really think about any automated way into doing it. Is it ok to keep including them manually at this PREDEFINED_PARAMS array?

recipe/common.php
```diff
@@ -46,41 +47,83 @@
...
+    }
+    return false;
+});
+
+set('composer_options', '{{composer_action}} --verbose --prefer-dist --no-progress --no-interaction --no-dev --optimize-autoloader --no-suggest', function ($value) {
+    return is_string($value);
+});
 
-set('env', []); // Run command environment (for example, SYMFONY_ENV=prod)
+set('env', [], function ($value) {
```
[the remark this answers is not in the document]
brunofontes: Ooops, sorry! Fixing it.

recipe/common.php
brunofontes: Here as well!

recipe/common.php
brunofontes: Fixing it as well.

src/Configuration/Configuration.php
brunofontes: I've tried to use it, but it asks for a Context and the function is called directly from the `set` function in Configuration file. Any tip here?

brunofontes:
> Also, please rebase your PR.

Oh, for sure!

src/Configuration/Configuration.php
antonmedv: I think we can check `Context::has()` in warning func and output message without host prefix.

src/Configuration/Configuration.php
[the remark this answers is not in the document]
antonmedv: We can check such an existing on `set()` function call. If there is already something similar to what we are trying to configure - it's a possible error other there.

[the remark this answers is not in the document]
brunofontes: This is a good idea! But we would still need a `PREDEFINED_PARAMS` array to avoid warnings when first setting `bin/dir` and `bin/git`, for example. Actually, it could be triggered every time if any recipe try to use a very similar name as an existing one.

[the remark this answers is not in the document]
antonmedv: If there no previous values, no need to warning. So no need to keep PREDEFINED_PARAMS.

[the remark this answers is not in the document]
brunofontes: Sorry, not sure if I am understanding correctly. Because the wrong configuration would not have a previous value, so it would not trigger the notification as well.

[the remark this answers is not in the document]
antonmedv: There is no such thing wrong configuration before first config are set.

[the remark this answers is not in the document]
antonmedv: Only next set calls can be wrong.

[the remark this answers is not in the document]
brunofontes: Oh, so there a way to identify which `set` calls came from the configuration file! Ok, I will try to find that.

recipe/common.php
brunofontes: I noticed you removed the `composer_options` and a few others, so I am removing them as well.

src/Configuration/Configuration.php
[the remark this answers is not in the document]
brunofontes: I've included a check for which file is calling the `set` function, so the verification is only going to run if it came from the deploy.php file. Could you take a look if that is ok? Thanks! :)

src/Configuration/Configuration.php
brunofontes: I updated it as well. I hope it works.

src/Configuration/Configuration.php
```diff
@@ -47,18 +47,35 @@ public function set(string $name, $value, callable $callback = null): void
...
         }
-        if (array_key_exists($name, $this->validations) 
-            && !call_user_func($this->validations[$name], $value)
-        ) {
-            throw new ConfigurationException("Config option \"$name\" has an invalid value.");
-        }
         $callingFile = pathinfo(debug_backtrace()[1]['file'], PATHINFO_FILENAME);
-        if (!$this->has($name) && $callingFile === 'deploy') {
-            $this->verifyConfigurationName($name);
+        if ($callingFile === 'deploy') {
```
antonmedv, line 51: What if recipe called something else?  This should be a general solution. I can merge it with hard coded recipe name.

antonmedv (changes requested):

src/Configuration/Configuration.php
brunofontes: Yes, sure, sorry! Can I match it against `$GLOBALS['deployFile']` or there is a better way? 

Thanks for you patience!

src/Configuration/Configuration.php
```diff
@@ -36,12 +40,64 @@ public function bind(Configuration $parent): void
...
+     * @param mix $value
+     */
+    protected function verifyConfigurationValue(string $name, $value): void
+    {
+        if ($this->parent) {
+            $definedCallable = &$this->parent;
+        } else {
+            $definedCallable = &$this;
+        }
+        if (array_key_exists($name, $definedCallable->validations)
```
loevgaard, line 72: If the callable is a native function, i.e. `is_array` it would be nice to have a check if the function exists and generate an exception telling this to the user.

src/Configuration/Configuration.php
antonmedv: This is also a bad move. Just check if it second set via levingshtain.

brunofontes: With just the levenshtein, it could trigger the warning for correct properties as `bin/git` after `bin/dir` being applied, as they are close enough. That's why I had included the `PREDEFINED_PARAMS`. It was working as a whitelist for some of the correct set names.

I couldn't think a better way to make it work by the moment other than using the `PREDEFINED_PARAMS` whitelist or checking if the set came from the user configuration file.

src/Configuration/Configuration.php
brunofontes: Oh, the `set` function argument is defined as `callable`, so if the function doesn't exist, it will already throw an error. :)

loevgaard: That's right. Didn't get that!

antonmedv:
Снимок экрана 2020-11-03 в 19 06 45
Really misleading on connect error.

src/Configuration/Configuration.php
```diff
@@ -36,12 +40,64 @@ public function bind(Configuration $parent): void
...
      * @param mixed $value
+     * @param callable $callback
      */
-    public function set(string $name, $value): void
+    public function set(string $name, $value, callable $callback = null): void
     {
+        if (is_callable($callback)) {
+            $this->validations[$name] = $callback;
+        }
+        $callingFile = pathinfo(debug_backtrace()[1]['file'], PATHINFO_FILENAME);
```
antonmedv, line 50: Unfortunately, this hack should not be used here, this is really not a solution.
I will try to cherry-pick some commits and implement it myself.

brunofontes: I don't like it either. It was more to understand better your idea on this.

The content validation is working way better, but I got stuck on this key validation. :(

antonmedv:
Fixed by 76f07b4e05b898db852686ef8cac93d8b1a7f692
Снимок экрана 2020-11-04 в 12 39 22

PR #2198, opened 2020-10-11
[#2197] Created E2E testing environment

### LINKED ISSUE #2197: Create an E2E testing environment to enable more complex testing scenarios
We should create an end-to-end testing environment that could be used to test more complex scenarios, that the unit tests cannot verify.

E2E testing environment should be easy to set up and should allow to test real-life scenarios, so it should incorporate SSH, HTTP server and so on.

bsacharski:
- [X] New feature
- [ ] BC breaks
- [ ] Deprecations
- [X] Tests added
- [X] Changelog updated

Created docker-based end-to-end testing environment, which allows
for testing more complex scenarios.

Additionally added FunctionE2ETest class that contains test scenario
for `run` function with placeholders.

The E2E tests have a safeguard, that checks whether `E2E_ENV` env variable is present and is truthy - if no, then the tests will be marked as skipped.

Closes #2197


bsacharski:
@antonmedv Additionally to implementing e2e testing env, I think I've also found a bug in `src/Executor/Master.php` - the `createConnectProcess` method didn't pass a deploy file name to the `ConnectCommand` process.

antonmedv:
> @antonmedv Additionally to implementing e2e testing env, I think I've also found a bug in src/Executor/Master.php - the createConnectProcess method didn't pass a deploy file name to the ConnectCommand process.

Yes, it's a bug. Also I'm thinking maybe combine createProcess and createConnectProcess together to reduce the chance of new errors. It will require writing stringify if input options only. Can you work on such PR?

antonmedv:
About PR. I'll review it more closely tomorrow. A few things:
- Lets move everything (docekr/, etc) to test folder as it related to tests.
- Try to add github action for e2e test as well.
- For e2e test it will be cool to try to deploy for example whole Laravel application, as it what users see.
- Let's keep phpunit and e2e separately.

bsacharski:
> * Lets move everything (docekr/, etc) to test folder as it related to tests.

Done - `docker/` directory is now under the `test/` dir.
> * Try to add github action for e2e test as well.

Done - see [this](https://github.com/deployphp/deployer/pull/2198/checks?check_run_id=1250233054) run
> * For e2e test it will be cool to try to deploy for example whole Laravel application, as it what users see.

Hmm...we would have to add the boilerplate laravel repo to the fixtures - it would add some size to the repo. Other solutions include:
  * creating the laravel project as a part of the E2E tests
  * creating a separate repo on GitHub with laravel app just for testing purposes

> * Let's keep phpunit and e2e separately.

I've separated the test suites, made `AbstractE2ETest` class rely directly on TestCase, prepared different phpunit config and separate bootstrap file.

antonmedv:
> creating a separate repo on GitHub with laravel app just for testing purposes

We can use external repo for sure.

> Will look into that after we have E2E testing env done ;)

Already did it myself. No need to worry.

test/docker/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+### `server` container
+
+The `server` container contains:
+
+* Apache (with the `DocumentRoot` set to `/var/www/html/current`)
+* git
+* PHP 7.3
+* SSH server with
+* sudo (user `deployer` can use `sudo` without having to enter passwords)
```
antonmedv, line 88: Maybe actually test it? Deployer now can ask sudo password to user. It will be cool to see it tested here as well.

bsacharski: Ok, though I will create a separate user like `deployer_sudopass` for this scenario - shouldn't increase overall complexity & size of the image, so we should be fine. Will also update the `README.md` to explain that there are 2 different deployer users.

antonmedv: I think we can go with one user. Just enable sudo. It will be enough. For, other cases just add `-o sudo_pass=password`.

bsacharski: I've added two testcases:
* use sudo with password passed via stdin
* use sudo with password set via `-o sudo_pass`

I've also discovered a discrepancy in `run()` behavior - the function wouldn't return any output, when the command was being executed with `sudo`.
I've modifed the function to make it more consistent with outher cases - I'm capturing an output of the actual command and then return it from the function.

bsacharski:
> We can use external repo for sure.
> 
Do you have any repo in mind? Maybe something created on the GitHub in the `deployphp` namespace? That way the tests wouldn't rely on any third party and the project would have full control over the sample project.

antonmedv:
Yes, for sure! Let’s create a repo in deployphp org. Can you create one? I'll push it other there. For example it can be some boilerplate Laravel app.

bsacharski:
> Can you create one? I'll push it other there. For example it can be some boilerplate Laravel app.

https://github.com/bsacharski/e2e-laraver-boilerplate - it's been a really, really long while since I've had anything to do with Laravel, so it's just `composer create-project` thing. 

Maybe others will contribute with more meaningful things like DB migrations and so on (the E2E env will have to be updated, but it can be done once necessary). 

Once the repo is accessible from the _deployphp_ organization, I'll update the e2e tests with final recipe for Laravel app.

bsacharski:
@antonmedv I've added very simple E2E for deploying laravel application. I'm not sure I like it (given that it doesn't clean up after itself, meaning other frameworks tests may be "tainted" with laravel files), but it'll do for now ;)

Github actions take around additional 3 minutes to finish testing - ~1:20 is spent while building docker stack. Technically it could be improved by bushing the images to external repository, but it might be worth exploring after everything settles down. ~1:40 is mostly spent to installing a lot of dependencies that come as a part of Laravel app itself during the deploy - not sure anything can be done about it to be honest.

antonmedv:
Cool 😎 
Also what about separation check: Union and e2e? In github actions. So unit test can be done faster. 

Also there is new provision recipe. Maybe it can be used to provision Docker server?

bsacharski:
> Also what about separation check: Union and e2e? In github actions. So unit test can be done faster.

Hmm...technically having units and e2e running independently in parallel could reduce tests duration a little, however running e2e tests in case of unit-test failure would probably be unnecessary (if unit-tests fail, there's a high probability that e2e will also fail and they are much expensive to run). I guess it's doable but I would do it only if its possible to break e2e tests as soon as unit-tests fail.

> Also there is new provision recipe. Maybe it can be used to provision Docker server?
Hmm, skimming through `provision.php` it looks like it's aimed at non-ephemeral entities like VMs or even normal machines. Unless it is done in `Dockerfile` as a part of image build process (not sure how to do that, given that it would have to be running on the "localhost" -in the intermediate container, ran as a build process - with PHP already installed, or the container that it's executed in is committed (not something I would recommend), I don't think we would benefit from it. 

It is definitely something that could be tested as a part of e2e tests though :) Although I imagine it would increase the testing time even more (packages would have to be downloaded & installed & processed, instead of simply relying on downloading pre-generated layers of docker filesystem).

antonmedv:
>  however running e2e tests in case of unit-test failure would probably be unnecessary 

In some cases unit tests are failing for because of side tests (as changelog tests, etc.), and running e2e make sense.

> It is definitely something that could be tested as a part of e2e tests though :)

It will be so cool I want to make this provision.php recipe core functionality of Deployer, as well as common.php recipe.

bsacharski:
> In some cases unit tests are failing for because of side tests (as changelog tests, etc.), and running e2e make sense.

Ok, will move the e2e to separate action

> It will be so cool I want to make this provision.php recipe core functionality of Deployer, as well as common.php recipe.

Sure. Let's have that done in a follow up issue ;)

bsacharski:
> Ok, will move the e2e to separate action

@antonmedv  E2E tests moved to separate job, see [here](https://github.com/deployphp/deployer/pull/2198/checks?check_run_id=1277067954)

antonmedv:
Btw, here is repo https://github.com/deployphp/test-laravel

src/functions.php
```diff
@@ -357,8 +357,9 @@ function run($command, $options = [])
             }
             $run("echo -e '#!/bin/sh\necho \"%sudo_pass%\"' > $askpass", array_merge($options, ['sudo_pass' => $password]));
             $run("chmod a+x $askpass", $options);
-            $run(sprintf('export SUDO_ASKPASS=%s; %s', $askpass, preg_replace('/^sudo\b/', 'sudo -A', $command)), $options);
+            $output = $run(sprintf('export SUDO_ASKPASS=%s; %s', $askpass, preg_replace('/^sudo\b/', 'sudo -A', $command)), $options);
             $run("rm $askpass", $options);
+            return $output;
```
antonmedv, line 362: Nice catch! 👍

bsacharski:
> Btw, here is repo https://github.com/deployphp/test-laravel

Updated the deploy-config for Laravel E2E test, to use the https://github.com/deployphp/test-laravel repo

antonmedv:
@bsacharski nice work 👏👏👏

Issue #2209, opened 2020-10-20
shared folder file copying is too verbose


Gamesh:
### Describe the bug
When shared folders in release contain lots of files, each and everyone of them is listed then copying. Which causes lots of spam output. In severe cases when lots of output fails GitLab CI builds because it's exceeds 4194304 bytes, which GitLab can handle: 
> Job's log exceeded limit of 4194304 bytes.

### To reproduce
1. create a release with a shared folder that has thousands of files
2. set that folder as shared in deployer
3. release

### Expected behavior
1. should specify only a folder that is getting copied.

### Environment
- Deployer version: 6.8.0
- PHP version: 7.3
- Deployment target(s) OS: Debian stretch

Gamesh:
this line is the culprit https://github.com/deployphp/deployer/blob/112601294c3c76273322ecc4e5f995e34f27b6f6/recipe/deploy/shared.php#L29
can we remove the `-v` ? 
or maybe make it depend on deplyer verbosity level say `-vvv` would use `cp -v` ?

antonmedv:
I think -v for -vvv is okay. Can you make PR?

Gamesh:
yes. how can i get verbosity level inside the task?

antonmedv:
output()->getVerbosity()

PR #2211, opened 2020-10-21
fixed issue #2209

Gamesh:
Copy shared files and folder should be verbose only when deployer is run with Very verbose output `-vvv`

- [X] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [ ] Changelog updated?

### DIFF
--- a/recipe/deploy/shared.php
+++ b/recipe/deploy/shared.php
@@ -30,7 +30,7 @@ task('deploy:shared', function () {
                 run(
                     sprintf(
                         'cp -r%s {{release_path}}/%s %s/%s',
-                        output()->getVerbosity() === OutputInterface::VERBOSITY_VERY_VERBOSE ? 'v' : '',
+                        output()->getVerbosity() === OutputInterface::VERBOSITY_DEBUG ? 'v' : '',
                         $dir,
                         $sharedPath,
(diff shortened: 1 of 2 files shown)


recipe/deploy/shared.php
```diff
@@ -53,7 +62,15 @@
...
             // Copy file in shared dir if not present
-            run("cp -rv {{release_path}}/$file $sharedPath/$file");
+            run(
+                sprintf(
+                    'cp -r%s {{release_path}}/%s %s/%s',
+                    output()->getVerbosity() === OutputInterface::VERBOSITY_DEBUG ? 'v' : '',
+                    $file,
+                    $sharedPath,
+                    $file
+                )
```
antonmedv, line 72: This look way too verbose. Better use $v var. Just inline it “-r$v”

antonmedv (changes requested):

recipe/deploy/shared.php
Gamesh: ok, it seemed to be more readable, but i guess different preferences 😄 changed it

antonmedv (approved):

PR #2213, opened 2020-10-22
[composer 3] [6.x] Removed deprecated option "--no-suggest"

andrey-helldar:
- [x] Deprecations?
- [x] Changelog updated?

> You are using the deprecated option "--no-suggest". It has no effect and will break in Composer 3.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 
 
+## master
+[v6.4.5...master](https://github.com/deployphp/deployer/compare/v6.4.5...master)
+
+### Changed
+- [composer 3] Removed deprecated option "--no-suggest"
+
+
 ## v6.4.5
 [v6.4.4...v6.4.5](https://github.com/deployphp/deployer/compare/v6.4.4...v6.4.5)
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 # Changelog
 
 
+## master
+[v6.4.5...master](https://github.com/deployphp/deployer/compare/v6.4.5...master)
+
+### Changed
+- [composer 3] Removed deprecated option "--no-suggest"
```
staabm, line 8: composer 2

andrey-helldar: No. Composer 3:
image

staabm: Hmm composer 2 is not even released.. Not sure it makes sense assumptions on what composer 3 will look like

andrey-helldar: I will try to explain it so that Google Translate will understand.

I am currently using Composer 2.0.0 RC2. When the "composer update/install" command is executed with the `--no-suggest` switch, it warns that in "composer 2" this switch does nothing and is contained there only as backward compatibility.
He also warns that in composer 3 this key will be removed.
Composer 2 will not display package suggestions by default. To view them, you need to explicitly run the "composer suggests" command.

image

antonmedv:
This is already in v7. I think better leave v6 as is.

PR #2218, opened 2020-10-23
added phpstan static analysis via GitHub Actions

clxmstaab:
- [ ] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [x] Changelog updated?

      $ php bin/changelog

this is only the first step which does the setup for phpstan.

USAGE

- run `composer phpstan` to start the analysis
- run `composer phpstan-baseline` to ignore all current warnings (put them into the baseline)

after this initial PR lands we can work PR for PR to level up the phpstan coverage so the value of having the tools gets better over time.


phpstan-baseline.neon
```diff
@@ -0,0 +1,82 @@
+parameters:
+	ignoreErrors:
```
clxmstaab, line 2: this file contains all errors phpstan currently reports. because of they are contained in the baseline phpstan will ignore those when invoking `composer phpstan`.

we should fix the errors in a separate PR.

.github/workflows/phpstan.yml
```diff
@@ -0,0 +1,34 @@
+name: phpstan static analysis
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
loevgaard, line 7:
```suggestion
  pull_request: ~
```

.github/workflows/phpstan.yml
```diff
@@ -0,0 +1,34 @@
...
+jobs:
+  phpstan:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Validate composer.json and composer.lock
+      run: composer validate
```
loevgaard, line 18:
```suggestion
      run: composer validate --strict
```

Will it work with strict?

.github/workflows/phpstan.yml
```diff
@@ -0,0 +1,34 @@
...
+      uses: actions/cache@v2
+      with:
+        path: vendor
+        key: ${{ runner.os }}-php-${{ hashFiles('**/composer.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-php-
+
+    - name: Install dependencies
+      if: steps.composer-cache.outputs.cache-hit != 'true'
+      run: composer install --prefer-dist --no-progress
```
loevgaard, line 31: You can use https://github.com/ramsey/composer-install instead of these lines

phpstan.neon.dist
```diff
@@ -0,0 +1,7 @@
+includes:
```
loevgaard, line 1: Why the `.dist` extension?

loevgaard (commented):
I would add this job to the existing workflow. It has a few advantages: You can rely on the result of this job in a later job if you need to and you can also have the result of this job reflected in the build status badge

clxmstaab:
> I would add this job to the existing workflow. It has a few advantages: You can rely on the result of this job in a later job if you need to and you can also have the result of this job reflected in the build status badge

I dont see we have requirements for that right now.

the result of the action is already reflected in the badges, see:

image

.github/workflows/phpstan.yml
clxmstaab: I took inspiration of the exising job for this settings

(same for the comment below)

phpstan.neon.dist
clxmstaab: because phpcs is configured the same way already

.github/workflows/phpstan.yml
clxmstaab: took inspiration of the already existing action

.github/workflows/phpstan.yml
clxmstaab: took inspiration of the exising job for this settings

loevgaard:
This is what I mean:

image

That's why your job should be in the same file as the other jobs, else that badge doesn't reflect the *actual* build status of the project

phpstan.neon.dist
loevgaard: Usually you'll have `*.dist` when it makes sense to have local configuration values. For example phpunit might have local env vars for tests to run correctly. I don't think that's the case for phpstan?

.github/workflows/phpstan.yml
loevgaard: If the existing job already has this it doesn't make sense to add it again

.github/workflows/phpstan.yml
loevgaard: It's never a shame to improve things :)

.github/workflows/phpstan.yml
loevgaard: I don't think @antonmedv would mind if you improved things :)

clxmstaab:
> That's why your job should be in the same file as the other jobs, else that badge doesn't reflect the _actual_ build status of the project


I see. thx for the followup. lets see what @antonmedv thinks.. having phpstan inside a "tests" actions isn't that clean either.. I will take the way anton prefers.

.github/workflows/phpstan.yml
```diff
@@ -0,0 +1,34 @@
+name: phpstan static analysis
```
antonmedv, line 1: Just leave phpstan)

.gitattributes
```diff
@@ -8,3 +8,5 @@
 .travis.yml export-ignore
 CHANGELOG.md export-ignore
 phpunit.xml.dist export-ignore
+phpstan-baseline.neon export-ignore
+phpstan.neon.dist export-ignore
```
antonmedv, line 12: So why two files? Let's leave only one?

staabm: One is the phpstan config and the other one is generated by phpstan baseline command.

Those cannot be combined

antonmedv:
What kind of errors phpstan can show? Deployer is not a really complicated PHP code. A lot of it is bash.

staabm:
It can report a lot of errors. Examples are
- redundant code
- dead code
- type errors (parameters, return-values, properties, local vars, etc)
- get most native php-src errors at analysis time, which you otherwise only happen at runtime

See https://phpstan.org/blog/find-bugs-in-your-code-without-writing-tests for a short introduction

staabm:
Thanks for merging. Will provide PRs next week for fixing the reported errors and leveling up phpstan, so it can detect even more problems

PR #2222, opened 2020-10-24
Add composer_version to choose which composer version to install

jkaan:
This solves https://github.com/deployphp/deployer/issues/2012 & https://github.com/deployphp/deployer/discussions/2220

- [ ] Bug fix #…?
- [X] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [X] Changelog updated?

I wasn't sure how to add tests for this, I could do that though.

### DIFF
--- a/recipe/deploy/vendors.php
+++ b/recipe/deploy/vendors.php
@@ -6,6 +6,15 @@ set('composer_action', 'install');
 
 set('composer_options', '--verbose --prefer-dist --no-progress --no-interaction --no-dev --optimize-autoloader');
 
+/**
+ * Can be used to choose what composer version to install.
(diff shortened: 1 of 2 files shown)


recipe/deploy/vendors.php
```diff
@@ -6,6 +6,8 @@
 
 set('composer_options', '--verbose --prefer-dist --no-progress --no-interaction --no-dev --optimize-autoloader');
 
+set('composer_version', null);
```
antonmedv, line 9: Please add a comment before set() function about what it is, what format, code example, etc.

antonmedv:
Other all look good. I think we can omit tests for this part right now.

jkaan:
@antonmedv Just updated the PR :).

PR #2228, opened 2020-10-25
Implemented E2E tests for provision recipe

bsacharski:
Created new container using Ubuntu 20.04 as a base, for a E2E test
environment for the provision recipe.

Due to container sharing a kernel with host os, the firewall provisioning
had to be turn off during the testing procedure.

- [ ] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [x] Tests added
- [ ] Changelog updated?

### DIFF
--- a/test/e2e/ProvisionE2ETest.php
+++ b/test/e2e/ProvisionE2ETest.php
@@ -1,0 +1,29 @@
+<?php
+namespace e2e;
+
+use Deployer\Exception\Exception;
+
+class ProvisionE2ETest extends AbstractE2ETest
+{
+    private const RECIPE = __DIR__ . '/recipe/provision.php';
+
+    /**
+     * @throws Exception
+     */
+    public function testProvisionRecipe(): void
+    {
+        $this->init(self::RECIPE);
+
+        $this->tester->setInputs(['deployer']); // set input for deployer user pass
+
+        $this->tester->run([
+            'provision',
+            '-f' => self::RECIPE,
+            '-o' => [ 'firewall=false' ], // disable firewall provisioning in docker e2e environment
+            'selector' => 'all'
+        ]);
+
+        $display = trim($this->tester->getDisplay());
+        self::assertEquals(0, $this->tester->getStatusCode(), $display);
+    }
+}
--- a/docs/recipe/provision.md
+++ b/docs/recipe/provision.md
@@ -27,19 +27,19 @@
 
 ## Config
 ### php_version
-[Source](/recipe/provision.php#L10)
+[Source](/recipe/provision.php#L11)
 
 
 
 ### sudo_password
-[Source](/recipe/provision.php#L11)
+[Source](/recipe/provision.php#L12)
 
 
 
 
 ## Tasks
 ### provision
-[Source](/recipe/provision.php#L16)
+[Source](/recipe/provision.php#L17)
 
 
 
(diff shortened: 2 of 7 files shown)


antonmedv:
I guess it will run a really LONG time. What if we split it to separate action `test / provision`, as most PRs can be merged without waiting for the provision test to finish. Also, maybe to rerun the provision test only in the recipe changed?

antonmedv:
> Successful in 5m

But maybe this is okay.

bsacharski:
E2E tests are tricky - for now we have only few of them, so they finish in 5 minutes, but in future we may have to think about some smart strategy for picking which scenarios to run or maybe finding a way to parallelize some parts of the verification process.

test/e2e/ProvisionE2ETest.php
```diff
@@ -0,0 +1,29 @@
...
+
+        $this->tester->run([
+            'provision',
+            '-f' => self::RECIPE,
+            '-o' => [ 'firewall=false' ], // disable firewall provisioning in docker e2e environment
+            'selector' => 'all'
+        ]);
+
+        $display = trim($this->tester->getDisplay());
+        self::assertEquals(0, $this->tester->getStatusCode(), $display);
```
antonmedv, line 27: Maybe also run some command to check what provision did succeed? Like check php version?

bsacharski: Ok, will add tests that verify presence of PHP, PHP-FPM & nginx

clxmstaab:
not sure this makes sense.. just dropping a random idea (I am not a docker expert).

As I read the github action it looks like the docker container is build within the action before the testsuite is started.

couldn't the container be pre-built (since it shouldn't change often) and the testsuite should live outside the container and be mapped from a local directory into the container environment for testing?

bsacharski:
Disclaimer: I have used docker images mostly in private projects, not in the public ones.

@clxmstaab you're mostly right ;)

The docker _images_ are being built as a part of the e2e github action. The images contain things that are necessary for the deployer to run (_deployer_ image) and to allow testing (_server_ and _provisioned_ images). The images themselves do not contain deployer files, nor do they contain test files (they are mostly just a runtime for actual deployer to run).
After the images are built, the files used for whole build process are being cached, so hopefully when the docker tries to build the images once again (for instance, when new commit is pushed) and nothing in docker context has been modified, the process will rely on cached data. This does speed the images build process a little.

Once the images are available, the `docker-compose up` command creates the containers and mounts the whole project directory (alongside test files) to the `/project` directory of the `deployer` container. Only then, the deployer files are available from inside said container.

Like you've noticed, the process could be sped up even more (at least in theory, given that downloading image layers also take time) with prebuilt images. However this also introduces some difficulties like:
* we need to have a separate action that will build the images and push them to the central docker images repository (this should be smart, to skip building images unless it is really necessary)
* introduces some difficulties, when adding/changing some "assumptions" about the E2E testing environment - the MR will have to update the docker files, docker-compose definition (to use newer version of the images) and then finally e2e test scenarios. I don't know how the docker images are being typically built for "public" - maybe it would have to be managed in the second project, to skip unwanted image revision bumps.

clxmstaab:
@bsacharski thx for the great description how things are going.

is there already known where the E2E tests spends the whole time? which parts take the most time?

bsacharski:
@clxmstaab Mostly running E2E test scenarios ;) Deploying Laravel app & provisioning ubuntu server take a lot of time, due to the fact that they pull a lot of packages. But that's what E2E tests are about - testing as close to the real case as possible :)

antonmedv:
It possible to speed up provision if skip dhparam generation

PR #2239, opened 2020-10-27
contrib: Add default value for missing {{target}} option

### LINKED ISSUE #2238: Slack recipe is missing "target" option
### Describe the bug
`contrib/slack.php` uses `{{ target }}`, but that option doesn't exists:

Screenshot 2020-10-27 at 11 31 53

### To reproduce

Just follow the docs to add a basic Slack notification. But below the lines I've added to `deploy.php`:
```
require 'contrib/slack.php';
set(
    'slack_webhook',
    '....'
);
after('deploy:failed', 'slack:notify:failure');
```

### Expected behavior

I would expect the target option to exists or information in the docs.

### Environment
- Deployer version: `v7.0.0-beta.2`
- PHP version: `7.4.10`
- Deployment target(s) OS: `Linux`

carstendietrich:
Most of the contrib recipes use the target option but do not come with a default value.
Therefore using those recipes will lead to an exception.

This PR adds the target option to all recipes and sets the hostname as default.

- [x] Bug fix #2238 
- [x] Changelog updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 
 ### Fixed
 - Lots, and lots of long-standing bugs.
+- Missing default value for {{target}} option in most of the contrib recipes. [#2238]
 
 
 ## v6.8.0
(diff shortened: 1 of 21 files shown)


antonmedv:
This is not the way)) Did it myself. Actual way is much simpler: c747a5945bcf678bf2a2143e9eef67f74a733d46

PR #2246, opened 2020-10-28
added type declarations for src/Documentation

nofutur3:
- [x]  Improvement #2187  (partial)
- [ ] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [ ] Changelog updated?


src/Documentation/DocRecipe.php
```diff
@@ -44,7 +44,8 @@ public function parse($content)
         foreach (explode("\n", $content) as $i => $line) {
             $m = [];
             $match = function ($regexp) use ($line, &$m) {
-                return preg_match("#$regexp#", $line, $m);
```
nofutur3, line 47: The result of this function was not used anywhere, so I changed it.

PR #2249, opened 2020-10-29
Allow option merge on host add()

### LINKED ISSUE #2248: Merge option on host "add()"
Currently using `host()->add()` will not merge with a option value defined globally via `set()` / `add()`.

A specific example would be setting Rsync excludes:

```php
add('rsync', [
    'exclude' => [
        'path/excluded/on/all/hosts',
    ],
]);

host('first')
    ->add('rsync', [
        'exclude' => [
            'path/excluded/on/first/host',
        ],
    ]);
host('second')
    ->add('rsync', [
        'exclude' => [
            'path/excluded/on/second/host',
        ],
    ]);
```

Currently no merge is performed here. What is worse: actually the whole option is redefined for the host which then misses unrelated nested options. In the example given here the `rsync` option for each host will only contain the `exclude` defined here, nothing else (see `rsync.php` recipe)

Thus I'd suggest to check if options set on hosts via `add()` can be merged with their defaults if not yet defined.

Currently the following workaround is required:

```php
host('first')
    ->set('rsync', function (): array {
        return array_merge_alternate(
            Deployer::getDefault('rsync'), // Avoid recursion by get('rsync')
            [
                'exclude' => [
                    'path/excluded/on/first/host',
                ],
            ]
        );
    });
host('second')
    ->set('rsync', function (): array {
        return array_merge_alternate(
            Deployer::getDefault('rsync'), // Avoid recursion by get('rsync')
            [
                'exclude' => [
                    'path/excluded/on/second/host',
                ],
            ]
        );
    });
```

mbrodala:
This ensures that global defaults are used as base on host()->add() in case no host-specific base has been set yet.

- [x] Bug fix #2248?
- [x] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [ ] Changelog updated?

      $ php bin/changelog

### DIFF
--- a/src/Configuration/Configuration.php
+++ b/src/Configuration/Configuration.php
@@ -56,15 +56,11 @@ class Configuration implements \ArrayAccess
 
     public function add(string $name, array $array): void
     {
-        if ($this->has($name)) {
-            $config = $this->get($name);
-            if (!is_array($config)) {
-                throw new ConfigurationException("Config option \"$name\" isn't array.");
-            }
-            $this->set($name, array_merge_alternate($config, $array));
-        } else {
-            $this->set($name, $array);
+        $config = $this->get($name);
+        if (!is_array($config)) {
+            throw new ConfigurationException("Config option \"$name\" isn't array.");
         }
+        $this->set($name, array_merge_alternate($config, $array));
     }
 
     /**


mbrodala:
Nothing to do here, see related ticket.

PR #2252, opened 2020-10-30
fix phpstan level 4 warnings

clxmstaab:
- [ ] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [ ] Changelog updated?

      $ php bin/changelog

like in previous PRs with the same topic, this changes don't have a CHANGELOG.md line. this was your prefered way to handle this in the past.

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -192,7 +192,7 @@ function task(string $name, $body = null): Task
  * @param string $task The task before $that should be run.
  * @param string|callable $do The task to be run.
  *
- * @return Task|void
+ * @return Task|null
  */
 function before(string $task, $do)
 {
@@ -210,7 +210,7 @@ function before(string $task, $do)
  * @param string $task The task after $that should be run.
  * @param string|callable $do The task to be run.
  *
- * @return Task|void
+ * @return Task|null
  */
 function after(string $task, $do)
 {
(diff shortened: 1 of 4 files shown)


src/functions.php
```diff
@@ -192,7 +192,7 @@ function task(string $name, $body = null): Task
  * @param string $task The task before $that should be run.
  * @param string|callable $do The task to be run.
  *
- * @return Task|void
```
clxmstaab, line 195: its invalid phpdoc to combine `void` with another type

antonmedv: This is strange. Why func can't return something in one code path, and not return in another?

clxmstaab: its `Task|null` then and requires and explicit `return null`.

thats the opinion of phpstan though. in php-src its not necessary a technical requirement.

clxmstaab:
cannot regenerate the docs, since https://github.com/deployphp/deployer/issues/2253

src/functions.php
antonmedv: Maybe it’s possible to allow Task|void? Because it’s valid php after all.

clxmstaab: feel free to discuss it with the phpstan maintainer :-)

https://github.com/phpstan/phpstan/issues/4014

clxmstaab:
remaining phpstan error is fixed in https://github.com/deployphp/deployer/pull/2260

clxmstaab:
ready to land

Issue #2202, opened 2020-10-13
Conditionally enable/disable maintenance mode in Magento2 recipe


andyexeter:
This Capistrano Magento2 gem has some smart maintenance mode handling which Deployer could take inspiration from:

https://github.com/davidalger/capistrano-magento2/blob/3d32c7d654bd9221f67e20c27fc4e62503bdd73d/lib/capistrano/tasks/magento.rake#L384-L466

As well as resolving #1379, this would enable zero-down deploys when the database schema/data and app config does not require an update.

I am happy to look at providing a PR for this feature when I get some time.

peterjaap:
This feature is the biggest reason why [Magento 2 Deployer Plus](https://github.com/jalogut/magento2-deployer-plus) exists. I'm not sure whether it's a good idea to start copying that project, since it uses Deployer as the basis. I would be more in favor of keeping this recipe simple and straight forward, and let people with advanced use-cases use Magento 2 Deployer Plus on top of Deployer.

andyexeter:
Interesting, I didn't know that existed.

I don't agree with the notion that Deployer should be intentionally deprived of a general enhancement just because a third party has made a recipe providing equivalent functionality. I also don't agree that this is an "advanced use-case". Surely 100% of users would prefer a zero-down deploy if it is possible.

For these reasons, I believe this does belong in Deployer.

antonmedv:
> Surely 100% of users would prefer a zero-down deploy if it is possible.

I also agree with @andyexeter.

Let's try to incorporate  Magento 2 Deployer Plus into Deployer recipes @peterjaap?  As soon v7 will be ready, it also will be really awesome to copy/paste such wonderful recipe from @jalogut.

Can we do it @jalogut @peterjaap @andyexeter @Schrank ?

Schrank:
I don't do much M2 lately and especially no deployer with it. But I think having more functionality is good. Maybe we add a configuration option to turn it off/on so it is OFF by default?

antonmedv:
I'm actually thinking of just implementing a more advanced recipe. Better to provide users the best functionality out of the box.

Schrank:
Agree. I didn't read the Capistrano implementation, but I think it is a good way. My fear is, that it breaks with translations, not sure the CLI is translated. So maybe it is a good idea to provide a PR on magento2 to make sure, that we get a machine readable format.

👍  from my side.

antonmedv:
@peterjaap what do you think?

[...]

PR #2262, opened 2020-11-02
Magento 2 Zero downtime deployment

unailopez16:
- [ ] Bug fix
- [X] New feature  #2202
- [ ] BC breaks
- [ ] Deprecations
- [ ] Tests added
- [ ] Changelog updated

Implements check for Magento 2 database migration needs, in order to get zero downtime deployments otherwise.

Also fixes magento 2.4 deployments due to composer autoload issue https://github.com/magento/magento2/issues/23251

Tested with Deployer 7-beta.3 and Magento 2.2 | 2.3 | 2.4 (to date only 2.3 and 2.4 are officially supported by magento).

I've checked it does not introduce so far BC breaks for these versions.

### DIFF
--- a/recipe/magento2.php
+++ b/recipe/magento2.php
@@ -6,6 +6,7 @@ require_once __DIR__ . '/common.php';
 use Deployer\Exception\RunException;
 use Symfony\Component\Process\Exception\ProcessFailedException;
 
+const CONFIG_IMPORT_NEEDED_EXIT_CODE = 2;
 const DB_UPDATE_NEEDED_EXIT_CODE = 2;
 const MAINTENANCE_MODE_ACTIVE_OUTPUT_MSG = 'maintenance mode is active';
 
@@ -74,6 +75,43 @@ task('magento:maintenance:disable', function () {
     run("if [ -d $(echo {{current_path}}) ]; then {{bin/php}} {{current_path}}/bin/magento maintenance:disable; fi");
 });
 
+desc('Config Import');
+task('magento:config:import', function () {
+    $configImportNeeded = false;
+
+    try {
+        run('{{bin/php}} {{release_path}}/bin/magento app:config:status');
+    } catch (ProcessFailedException $e) {
+        if ($e->getProcess()->getExitCode() == CONFIG_IMPORT_NEEDED_EXIT_CODE) {
+            $configImportNeeded = true;
+        } else {
+            throw $e;
+        }
+    } catch (RunException $e) {
+        if ($e->getExitCode() == CONFIG_IMPORT_NEEDED_EXIT_CODE) {
+            $configImportNeeded = true;
+        } else {
+            throw $e;
+        }
+    }
+
+    if ($configImportNeeded) {
+        // check whether maintenance mode is enabled or not in order to keep the status afterwards
+        $maintenanceModeStatusOutput = run("{{bin/php}} {{release_path}}/bin/magento maintenance:status");
+        $maintenanceModeActive = strpos($maintenanceModeStatusOutput, MAINTENANCE_MODE_ACTIVE_OUTPUT_MSG) !== false;
+
+        if (!$maintenanceModeActive) {
+            invoke('magento:maintenance:enable');
+        }
+
+        run('{{bin/php}} {{release_path}}/bin/magento app:config:import --no-interaction');
+
+        if (!$maintenanceModeActive) {
+            invoke('magento:maintenance:disable');
+        }
+    }
+});
+
 desc('Upgrade magento database');
 task('magento:upgrade:db', function () {
     $databaseUpgradeNeeded = false;
@@ -120,6 +158,7 @@ desc('Magento2 deployment operations');
 task('deploy:magento', [
     'magento:compile',
     'magento:deploy:assets',
+    'magento:config:import',
     'magento:upgrade:db',
     'magento:cache:flush'
 ]);
--- a/recipe/magento2.php
+++ b/recipe/magento2.php
@@ -61,7 +61,7 @@ task('magento:compile', function () {
 
 desc('Deploy assets');
 task('magento:deploy:assets', function () {
-    run("{{bin/php}} {{release_path}}/bin/magento setup:static-content:deploy {{static_content_locales}}");
+    run("{{bin/php}} {{release_path}}/bin/magento setup:static-content:deploy {{static_content_locales}} --force");
 });
 
 desc('Enable maintenance mode');
--- a/recipe/magento2.php
+++ b/recipe/magento2.php
@@ -25,15 +25,15 @@ set('shared_files', [
 set('shared_dirs', [
     'var/composer_home',
     'var/log',
-    'var/cache',
     'var/export',
     'var/report',
+    'var/import',
     'var/import_history',
     'var/session',
     'var/importexport',
     'var/backups',
     'var/tmp',
-    'pub/sitemaps',
+    'pub/sitemap',
     'pub/media'
 ]);
 set('writable_dirs', [
--- a/recipe/magento2.php
+++ b/recipe/magento2.php
@@ -4,7 +4,6 @@ namespace Deployer;
 require_once __DIR__ . '/common.php';
 
 use Deployer\Exception\RunException;
-use Symfony\Component\Process\Exception\ProcessFailedException;
 
 const CONFIG_IMPORT_NEEDED_EXIT_CODE = 2;
 const DB_UPDATE_NEEDED_EXIT_CODE = 2;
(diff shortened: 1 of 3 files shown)


recipe/magento2.php
```diff
@@ -19,15 +25,15 @@
...
-    'var/cache',
     'var/export',
     'var/report',
+    'var/import',
     'var/import_history',
     'var/session',
     'var/importexport',
     'var/backups',
     'var/tmp',
-    'pub/sitemaps',
```
peterjaap, line 30: Why did you change this?

recipe/magento2.php
```diff
@@ -48,13 +54,14 @@
 // Tasks
 desc('Compile magento di');
 task('magento:compile', function () {
+    run('cd {{release_path}} && {{bin/composer}} dump-autoload -o');
```
peterjaap, line 57: Any specific reason why this is in here?

recipe/magento2.php
```diff
@@ -48,13 +54,14 @@
...
     run("{{bin/php}} {{release_path}}/bin/magento setup:di:compile");
     run('cd {{release_path}} && {{bin/composer}} dump-autoload -o');
 });
 
 desc('Deploy assets');
 task('magento:deploy:assets', function () {
-    run("{{bin/php}} {{release_path}}/bin/magento setup:static-content:deploy {{static_content_locales}}");
+    run("{{bin/php}} {{release_path}}/bin/magento setup:static-content:deploy {{static_content_locales}} --force");
```
peterjaap, line 64: Any specific reason why we're suddenly forcing this?

recipe/magento2.php
```diff
@@ -69,7 +76,39 @@
...
+        if ($e->getExitCode() == DB_UPDATE_NEEDED_EXIT_CODE) {
+            $databaseUpgradeNeeded = true;
+        } else {
+            throw $e;
+        }
+    }
+
+    if ($databaseUpgradeNeeded) {
+        // check whether maintenance mode is enabled or not in order to keep the status afterwards
+        $maintenanceModeStatusOutput = run("{{bin/php}} {{release_path}}/bin/magento maintenance:status");
```
peterjaap, line 99: Wouldn't it be quicker to check for the maintenance flag instead of kicking off another bin/magento command?

recipe/magento2.php
unailopez16: Starting magento 2.2 i guess, static content deploy throws an error when env is not in production mode
I may have a reason to have my environment in developer o default mode but still want to deploy it  
--force just allows to generate the static content in any mode

recipe/magento2.php
unailopez16: Indeed it would be, but I decided to stick to the framework tools.  Do you see any other cons beside speed?

recipe/magento2.php
peterjaap: I think we should add 'pub/sitemap' (instead of replace) to keep backwards compatibility, or we should include this change in the next major release.

recipe/magento2.php
peterjaap: This problem should not occur on deploy - the generated folder will be recreated each deploy, leaving no stale classes behind.

recipe/magento2.php
peterjaap: I think your case is an exception - by default, Magento should not be in developer mode (and never in default mode - that is the worst of both worlds) when deploying to production. Leaving `--force` would generate an error - exactly as it should. If you have an exception use case, you can override this in your own customizations to the recipe.

recipe/magento2.php
peterjaap: Not really, this is fine, just wanted to know the reason :)

recipe/magento2.php
peterjaap: It doesn't break BC in terms of deployment, but people's sitemaps could be gone if they've relied on the pub/sitemaps naming dictated by this recipe.

antonmedv: We are actually about to land v7. So breaking BC is ok.

peterjaap: Fine by me!

recipe/magento2.php
unailopezvdshop: was impossible for me to deploy a m2.4 without dumping the autoload before di-compiling, got this error all the time
```
Warning: include(/var/www/html/vendor/composer/../../generated/code/Magento/Framework/App/ResourceConnection/Proxy.php):
failed to open stream: No such file or directory in /var/www/html/vendor/composer/ClassLoader.php on line 444
```
everything fine in m2.2 and m2.3

recipe/magento2.php
unailopezvdshop: i agree with the production envs part, but in a normal workflow we also use deployer for dev and test environments, were it's usual  to have them in developer mode in order to get more insights about errors for instance  
do you see any way to compatibilize this at least for previous environments?

peterjaap: Use `on()`? https://deployer.org/docs/api.html#on

recipe/magento2.php
peterjaap: Ok LGTM

recipe/magento2.php
unailopezvdshop: don't see the way to identify previous envs, since there could be a wide variety on stage naming conventions: dev, development, testing, staging...  
anyway thinking twice we should not leave responsability of checking appropiate env mode to this static-content command, but maybe moving this check to an independent task when we find the way to identify non-production environments

peterjaap: I'm still in favor of not putting this in the default recipe at all and leaving it up to the user of the recipe to think of a way to identify their non-production environments by overriding this task, if they want to.

unailopez16: fine for me

barryvdh: I would skip the --force indeed.

peterjaap:
LGTM!

barryvdh:
Does this check also take care of config.php updates? I think Deployer Plus checks that as well? (Is normally done by the setup upgrade, right?)

unailopez16:
The new feature keeps using the same `setup:upgrade` command, no changes on that.  
It handles the config.php file by adding new added modules to it in case the file is not control versioned.  
Ideal case is to control version the config.php, in that case setup:upgrade command won't add anything to it since changes already come from the new release git clone.  
  
Regarding other configurations stored in config.php (themes, stores, general config...), current recipe still does not have anything implemented to automatically import this new config (e.g app:config:import), could be made in the scope of another PR

barryvdh:
If I make a change locally to app/etc/config.php (which is in git control) I get this warning:
> Exception #0 (Magento\Framework\Exception\LocalizedException): The configuration file has changed. Run the "app:config:import" or the "setup:upgrade" command to synchronize the configuration.

So that would suggest that config changes are handled by the setup:upgrade command as well. I'm just not sure if config changes are picked up by that db:status command, or that this check is also needed: https://github.com/jalogut/magento2-deployer-plus/blob/master/recipe/magento_2_2_5/config.php Otherwise the config would not be imported anymore when not database changes are made.

antonmedv:
Some test failing?

unailopez16:
@barryvdh you are totally right, new config (system, scopes...) would not be imported unless database changes forced a `setup:upgrade`.  

Looks like we need to add a `app:config:import`  task in order to handle this use case.

unailopez16:
> Looks like we need to add a `app:config:import` task in order to handle this use case.

Have a look please https://github.com/deployphp/deployer/pull/2262/commits/cbe59376a73c6a2ef122dd2ee946b365fbba1167

This implementation approach also uses `app:config:status` command to detect changes in configuration, which was introduced in magento 2.2.4 (Apr '18, currently out of support by magento). Lower version deploys would fail.

barryvdh:
So this would be breaking, should we rename this to magento2.3 or something? 
Also the code maintenance code is duplicated, but if you do setup first, it wouldn't need the config anymore, right?

unailopez16:
> So this would be breaking, should we rename this to magento2.3 or something?

other option could be maintaining compatibility checking the current version and taking this approach only with newer versions. What do you think?


> Also the code maintenance code is duplicated, but if you do setup first, it wouldn't need the config anymore, right?

I thought about the code maintenance duplicity, but it's a tricky one. Could happen that config:import command is needed but not  setup:upgrade. I guess both commands need a maintenance mode active to be safely executed.

recipe/magento2.php
```diff
@@ -67,9 +75,78 @@
     run("if [ -d $(echo {{current_path}}) ]; then {{bin/php}} {{current_path}}/bin/magento maintenance:disable; fi");
 });
 
+desc('Config Import');
+task('magento:config:import', function () {
+    $configImportNeeded = false;
+
+    try {
+        run('{{bin/php}} {{release_path}}/bin/magento app:config:status');
+    } catch (ProcessFailedException $e) {
```
antonmedv, line 84: This already not needed. v7 throws RunExection for localhosts too now.

recipe/magento2.php
```diff
@@ -67,9 +75,78 @@
...
+});
+
 desc('Upgrade magento database');
 task('magento:upgrade:db', function () {
-    run("{{bin/php}} {{release_path}}/bin/magento setup:upgrade --keep-generated");
+    $databaseUpgradeNeeded = false;
+
+    try {
+        run('{{bin/php}} {{release_path}}/bin/magento setup:db:status');
+    } catch (ProcessFailedException $e) {
```
antonmedv, line 121: Same here.

antonmedv:
So, thus pr ready to merge?

unailopez16:
still pending to decide if it would make sense to keep compatibility with magento <= 2.2.3
what are your thoughts?

antonmedv:
Let's introduce `magento_version` config:
```php
set('magento_version', function () {
    // detect version
    return run('{{bin/magento}} --version');
});
```

And add if (com_version) there needed. For example take a look at laravel recipe.

antonmedv (approved):

PR #2265, opened 2020-11-03
Allow changing Composer channel

Mika56:
Composer installer can now take an argument to select which version channel to use: https://github.com/composer/getcomposer.org/pull/156/files
This allows for exemple to stay on major version 1 without specifying a specific Composer version

- [ ] Bug fix #…?
- [X] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [ ] Changelog updated?

      $ php bin/changelog

### DIFF
--- a/recipe/deploy/vendors.php
+++ b/recipe/deploy/vendors.php
@@ -17,6 +17,11 @@ set('composer_options', '--verbose --prefer-dist --no-progress --no-interaction 
  */
 set('composer_version', null);
 
+/**
+ * Set this variable to stable, snapshot, preview, 1 or 2 to select which Composer channel to use
+ */
+set('composer_channel', null);
+
 set('bin/composer', function () {
     if (commandExist('composer')) {
         return '{{bin/php}} ' . locateBinaryPath('composer');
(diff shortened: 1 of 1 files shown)


recipe/deploy/vendors.php
```diff
@@ -27,11 +32,19 @@
...
 
     if ($composerVersionToInstall) {
         $installCommand .= " -- --version=" . $composerVersionToInstall;
     }
+    elseif ($composerChannel) {
+        $composerValidChannels = ['stable', 'snapshot', 'preview', '1', '2', ];
+        if(!in_array($composerChannel, $composerValidChannels)) {
+            throw new \Exception('Selected Composer channel '.$composerChannel.' is not valid. Valid channels are: '.implode(', ', $composerValidChannels));
+        }
+        $installCommand .= " -- --".$composerChannel;
```
antonmedv, line 46: Does it need additional `--`?

Mika56: Yes it does. Usage is:
```shell
$ php composer-installer.php -- --1
$ php composer-installer.php -- --2
$ php composer-installer.php -- --snapshot
$ php composer-installer.php -- --preview
```

antonmedv: Will it work with a specified composer version? See a few lines above.

antonmedv:
Ping @Mika56

recipe/deploy/vendors.php
Mika56: I don't think it will, which is why I've used an `elseif`. I'm not sure it would even make sense: if you want a specific version, use `composer_version`, if you want the latest version of a channel, use `composer_channel`?

antonmedv: Do it. Thanks.

antonmedv (approved):

kszymukowicz:
I can see one drawback of current `set('bin/composer'` 

It does not check at all if something changed on `composer_version` or `composer_channel`. So always the first version downloaded and stored in `{{deploy_path}}/.dep/composer.phar` will be used.

Would be nice to have kind of check if version required is equal to the one from `'{{bin/php}} ' . locateBinaryPath('composer');` or from `'{{bin/php}} {{deploy_path}}/.dep/composer.phar`

antonmedv:
I agree. Maybe you can create pr to fix it?

kszymukowicz:
I am on it.

kszymukowicz:
I would also throw an exception if both `composer_version` or `composer_channel` are not set. 
This can not be a desired behaviour that composer version is unexpected. 
IMO - one of the main feature of deploy - it must be stable. 
Having unexpected composer version could be dangerous.

kszymukowicz:
An update - I will have it today at evening.

PR #2284, opened 2020-11-15
Safety to avoid modifying files outside deploy_path

hansemschnokeloch:
Using writable_dirs with writable_use_sudo can lead to a disaster on a server (example using /var/log instead of var/log as writable dir).

- [ ] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [x] Changelog updated?

      $ php bin/changelog

### DIFF
--- a/recipe/deploy/writable.php
+++ b/recipe/deploy/writable.php
@@ -39,12 +39,16 @@ set('writable_chmod_mode', '0755');
 // For chmod mode only (if is boolean, it has priority over `writable_recursive`)
 set('writable_chmod_recursive', true);
 
+// Allow absolute path in writable_dirs
+set('writable_allow_absolute_path', false);
 
 desc('Make writable dirs');
 task('deploy:writable', function () {
-    // Safety to avoid modifying files outside the deploy_path
-    $safeDirs = array_map(function($v) { return ltrim($v, '/'); }, get('writable_dirs'));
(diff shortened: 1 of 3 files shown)


antonmedv:
Nice addition. 👍🏻

martijngastkemper:
Maybe add this to the docs. Because this feature removes the possibility to use absolute paths.

hansemschnokeloch:
I modified my draft. My first version was not safe enough, because with a space before the slash "  /var/log" the trim was not working anymore.
Now I add the deploy_path to the directories. I also added an option to disable this behavior and use absolute paths if needed.

recipe/deploy/writable.php
```diff
@@ -39,12 +39,16 @@
 // For chmod mode only (if is boolean, it has priority over `writable_recursive`)
 set('writable_chmod_recursive', true);
 
+// Allow absolute path in writable_dirs
+set('writable_allow_absolute_path', false);
```
antonmedv, line 41: I think it’s otherkill. I assume setting writable outside deploy path is really corner case and should be ignored.

antonmedv:
But instead trimming slashes error message to user can be shown. It’s much better design.

hansemschnokeloch:
I agree, throwing an error message is much better, and we need less code lines.

PR #2293, opened 2020-11-18
Sync content_version for magento assets between hosts.

### LINKED ISSUE #2281: [Magento2 recipe] deployed_version.txt
- Deployer version: N/A
- Deployment OS: N/A

If you are running magento2 on multiple servers, and deploying static content.
It will create a "deployed_version.txt" with a number in pub/static.

Now if this number is different from different servers (and you are running varnish/etc in front) it will fail to get the content from that server.

deployed_version just contains a timestamp so if they are not deployed at the exact second, it will be missed.

Example
Server1 - deployed_version.txt contains: 1
Server2 - deployed_version.txt contains: 2

Now it will try to fetch `https://example.com/static/1/file.js` from server 1 - Everything is fine - But then it will try and fetch `https://example.com/static/1/secondfile.js` from server 2 - which does not exists (because the deployed version number is 2) and throws a 404.

### How it could be fixed

Copy deployed_version.txt (or the content) from 1 server and then insert it in all servers just before deployment.

The js/css/etc files does not needs to be changed, and no folders needs to be changed, just that number in that file.

antonmedv:
- [x] Bug fix #2281?
- [x] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [ ] Changelog updated?

### DIFF
--- a/docs/recipe/magento2.md
+++ b/docs/recipe/magento2.md
@@ -12,6 +12,7 @@
   * [`recipe/common.php`](/docs/recipe/common.md)
 * Config
   * [`static_content_locales`](#static_content_locales)
+  * [`content_version`](#content_version)
   * [`shared_files`](#shared_files)
   * [`shared_dirs`](#shared_dirs)
   * [`writable_dirs`](#writable_dirs)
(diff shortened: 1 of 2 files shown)


antonmedv:
Sync will be done only on one host and spread to others. 
Снимок экрана 2020-11-18 в 23 25 52

antonmedv:
Ping @peterjaap

antonmedv:
Ping @peterjaap 😃please check my solution.

peterjaap:
@antonmedv LGTM!

PR #2303, opened 2020-11-26
Speedup cleanup

TheRatG:
- [ ] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [x] Changelog updated?

Speedup cleanup. Some update after we faced `rm -rf ` 3-4 minutes long.

- https://unix.stackexchange.com/questions/37329/efficiently-delete-large-directory-containing-thousands-of-files
-  https://www.kinamo.be/en/support/faq/efficiently-remove-a-zillion-files-on-linux-servers

### DIFF
--- a/recipe/deploy/cleanup.php
+++ b/recipe/deploy/cleanup.php
@@ -18,6 +18,7 @@ task('deploy:cleanup', function () {
     }
 
     foreach ($releases as $release) {
+        run("$sudo find {{deploy_path}}/releases/$release/* -delete", $runOpts);
         run("rm -rf {{deploy_path}}/releases/$release", $runOpts);
     }
 
--- a/recipe/deploy/cleanup.php
+++ b/recipe/deploy/cleanup.php
@@ -18,7 +18,7 @@ task('deploy:cleanup', function () {
     }
 
     foreach ($releases as $release) {
-        run("ionice -c 3 nice -19 find {{deploy_path}}/releases/$release/ -delete", $runOpts);
+        run("find {{deploy_path}}/releases/$release/ -delete", $runOpts);
         run("rm -rf {{deploy_path}}/releases/$release", $runOpts);
     }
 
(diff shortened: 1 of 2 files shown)


antonmedv:
Looks like some checks are failing.

TheRatG:
> Looks like some checks are failing.

fixed

recipe/deploy/cleanup.php
```diff
@@ -18,6 +18,7 @@
     }
 
     foreach ($releases as $release) {
+        run("ionice -c 3 nice -19 find {{deploy_path}}/releases/$release/ -delete", $runOpts);
```
staabm, line 21: I think this line deserves a comment, what it is expected todo.
never heard of this command.

antonmedv: Will it work on all *nix systems? Solaris? Bsd?

antonmedv: Not sure if this is a solution for general case. Maybe your setup is too specific.

TheRatG: Yep, I should delete ionice, Bsd doesn't support.

recipe/deploy/cleanup.php
```diff
@@ -18,7 +18,7 @@
     }
 
     foreach ($releases as $release) {
-        run("ionice -c 3 nice -19 find {{deploy_path}}/releases/$release/ -delete", $runOpts);
+        run("find {{deploy_path}}/releases/$release/ -delete", $runOpts);
```
antonmedv, line 21: So what is performance gain? rm rf vs find?

TheRatG: I found the problem with my deploy, it was mount option **discard**. So I will close request.

TheRatG:
Thanks for support.

PR #2324, opened 2020-12-03
Refresh composer binary version according to 'composer_version' or 'composer_channel'

kszymukowicz:
https://github.com/deployphp/deployer/pull/2265#issuecomment-736771912

- [ ] Bug fix #…?
- [x ] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [ ] Changelog updated?

I think we need to set at least 'composer_channel' to '2' or 'stable'. Otherwise we can not know to what channel compare the currently installed composer version. This is why if both 'composer_version' is null and 'composer_channel' is null the script will download composer each time. 
When you set 'composer_version' or 'composer_channel' everything is working as expected.

### DIFF
--- a/recipe/deploy/vendors.php
+++ b/recipe/deploy/vendors.php
@@ -26,7 +26,7 @@ set('bin/composer', function () {
     $composerVersionToInstall = get('composer_version', null);
     $composerChannelToInstall = get('composer_channel', null);
     $composerValidChannels = ['stable', 'snapshot', 'preview', '1', '2',];
-    if (!in_array((string)$composerChannelToInstall, $composerValidChannels, true)) {
+    if ($composerChannelToInstall && !in_array((string)$composerChannelToInstall, $composerValidChannels, true)) {
         throw new \Exception('Selected Composer channel ' . $composerChannelToInstall . ' is not valid. Valid channels are: ' . implode(', ', $composerValidChannels));
     }
 
@@ -40,7 +40,7 @@ set('bin/composer', function () {
     }
 
     if ($composerBin) {
-        $currentComposerVersionRaw = run('{{bin/php}} ' . $composerBin . ' --version');
+        $currentComposerVersionRaw = run($composerBin . ' --version');
         if (preg_match('/(\\d+\\.\\d+)(\\.\\d+)?-?(RC|alpha|beta|dev)?\d*/', $currentComposerVersionRaw, $composerVersionMatches)) {
             $currentComposerVersion = $composerVersionMatches[0] ?? null;
             if ($currentComposerVersion) {
--- a/recipe/deploy/vendors.php
+++ b/recipe/deploy/vendors.php
@@ -24,11 +24,7 @@ set('composer_channel', null);
 
 set('bin/composer', function () {
     $composerVersionToInstall = get('composer_version', null);
-    $composerChannelToInstall = get('composer_channel', null);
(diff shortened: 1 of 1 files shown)


recipe/deploy/vendors.php
```diff
@@ -23,26 +23,49 @@
 set('composer_channel', null);
 
 set('bin/composer', function () {
+    $composerVersionToInstall = get('composer_version', null);
+    $composerChannelToInstall = get('composer_channel', null);
+    $composerValidChannels = ['stable', 'snapshot', 'preview', '1', '2',];
+    if ($composerChannelToInstall && !in_array((string)$composerChannelToInstall, $composerValidChannels, true)) {
+        throw new \Exception('Selected Composer channel ' . $composerChannelToInstall . ' is not valid. Valid channels are: ' . implode(', ', $composerValidChannels));
```
antonmedv, line 30: Well, after thinking, let's omit this part. Let's install stable composer and allow user yo override it is more user friendly.

antonmedv:
Also, test will really good for such a complex addition.

recipe/deploy/vendors.php
kszymukowicz: Can you plz check if the latest version is what you were thinking about ?

recipe/deploy/vendors.php
```diff
@@ -20,29 +20,48 @@
...
     if (test('[ -f {{deploy_path}}/.dep/composer.phar ]')) {
-        return '{{bin/php}} {{deploy_path}}/.dep/composer.phar';
+        $composerBin = '{{bin/php}} {{deploy_path}}/.dep/composer.phar';
     }
 
-    $composerVersionToInstall = get('composer_version', null);
-    $composerChannel = get('composer_channel', null);
-    $installCommand = "cd {{release_path}} && curl -sS https://getcomposer.org/installer | {{bin/php}}";
+    if ($composerBin) {
+        $currentComposerVersionRaw = run($composerBin . ' --version');
```
antonmedv, line 39: This code is complicated. Please, add comments. 
Also, do we really need to support channel install? I'd like to keep code simple, what is the point in channel installs? I'm for vomiting this part., let's by default install from stable, or specified version if specified.

kszymukowicz: What do you mean by " install from stable" - from channel named "stable"?

kszymukowicz: Plz look here: https://getcomposer.org/versions
This are the only available information we have to figure out if we need to download composer or not.
So if I understand well if we install composer the standard way without any channel option then we install from "stable" channel.

antonmedv: Correct.

kszymukowicz: So if you want to support only "stable" channel then the only simplifing I can do is to remove line 47-49 - this line is support for snapshot channel.

BryanLaura (approved):

recipe/deploy/vendors.php
kszymukowicz: I added some comments. Please give me info if this can be final code so I can write some tests.

antonmedv:
Actually, I'm thinking to completely remove this functionality. This code has lots of code paths, regexp and looks completely unmaintainable (also code that is already in the master branch). So I'm going to delete it and write a better doc on how to approach composer versions.

PR #2365, opened 2020-12-28
InitCommand: ask for confirmation before overwriting potentially existing deploy.[php/yaml]

thyseus:
if deploy.[php/yaml] already exists

- [X] New feature?
- [X] Changelog updated?

### DIFF
--- a/src/Command/InitCommand.php
+++ b/src/Command/InitCommand.php
@@ -57,6 +57,15 @@ class InitCommand extends Command
             $recipePath = "deploy.$language";
         }
 
+        // Avoid accidentally override of existing file.
+        if (file_exists($recipePath)) {
+            $message = "Warning: $recipePath already exists. Do you want to override the existing file?";
+            if (!$io->confirm($message, false)) {
+                $io->text('Exiting due to user choice.');
+                exit(0);
+            }
+        }
+
         // Template
         $template = $io->choice('Select project template', $this->recipes(), 'common');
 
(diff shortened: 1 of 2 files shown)


src/Command/InitCommand.php
```diff
@@ -100,7 +109,13 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         if (isset($tempHostFile)) {
             $host = file_get_contents($tempHostFile);
         }
-        $hosts = explode(',', $io->ask('Hosts (comma separated)', $host));
+        $hosts = [];
+        $answer = $io->ask('Hosts (comma separated)');
```
antonmedv, line 113: Why this change? Also looks like `$host` default value is missing.

antonmedv (changes requested):

src/Command/InitCommand.php
thyseus: Behavior before this change:

```
 Hosts (comma separated):
 > 

 TypeError  in InitCommand.php on line 103:

  explode() expects parameter 2 to be string, null given
  ```
  
  Behavior after this change:
  
  ```
   Hosts (comma separated):
 > 

No hosts given, you need to define them manually in deploy.php.
Successfully created deploy.php
```

`$host` is initialized in line 108.

antonmedv: $host is initialized but not used. It should go to `$io->ask('Hosts (comma separated)', $host)`.

thyseus: Ah i see. It is the default value of the potential guess. I added it to the appropriate place.

antonmedv:
Thanks! I added it manually.

PR #2366, opened 2020-12-28
Introduce sendfile command

thyseus:
This PR adds a "sendfile" command which allows to quickly "deploy" a single file to a single or multiple configured hosts.
This can be handy for hotfixes which needs to be deployed immediately or for using deployer as managing tool for multiple servers.
It is simply a wrapper for scp.

- [X] New feature?
- [X] Changelog updated?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 *.phar
 .phpunit.result.cache
 docker-compose.override.yml
+.php_cs.cache
--- a/src/Command/SendFileCommand.php
+++ b/src/Command/SendFileCommand.php
@@ -118,15 +118,35 @@ class SendFileCommand extends Command
 
         $scpOptions = $input->getOption('scpOptions');
 
+        $verbosity = $this->determineVerbosity($output);
+
         foreach ($hosts as $host) {
             $port = '';
 
             if ($host->has('port')) {
-                $port = "-P {$host->getPort()}";
+                $port = " -P {$host->getPort()}";
             }
+
             $connectionString = $host->getConnectionString();
-            passthru("scp $scpOptions $port $source $connectionString:$deployPath");
+
+            $command = "scp$verbosity$port $source $connectionString:$deployPath";
+
+            if ($output->getVerbosity() != OutputInterface::VERBOSITY_QUIET) {
+                $output->writeln("Executing: $command");
+            }
+
+            passthru($command);
         }
+
         return 0;
+    }
+
+    protected function determineVerbosity(OutputInterface $output): string
+    {
+        if ($output->getVerbosity() == OutputInterface::VERBOSITY_DEBUG) {
+            return ' -v ';
+        }
+
+        return '';
     }
 }
--- a/src/Command/SendFileCommand.php
+++ b/src/Command/SendFileCommand.php
@@ -43,6 +43,11 @@ class SendFileCommand extends SelectCommand
             'Source'
         );
         $this->addArgument(
+            'selector',
+            InputArgument::REQUIRED,
+            'One or more target hosts where the file should be sent to'
+        );
+        $this->addArgument(
             'hostname',
             InputArgument::OPTIONAL,
             'Hostname'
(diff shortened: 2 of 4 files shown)


antonmedv:
Better to split into two.

thyseus:
> Better to split into two.

You mean https://github.com/deployphp/deployer/pull/2365 and https://github.com/deployphp/deployer/pull/2366 ? 
I just rebased and splitted them as prompted.

thyseus:
@antonmedv what do you think about renaming the `selector` argument in `SelectCommand` to something like `targetHosts` to better name, what they do ?

antonmedv:
If it possible to use as trait it’s good.  Will be cool to see it. 

But what use case of this command? Maybe upload task is enough?

antonmedv:
Also for hotfixes Deployer has dep push command.

thyseus:
I see. `dep push` is great for the 'classical' deployment.
My specific (real world1) use case is, i have several hosts (100+) which i define via a foreach in deploy.php like this:
```php
for ($i = 1; $i <= 254; $i++) {
 host("host$i")->hostname("192.168.1.$i");
}
```
Now, using `vendor/bin/dep sendfile random_config_file host4,host6,host8` i quickly can decide where to push which file in a gitlab pipeline or similar scripts. This makes deployer very powerful and a alternative to chef, puppet, ansible, and what all this server configuration tools are around.

I would like to refactor into a trait in a seperate PR independent of your decision if you include this feature into the deployer core. If you choose not to, i simple copy this command into my specific project. In my opinion it is very useful for the rest of the world, tough. Its up to you :)

Schrank:
Thank you very much! Chiming in here :-) I think it is a nice idea. Adding a few tests and then merge it.

@thyseus do you have time and motivation to add a couple tests for `SendFileCommand`?

thyseus:
@Schrank hi - thanks a lot for answering. 

time to write tests: yes
motivation: yes

- but do you have an idea how to "mock" the "real" execution of rsync in an unit test? is there an example where i can lean on ?

src/Command/SendFileCommand.php
```diff
@@ -0,0 +1,148 @@
...
+        foreach ($hosts as $host) {
+            $port = '';
+
+            if ($input->getOption('targetPath')) {
+                $deployPath = $input->getOption('targetPath');
+            } else {
+                $deployPath = $host->get('deploy_path', '~');
+            }
+
+            if ($host->has('port')) {
```
Schrank, line 122: you might want to use $host->getConnectionString here, but I don'T know wether this is compatible with scp or only rsync.

Schrank: stupid me - sorry :)

.gitignore
```diff
@@ -2,3 +2,4 @@
 *.phar
 .phpunit.result.cache
 docker-compose.override.yml
+.php_cs.cache
```
Schrank, line 5: Totally unsure about this: I would prefer that in a global gitignore, but I'm fine merging it anyway. @antonmedv @peterjaap if you have any opinion on that, tell me :)

Schrank:
You can have a look into the e2e tests. Tbh I haven't yet but I HOPE there is in what I expect: docker boxes wot do real end to end tests 🙈 

That said: What I was thinking about were unit tests

.gitignore
antonmedv: I think it’s okay.

antonmedv:
👍🏻 Let’s work on some tests and merge it.

antonmedv:
BTW, why only sendfile? What about folders?

antonmedv:
Also, I was able to do replicate this command with a simple task:
```php
option('from', null, InputOption::VALUE_REQUIRED);
option('to', null, InputOption::VALUE_REQUIRED);
task('upload', function () {
    upload(input()->getOption('from'), input()->getOption('to'));
});
```
I think such an easy task can be implemented in user's recipes/or contrib.

PR #2393, opened 2021-01-28
Check (shared) file exists

dsmink:
Doing a check if the shared file already exists.
A touch could result in a 'Permission denied' causing the deploy to stop.


antonmedv:
I think in this case deploy should stop?

dsmink:
@antonmedv
Why? The shared file does not have to be writable by the person (or its group) performing the deploy is it? This is the case in our situation. Also, touching a file that already exists only changes the modification time. That is where our deploy fails now.

antonmedv:
Got it.

Schrank (approved):

Schrank:
@antonmedv do you want/need anything to merge this? A test or something?

antonmedv:
Nice! Thanks for pr and merge hustle.

PR #1207, opened 2017-05-14
Added Lumen recipe

sandervankasteel:
This PR adds support for Lumen, this is PR is based on the work by Anton Medvedev with his Laravel recipe. 

| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | N/A

> Do not forget to add notes about your changes to [CHANGELOG.md](https://github.com/deployphp/deployer/blob/master/CHANGELOG.md)

PR #2405, opened 2021-02-09
Create Lumen recipe

mikestreety:
- [ ] Bug fix #…?
- [x] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [x] Changelog updated?

      $ php bin/changelog

- - -

Thanks for the great work! Makes deployment so much easier, thank you.

Following on from https://github.com/deployphp/deployer/pull/1207, this is a recipe that extends Laravel and customises for Lumen.

It unsets and makes private a lot of the Lumen specific commands (so when you run just `dep` they don't show up) and adds `artisan cache:clear` to the deployment. 

I wasn't sure about the tests or what "BC breaks" is.

Thanks


mikestreety:
On review of the tests, it seems `setPrivate()` is not valid.

How would you recommended unsetting tasks? Maybe a `laravel.shared` recipe (or similar) is required which specifies the shared paths, but then Lumen and Laravel extend with their tasks instead?

antonmedv:
Maybe just create separate recipe not extending laravel? I now think this is better approach.

mikestreety:
Apologies. I only did as in the other Merge Request you recommended it

antonmedv:
Yes, now I think maintaining  separate recipes will will be easier for us.

lorisleiva:
Hi there 👋 

I'm not sure what it the point of overriding Laravel artisan tasks to noop functions.

If you're using Lumen, you should just not add any tasks that are not compatible with Lumen.

Furthermore, bear in mind that Lumen is extendable and you can cherry-pick the Illuminate components you need so this could actually break a valid Lumen application.

I think the best thing to do for Lumen is to simply override the Laravel recipe and provide a more appropriate default `deploy` task. The rest shouldn't need to change.

PR #2414, opened 2021-02-18
Bugfix shopware recipe plugin updates

rickschippers:
- [X] Bug fix. - Shopware recipe plugin activation/migration fix.
- [X] New feature - Upgrades upgradeable plugins
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [X] Changelog updated?

The shopware recipe parses output of the `plugin:list` command. This parsing is incorrect, since the version column is incorrectly duplicated. Because of this activation and db migration of plugins doesn't work all the time. I fixed the column definition to fix this issue.

The `sw:plugin:activate:all` task tries to run the `sw:plugin:refresh` but it does this incorrecty by using `task` instead of `invoke`, so it actually doesn't run the task. Changed that to `invoke` so it is actuall ran.

I Also added a `sw:plugin:upgrade:all` task that runs a `plugin:update` command on all plugins that have a new version available for update (upgradeable column set to Yes.). It uses the output of `plugin:list` the same way the other plugin commands work.

TLDR
With this fix applied plugins will be activated, migrated and updated successfully on deploy.


peterjaap:
@Schrank and I have discussed this and have decided against this.

As you can see [here](https://github.com/shopware/platform/blob/2f0f1379273f6869d20e5e5032f720aae746d80c/src/Core/Framework/Plugin/PluginLifecycleService.php#L279), the `plugin:update` command does a `composer update` call. This brings in new code - but you don't want to be introducing new code in your _deployment stage_; updating plugins should be done locally, after which they are locked in `composer.lock`.

Even if you were to update plugins during deployment, you'd have to commit to git too, because otherwise the code wouldn't be available on local machines. So in all, this breaks a standard development and deployment flow.

Schrank:
@rickschippers Thank you very much for the bug fix! I'm all on for all the changes except the last one. Having an update command might be a cool thing, especially if one can run it local (which I don't have a clue how to, but deployer works locally, so there must be a way 😅)

That said I agree with @peterjaap, updating plugins on deployment sounds pretty dangerous and I opt against. So having the command is cool and if people want to run it on deployment, I think it is a bad idea, but I'm fine with that 😂, but let's put it in the local config and not the default behavior! <3

rickschippers:
Ok two things to consider - this pr also contains a couple of bugfixes for the processes that are already built and don't work properly. So please check those. I can create a seperate pr for those (without the plugin:update) if you want.

Then the plugin:update. We build our own plugins. Currently we have those in the custom/static-plugins dir which is configured as a static repo in composer (which symlinks them into vendor). When I update a version in the plugin, run the composer update locally. And commit, including the updated composer.lock. The plugin will stay on the old version after deployment. 

If you do not run the plugin:update command, the plugin's update method will not be run. I use this method e.g. to create properties or custom fields in the database for example. Any code in the update method will not be run if you skip this method.

peterjaap:
@rickschippers could you create a separate PR for the fixes?

Concerning the `plugin:update`, what you're talking about (the creating properties & custom fields) is part of the migration task, which indeed is run by `plugin:update` (see https://github.com/shopware/platform/blob/2f0f1379273f6869d20e5e5032f720aae746d80c/src/Core/Framework/Plugin/PluginLifecycleService.php#L338) but _also_ by `database:migrate --all <plugin>`, which is run by the Deployer task `sw:plugin:migrate:all` (see https://github.com/deployphp/deployer/blob/bd7e90ad674ca2421334c450e835b8d454f58b6c/recipe/shopware.php#L95). So there is no need to run `plugin:update`.

rickschippers:
@peterjaap I'll make a separate pr for the fixes today.

In my case i'm using the PropertyGroup repository to create / update a property in the plugin's update method. So this is not run by the `databaase:migratate`. It really needs the `plugin:update` to execute those methods.

So in my plugin class I have these methods that are being run by `plugin:install` and `plugin:update` respectively:

```
public function install(InstallContext $installContext): void
{
    (new ProductTypePropertyInstaller($this->container))->install($installContext);
}

public function update(UpdateContext $updateContext): void
{
    (new ProductTypePropertyInstaller($this->container))->update($updateContext);
}
```
The ProducTypePropertyInstaller methods:

```  
public function install(InstallContext $context): void
{
    $this->upsertProductTypeProperty($context);
}

public function update(UpdateContext $context): void
{
    $this->upsertProductTypeProperty($context);
}
private function upsertProductTypeProperty(InstallContext $context): void
{
    $this->propertyGroupRepository->upsert([
            [
                'id' => ProductTypeService::PRODUCT_TYPE_PROPERTY,
                'name' => 'Product Type',
                'filterable' => false,
                'sorting_type' => PropertyGroupDefinition::SORTING_TYPE_ALPHANUMERIC,
                'display_type' => PropertyGroupDefinition::DISPLAY_TYPE_TEXT,
                'options' => [
                   // 8< Array with options which that can new options in plugin updates
                ]
            ]
        ]
        , $context->getContext()
    );
}
```

So basically the `plugin:install / update` methods run code to upsert a property in this case. This is not handled by database:migrate.

Schrank:
That sounds like we want to run `plugin:update` for all static plugins, because we know this code and more important - expect it to be run 😂 . And that makes total sense. @peterjaap What do you think about limiting the update to these?

Schrank:
P.S. @rickschippers Thanks for the explanation!

peterjaap:
@Schrank in theory that sounds good, but are we technically able to limit it to those?

rickschippers:
Maybe a simpler solution, where you define which plugin's should auto-update

Schrank:
That sounds like a good first step: provide a list with plugin names. To answer your question @peterjaap plugin:update can be limited, so we only™ need to parse all static modules directory names(?) to get the module name. Not even the composer.json file or the module definition php class! 😅

rickschippers:
PR for just the fixes is here: https://github.com/deployphp/deployer/pull/2419

antonmedv:
So what the status of this pr?

Schrank:
@rickschippers I removed the default upgrade step and resolved the conflicts. In this state I would love to merge it. Does it make you happy too?
Thanks for your patience!

cc @peterjaap

Schrank (approved):

rickschippers:
@Schrank sure, but it's basically the same as this now: https://github.com/deployphp/deployer/pull/2419.

I made a custom recipe that extends on the base shopware recipe for our situation in the mean time, that just runs the upgrade procedure for our own plugins, so that's a workable sitaution.

Schrank:
Ah, cool if that works for you :)

PR #2418, opened 2021-02-19
Fix for option without value

### LINKED ISSUE #2364: `--ansi` causes an error
- Deployer version: dev-master@ da432e01399489a896e880e9469c01ffad5491e6
- Deployment OS: Mac, Ubuntu 20.04

```bash
vendor/bin/dep deploy staging --ansi
```

Output is then:
```
 TypeError  in Stringify.php on line 67:

  Deployer\Support\Stringify::escape(): Argument #1 ($token) must be of type string, bool given, called in /Users/jkuchar1/Documents/www/grifartis/source/PHP/vendor/deployer/deployer/src/Support/Stringify.php on line 35

```

It looks like option --ansi is interpreted as `true` which then fails on escaping code which can handle only strings.

martijngastkemper:
- [x] Bug fix #2364?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [x] Tests added?
- [ ] Changelog updated?

I first created a failing test and tried to solve it the problem in `Stringify.php`. Something like this:

```php
foreach ($option as $value) {
     $value = is_string($value) ? self::escape($value) : '';
     $options[] = "--$name " . self::escape($value);
}
```

But I think `Stringify::options` can be replaced by Symfony `ArrayInput` class ([docs](https://symfony.com/doc/current/console/calling_commands.html)), because that's exactly doing what you want. But maybe I'm completely wrong.

### DIFF
--- a/test/src/Support/StringifyTest.php
+++ b/test/src/Support/StringifyTest.php
@@ -32,5 +32,10 @@ class StringifyTest extends TestCase
             new ArgvInput(['deploy', '-o', 'env=prod', '-l1'], $definition),
             new ConsoleOutput(Output::VERBOSITY_DEBUG, false)
         ));
+
+        self::assertEquals("--plan", Stringify::options(
+            new ArgvInput(['deploy', '--plan'], $definition),
+            new ConsoleOutput(Output::VERBOSITY_DEBUG, false)
+        ));
     }
 }


Schrank:
@martijngastkemper Thanks for the fix! Would you mind fixing the test?

antonmedv:
This was fixed in another pr.

PR #2419, opened 2021-02-19
Shopware recipe - fix invalid parsing of plugin:list and invoke nested sw:plugin:refresh instead of redefining it

rickschippers:
- [x] Bug fix?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [x] Changelog updated?

- This fixes incorrect parsing of the `plugin:list` command. Removed an incorrect duplicate version column.
- Invoke the nested `sw:plugin:refresh` task instead of redefining it, so that it actually runs.


peterjaap:
Thanks! :partying_face:

PR #2422, opened 2021-02-19
Normalize shopware recipe and add common.php

Schrank:
- [ ] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [X] Changelog updated?


docs/recipe/shopware.md
```diff
@@ -34,26 +36,36 @@
 ### repository
 [Source](https://github.com/deployphp/deployer/search?q=%22repository%22+in%3Afile+language%3Aphp+path%3Arecipe+filename%3Ashopware.php)
 
+* Overrides [`repository`](/docs/recipe/common.md#repository) from `recipe/common.php`
```
Schrank, line 39: @peterjaap maybe remove set('repository')?

peterjaap: Yeah possible, but I'd rather leave it in to remind people of what to do (for those less familiar with Deployer)

peterjaap:
Why is this necessary? Isn't common automatically loaded by Deployer?

Schrank:
Good question, need to check.

I assumed no and this recipe is the base for `dep init` For shopware and therefore should require it. Will check and if deployer takes care of it remove it from all other recipes 😂

peterjaap:
I check my installs and we usually do this;

```
<?php
namespace Deployer;

require 'vendor/deployer/deployer/recipe/common.php';
require 'vendor/deployer/deployer/recipe/shopware.php';
```

So adding it through `require_once` should be backwards compatible with that.

antonmedv:
So, should we merge it?

Schrank:
I’ll have a look and will take care of it :/)

Schrank:
common.php is needed and not auto-included by deployer.

recipe/shopware.php
```diff
@@ -4,6 +4,8 @@
 
 use MJS\TopSort\Implementations\FixedArraySort;
```
antonmedv, line 5: Maybe use independent implementation? And recipe can work without autoloader.

Schrank: Great idea. I never implemented a topological sorting :) And I think as long as no one is complaining it has low priority :)

PR #2425, opened 2021-02-20
Feature/2423 shopware order modules

### LINKED ISSUE #2423: shopware: task sw:plugin:activate:all runs migration script in wrong order
If I deploy I have migration scripts which connect languages with sales_channels. Unfortunately is my module the first in the list and therefore called first. I added the dependency into the composer.json but of course this is ignored.

Solution (I'm goin to implement now) is, to parse all composer.json files and calculate the correct order to activate/update the modules

Schrank:
- [X] Bug fix #2423 ?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [X] Changelog updated?

This PR reads the dependency infos of shopware modules from composer.lock and sorts the modules according to the `require` dependencies before activating/migration them.

@peterjaap please have a look for the content
@antonmedv I See no way around marcj/topsort, but for PHP8 I only need `str_ends_with`, so in theory we could implement this method on our own and remove the dependency on polyfill80.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@
 ### Fixed
 - Lots, and lots of long-standing bugs.
 - Fixed incorrect plugin:list parsing (remove duplicate version column). Invoke nested sw:plugin:refresh task instead of redefining it, so that it actually runs.
+- Shopware activates/runs migration in order (respects dependencies in composer.json). [#2423] [#2425]
 
 
 ## v6.8.0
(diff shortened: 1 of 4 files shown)


Schrank:
I'm sooo happy...

```
[stage.winkelwagen.de] info successfully deployed!
```

Works - once you pushed the updated composer.lock :D

peterjaap:
Awesome!

PR #2429, opened 2021-02-22
Fix shared configuration file

### LINKED ISSUE #2413: Recipe Magento2 first deploy
- Deployer version: 6.8.0
- Deployment OS: Ubuntu 20.04 LTS / WSL 2 / Windows 10

With the magento2 deploy recipe, I cannot find out how to perform the initial installation. It seems like the recipe can only handle upgrades, no setup:install, so how should the first install be performed on the host?

The Production server was installed with all the required packages, databases and settings that are needed to run a successful magento install. Then the SSH was setup to allow access from the deployer server. Then the following was tried on the deployer server, without success.

Trying `dep deploy host` on the empty directory returns an `Invalid configuration file: /var/www/my-project/releases/1/app/etc/env.php` error during the magento setup:di:compile. This is probably related to #2377 and more closely to #2106, however, I cannot find a solution in those issues.

I also tried to add a deploy_install task replacing the deploy:magento line from the original deploy task with a deploy:magento:install task, which is effectively calling the magento setup:install with all the required parameters. However, on an empty directory, the `dep deploy_install host` returns a `The command "readlink /var/www/my-project/current" failed`. The code added to the deploy.php file to do the setup:install:
```
desc('Magento2 deployment for install');
task('deploy:magento:install', function () {
        run("{{bin/php}} {{release_path}}/bin/magento module:enable --all");
        run("{{bin/php}} {{release_path}}/bin/magento setup:install --admin-firstname={{magento_adminfirstname}} --admin-lastname={{magento_adminlastname}} --admin-email={{magento_adminemail}} --admin-user={{magento_adminuser}} --adminpassword={{magento_adminpassword}} --db-host={{magento_dbhost}} --db-name={{magento_dbname}} --db-user={{magento_dbuser}} --db-password={{magento_dbpassword}}");
});

desc('Deploy your project first time');
task('deploy_install', [
    'deploy:prepare',
    'deploy:vendors',
    'deploy:clear_paths',
    'deploy:magento:install',
]);
```

The git repository consists of simply the magento2 composer.json that was retrieved from the composer create-project --noinstall command. Should there be more files included in the Git repo?

How do you perform the first deploy on a freshly installed host?

memen45:
According to the [magento2 devdocs](https://devdocs.magento.com/guides/v2.4/config-guide/prod/config-reference-sens.html):
*  The [`magento app:config:dump` command](https://devdocs.magento.com/guides/v2.4/config-guide/cli/config-cli-subcommands-config-mgmt-export.html) writes system-specific settings to the system-specific configuration file, `app/etc/env.php`, which should _not_ be in source control. It also writes shared configuration for all Magento instances to `app/etc/config.php`, this file _should_ be in source control.

Therefore the only change is the shared file 'app/etc/env.php' is removed and replaced by the shared file 'app/etc/config.php'.
- [x] Bug fix #2413
- [x] New feature: no
- [x] BC breaks: no
- [x] Deprecations: no
- [x] Tests added: no 
- [ ] Changelog updated: no

      $ php bin/changelog

### DIFF
--- a/recipe/magento2.php
+++ b/recipe/magento2.php
@@ -24,7 +24,7 @@ set('content_version', function () {
 });
 
 set('shared_files', [
-    'app/etc/env.php',
+    'app/etc/config.php',
     'var/.maintenance.ip',
 ]);
 set('shared_dirs', [


peterjaap:
The shared files are not in version control, they exist solely on the production server and are therefore by definition _shared_. If we would merge this, the config file would be shared on the server between releases - which is exactly what we don't want.

PR #2443, opened 2021-03-05
Fix formatting

sprankhub:
- [ ] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [ ] Changelog updated?

      $ php bin/changelog


sprankhub:
I do not think this is worth a changelog entry.

TBH, I do not get why `docgen` fails... If I run it, the fixed issue is introduced again. Could you please help, @antonmedv?

sprankhub:
Got it, should be fixed now :)

Issue #2439, opened 2021-03-02
[Deployer 7] Selector is lacking support for arrays


ostrolucky:
deploy.php:
```php
task('foo', function() {
    info('yes!');
})->select('roles=foo');

localhost()
    ->set('labels', ['roles' => ['foo', 'bar', 'baz']]);
```

```
vagrant@shopping:/vagrant/deployer-example$ php vendor/bin/dep foo -vvv
task foo
 TypeError  in Selector.php on line 67:

  Deployer\Selector\Selector::compare(): Argument #2 ($a) must be of type ?string, array given, called in /vagrant/deployer-example/vendor/deployer/deployer/src/Selector/Selector.php on line 58

#0 /vagrant/deployer-example/vendor/deployer/deployer/src/Selector/Selector.php(58): Deployer\Selector\Selector::compare()
#1 /vagrant/deployer-example/vendor/deployer/deployer/src/Executor/Master.php(90): Deployer\Selector\Selector::apply()
#2 /vagrant/deployer-example/vendor/deployer/deployer/src/Command/MainCommand.php(119): Deployer\Executor\Master->run()
#3 /vagrant/deployer-example/vendor/symfony/console/Command/Command.php(256): Deployer\Command\MainCommand->execute()
#4 /vagrant/deployer-example/vendor/symfony/console/Application.php(971): Symfony\Component\Console\Command\Command->run()
#5 /vagrant/deployer-example/vendor/symfony/console/Application.php(290): Symfony\Component\Console\Application->doRunCommand()
#6 /vagrant/deployer-example/vendor/symfony/console/Application.php(166): Symfony\Component\Console\Application->doRun()
#7 /vagrant/deployer-example/vendor/deployer/deployer/src/Deployer.php(289): Symfony\Component\Console\Application->run()
#8 /vagrant/deployer-example/vendor/deployer/deployer/bin/dep(105): Deployer\Deployer::run()
#9 {main}
```

This makes Deployer 7.x unusable for us. We used to do
```php
task('foo', function() {
    info('yes!');
})->onRoles('foo');

localhost()
  ->roles('foo', 'bar', 'baz');
```

and now there doesn't seem to be alternative

ostrolucky:
[...]

PR #2444, opened 2021-03-05
Make selector accept array values

ostrolucky:
Fix for https://github.com/deployphp/deployer/issues/2439

### DIFF
--- a/src/Selector/Selector.php
+++ b/src/Selector/Selector.php
@@ -62,13 +62,26 @@ class Selector
         return false;
     }
 
-    private static function compare(string $op, ?string $a, ?string $b): bool
+    /**
+     * @param string|array $a
+     */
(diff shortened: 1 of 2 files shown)


Schrank:
Would you mind adding a line to the changelog please? Happy to do it myself, but I like if you are on the git annotate <3  

>  Give credit where credit is due

ostrolucky:
I think this is included under https://github.com/deployphp/deployer/blob/27c7b6182c77ed69059bffee8b2e288526ccde67/CHANGELOG.md#L18

We don't have changelog documenting changes between beta versions, do we? Or what should I write where?

antonmedv:
Awesome let’s merge it.  No need to add to changelog. Also will be cool if someone can implement array support for selector expressions.

Issue #2445, opened 2021-03-06
[Deployer 7] GroupTask via invoke() is ignored


ostrolucky:
- Deployer version: 1669ffb11d6b176dbc00bebf915af1327d6f557d

This works in Deployer 6.x, but not 7.x

```php
task('baz', function() {
    info('baz');
});
task('quix', function() {
    info('quix');
});
task('bar', ['baz', 'quix']);

task('foo', function() {
    invoke('bar');
});

localhost();
```

Output:
```
task foo
task bar
```

Expected output:
```
task foo
task bar
task baz
[localhost] info baz
task quix
[localhost] info quix
```

antonmedv:
Are you sure what it work in v6?  Invoke was just call to one task, no hooks and no group tasks. But as v7 is still going on we can add support for group tasks.

ostrolucky:
yep i'm sure

antonmedv:
Can somebody create a pr with fix?

ostrolucky:
PR created

PR #2446, opened 2021-03-06
Fix execution of group tasks via invoke()

ostrolucky:
- [x] Bug fix https://github.com/deployphp/deployer/issues/2445
- [x] Tests added?

### DIFF
--- a/src/Selector/Selector.php
+++ b/src/Selector/Selector.php
@@ -63,7 +63,7 @@ class Selector
     }
 
     /**
-     * @param string|array $a
+     * @param string|string[] $a
      */
     private static function compare(string $op, $a, ?string $b): bool
     {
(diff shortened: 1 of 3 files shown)


test/joy/DeployTest.php
```diff
@@ -17,7 +17,7 @@ public function testDeploy()
     {
         $this->init(self::RECIPE);
         $this->tester->run([
-            'deploy',
+            'deploy:invoke',
```
antonmedv, line 20: Better to create separate test for it. As normal deploy test is also needed. Or even better - unit test for this part of the code.

antonmedv (changes requested):

antonmedv (approved):

ostrolucky:
Thanks for quick merge. Can we have new beta release soon? 🙏

PR #2486, opened 2021-04-03
Add and use new release_or_current_path

lorisleiva:
As discussed in PR https://github.com/deployphp/deployer/pull/2485, This PR enables users to run tasks outside a deployment by providing a new `release_or_current_path` option.

This option will always return the `release_path` if it exists and fallback to the `current_path` if it doesn't.

That way, outside of deployment, we can still run `dep my:task` and expect our task to be executed on the latest active release.

This PR replaces the use of `release_path` with `release_or_current_path` in all the custom tasks of platform-specific recipes — like laravel, symfony, etc. It also replaces the use of `release_path` in the `deploy:vendor` task, allowing users to run `composer install` on the current release if they need to.

Let me know if there are other places that would benefit from this new path.

Let me know what you think. 🙂 

---

- [ ] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks? No, the existing behaviour stays the same but we're adding the possibility to also run tasks directly to the current release.
- [ ] Deprecations?
- [ ] Tests added?
- [x] Changelog updated?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 - Added support for PHP8.
 - Added slack_channel option to Slack recipe.
 - Chatwork contrib recipe.
+- Added `release_or_current_path` option that fallbacks to the `current_path` when the `release_path` does not exist. [#2486]
 
 ### Changed
 - Refactored executor engine, up to 2x faster than before.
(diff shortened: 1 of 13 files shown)


antonmedv:
Awesome! Thanks for also fixed other recipes as well! Will release a new beta right now.

PR #2487, opened 2021-04-03
Add php-fpm as a contrib recipe

lorisleiva:
Hi again 😊 

One very common task to add to PHP deployments is reloading the `php-fpm` service at the very end.

I was wondering if you'd consider the following `contrib` recipe that makes it slightly easier for users to reload PHP-fpm from their deployments.

It has the following configurations:
- `php_fpm_version` – The PHP-fpm version. For example: `8.0`.
- `php_fpm_service` – The full name of the PHP-fpm service. Defaults to `php{{php_fpm_version}}-fpm`.
- `php_fpm_command` – The command to run to reload PHP-fpm. Defaults to `echo "" | sudo -S /usr/sbin/service {{php_fpm_service}} reload`.

By default, `php_fpm_version` throws an Exception. This is to ensure users explicitly provide their version instead of relying on a default value that might change at any moment.

Let me know what you think. 🙂 

---

- [ ] Bug fix #…?
- [x] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [x] Changelog updated?


antonmedv:
I like this recipe really much. Especially documentation. Nice work!

contrib/php-fpm.php
```diff
@@ -0,0 +1,41 @@
...
+```php
+set('php_fpm_version', '8.0');
+after('deploy', 'php-fpm:reload');
+```
+
+ */
+namespace Deployer;
+
+set('php_fpm_version', function () {
+    throw new \Exception('Please provide the PHP-fpm version using the `php_fpm_version` option.');
```
antonmedv, line 33: What adding detection of php-fpm version installed? And if two or more installed only then throw an exception?

lorisleiva: Good idea! I've managed to do it by looking at the running processes so even if multiple PHP-FPM versions exist, we default to using the one actually being used. 🙂

lorisleiva: Thank you for the invitation to the Deployer organization btw! 😊

PR #2488, opened 2021-04-04
Tidy artisan commands

lorisleiva:
This PR aims to keep the artisan tasks from the Laravel recipe up-to-date with the framework. It also cleans any redundant or obsolete tasks. Here's the complete changelog.

- **Removed** task `deploy:public_disk`. This task was redundant since this can be achieved with the `artisan:storage:link` task since 5.3. 5.2 support was dropped in June 2016 so it is safe to stop supporting it ourselves.
- **Removed** tasks that published assets. Namely: `horizon:assets`, `horizon:publish`, `telescope:publish` and `nova:publish`. These commands are made to generate some code in our local environment that should then be committed and pushed. These should never be needed directly in production.
- **Removed** the `min` and `max` constraints on the `optimize` and `optimize:clear` tasks. Whilst its implementation changed slightly at some point, the command was actually always available. I believe this was a mistake on my side.
- **Added** the following artisan tasks to stay consistent with what we provide. For example, if we provide `config:cache` and `config:clear`, we should also provide `route:cache` and `route:clear`.
  - `route:clear`
  - `route:list`
  - `horizon:clear`
  - `horizon:list`
  - `event:list`
- **Added** the following artisan tasks that could be of great benefit for handling queue jobs in production.
  - `horizon`
  - `horizon:continue`
  - `horizon:pause`
  - `horizon:purge`
  - `horizon:status`
  - `queue:failed`
  - `queue:flushed`
- **Changed** the order in which tasks are defined. First, by grouping them into categories, then alphabetically.

This PR breaks backward compatibility but I thought since we're still in Beta, it was the perfect opportunity for a cleanup. 😊 

---

- [ ] Bug fix #…?
- [ ] New feature?
- [x] BC breaks? Yes, some tasks have been deleted but they were not meant to be used directly in production.
- [ ] Deprecations?
- [ ] Tests added?
- [ ] Changelog updated?


antonmedv:
Wow! Awesome work! And totally perfect timing to introduce BC. Will release a new beta right now.

lorisleiva:
Thanks! :) Hang on, I forgot to update the changelog, I'm creating another quick PR.

PR #2491, opened 2021-04-05
Feature/add shared configs for common template only

lorisleiva:
Hi 👋 

I'm just writing an article on how to get started with Laravel and Deployer 7 and I've noticed that I've got to tell the readers to remove a bunch of options that conflict with the default values provided by the Laravel recipe.

Capture 2021-04-05T15 35 44@2x

This is because, I'm using the `YAML` format in that tutorial and it cannot make use of `add('shared_files', [])` like the `deploy.php` file can.

When using project template like the one provided by Laravel, these values are already set up with some good defaults inside the recipe file. But when adding the `shared_files`, `shared_dirs` and `writable_dirs` options to the YAML file it will override these values with something irrelevant to the type of project. E.g. here Laravel does not have an `uploads` folder. 

This PR fixes this by removing these values from the `deploy.yaml` template **unless** the provided template was `common`.

What do you think?

---

- [ ] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [x] Changelog updated?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@
 - Refactored `dep init` command.
 - Normalize shopware recipe (require common.php).
 - Removed the `min` and `max` constraints on the `artisan:optimize` and `artisan:optimize:clear` tasks. [#2488]
+- Excluded the `shared_files`, `shared_dirs` and `writable_dirs` configs from the `deploy.yaml` default template unless the `common` template was chosen.
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


antonmedv:
I think this is good solution for now. Yaml recipes lack add config support.  Maybe something similar should be added. Like - add: or inheritance via - @parent.

antonmedv:
Released beta.19 with this changes.

PR #2498, opened 2021-04-10
php 7.4 compatibility

fliespl:
Couldn't find info if deployer 7 will be php8+ only or also php 7.4, but I would assume it's 7.4.

Hence this PR

- [x ] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [ ] Changelog updated?

      $ php bin/changelog

### DIFF
--- a/src/Support/Config/verification.php
+++ b/src/Support/Config/verification.php
@@ -9,7 +9,7 @@ task('verification', [
 
 task('verification:timeouts', function () {
     try {
-        run("php -r 'while(true){}'", timeout: 1);
+        run("php -r 'while(true){}'", ['timeout' => 1]);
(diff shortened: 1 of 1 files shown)


antonmedv:
Deployer 7 will support php8 and php7.4. But this change isn’t needed as it only run by me.

fliespl:
Gotcha. 

I have a feeling though that since you support php 7.4, example deploy.yaml functions should follow this requirement and not use php8.0. Since it's only run by you, maybe this shouldn't be in repository at all, but rather moved to tests?

antonmedv:
I didn’t find better place for it)

Issue #2497, opened 2021-04-09
upload/download - return output


fliespl:
Sometimes it might be needed to get only final stats and not progress of rsync command.

Currently we can only do it by using verbose flag which is a bit noisy + changing upload configs flags to -az and adding --stats to options (or alternatively use rsync contrib recipe which also would need some code overwriting since it doesn't return run / runLocally output as well).

```
    upload(__DIR__ . "/bin/", '{{release_path}}', [
        'options' => ['--stats'],
        'flags' => '-az'
    ]);
```

If would be great if upload / download were returning final output so anyone can use it as desired if needed.

antonmedv:
Nice idea! Or maybe as a configurable option? Can you create a PR for it?

fliespl:
Sure, will try to figure it out over the weekend.

Schrank:
May I assume this ticket is fixed @fliespl ? <3 If not please tell me, then I'll reopen.

Thanks for the PR!

fliespl:
@Schrank that's correct :) And you are welcome!

PR #2500, opened 2021-04-10
Rsync stats config option

fliespl:
PR for #2497 

- [ ] Bug fix #…?
- [X ] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [ ] Changelog updated?

PR for rsync stats. I have target php 7.4, but if you prefer 8.0 `strpos` can be replaced with `str_starts_with`

### DIFF
--- a/src/Utility/Rsync.php
+++ b/src/Utility/Rsync.php
@@ -35,7 +35,9 @@ class Rsync
      *
      * - `flags` for overriding the default `-azP` passed to the `rsync` command
      * - `options` with additional flags passed directly to the `rsync` command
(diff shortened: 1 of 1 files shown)


src/Utility/Rsync.php
```diff
@@ -46,11 +47,17 @@ public function call(Host $host, $source, string $destination, array $config = [
             'options' => [],
             'flags' => '-azP',
             'progress_bar' => true,
+            'display_stats' => false
```
antonmedv, line 50: Also, please, add comment to docblock.

PR #2501, opened 2021-04-11
Fix ACL error message

lorisleiva:
I came across the "ACL missing" error message the other day and realised that the suggestion to fix it does not work.

Capture 2021-04-08T17 46 35@2x

According to the `dep run` signature, we should first provide the command as *one* argument before providing optional selectors.

Capture 2021-04-11T13 34 59@2x

Therefore, running **`dep run prod -- ls -la`** throws an error. ❌

Capture 2021-04-11T13 31 13@2x

But, running **`dep run 'ls -la' -- prod`** (or just `dep run 'ls -la' prod`) works fine. ✅ 

Capture 2021-04-11T13 32 35@2x

This PR, fixes the suggestion of the "ACL missing" error.

I can't seem to find any other error messages that suggest to run `dep run` but please let me know if I've missed one and I'll update it as well.

---

- [x] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [x] Changelog updated?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@
 - Shopware activates/runs migration in order (respects dependencies in composer.json). [#2423] [#2425]
 - Boolean options should not go through the `self::escape` function. [#2392]
 - Check if shared file exists before touching it (and fail because no write permission). [#2393]
+- Fixed "dep run" suggestion in ACL error message. [#2501]
 
 ### Removed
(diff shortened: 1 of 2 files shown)


antonmedv:
Nice catch! Also I’m thinking what host first command second is better for run command. But selector is array. Maybe actually fix run command? And allow to specify as in old error message?

lorisleiva:
That's a good point. I do find `dep run pros -- ls -la` more elegant. ☺️

PR #2507, opened 2021-04-16
Fix .env for Symfony recipe

### LINKED ISSUE #2506: .env.local.php in Symfony recipe
Hi !

I'm a bit confused with the .env.local.php value in the shared_files of the Symfony recipe (and regarding #1907 and #1923 i am probably not the only one :p)

For me, this file should not be include here, as it have to be generate using the symfony:dump-env command (which is part of Symfony/Flex, see https://symfony.com/doc/current/configuration.html#configuring-environment-variables-in-production) in order to compile the .env file which is versioned, with the .env.local which is stored on the server.

The problem here is that if you add a new bundle requiring a new env value, deploy:vendors will fail as the .env.local.php will not include it even if you already added it to your .env.local, because .env.local.php will be called first. And it is not possible to generate a new .env.local.php as the vendors aren't installed yet.

For me, the best solution would be to remove the .env.local.php file from the shared_files, create a deploy:dump-env task and call it just after the deploy:vendors task

victor-paumier:
- [x] Bug fix #2506 
- [X] New feature
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [X] Changelog updated

      $ php bin/changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@
 - Check if shared file exists before touching it (and fail because no write permission). [#2393]
 - Fixed "dep run" suggestion in ACL error message. [#2501]
 - TypeError, port is int on escapeshellarg. [#2503]
+- .env.local.php in Symfony recipe. [#2506]
 
 ### Removed
 - Removed the `artisan:public_disk` task. Use the `artisan:storage:link` task instead. [#2488]
@@ -615,6 +616,7 @@
 - Fixed `DotArray` syntax in `Collection`.
 
 
+[#2506]: https://github.com/deployphp/deployer/issues/2506
 [#2503]: https://github.com/deployphp/deployer/issues/2503
 [#2501]: https://github.com/deployphp/deployer/pull/2501
 [#2488]: https://github.com/deployphp/deployer/pull/2488
--- a/recipe/symfony.php
+++ b/recipe/symfony.php
@@ -13,11 +13,10 @@ set('symfony_version', function () {
 
 set('shared_dirs', [
     'var/log',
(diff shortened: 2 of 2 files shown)


Schrank:
Awesome! <3

victor-paumier:
Also, is `cd {{release_or_current_path}} &&` really necessary here?https://github.com/deployphp/deployer/blob/f0ac01a24801b17fcf1857360223e40c64244c40/recipe/symfony.php#L43

https://github.com/deployphp/deployer/blob/f0ac01a24801b17fcf1857360223e40c64244c40/recipe/symfony.php#L30

antonmedv:
Please, run docgen.

victor-paumier:
> Please, run docgen.

I have tried but the result does not look good. Do you know why?

Capture d’écran 2021-04-16 à 10 52 12

antonmedv:
Hmm, can't tell. I'll try to figure this out by my self.

antonmedv:
Yes, there was a bug is docgen in parse regexp of task names (no `-`).

Schrank:
> Also, is `cd {{release_or_current_path}} &&` really necessary here?
> 
> https://github.com/deployphp/deployer/blob/f0ac01a24801b17fcf1857360223e40c64244c40/recipe/symfony.php#L43
> 
> 
> As it seems already defined here:
> https://github.com/deployphp/deployer/blob/f0ac01a24801b17fcf1857360223e40c64244c40/recipe/symfony.php#L30

I think it is needed. The console might be called with an absolute path, but you need to be in the application directory so the console works correctly.

PR #2510, opened 2021-04-16
Fix regex that identifies "cd" commands in YAML scripts

### LINKED ISSUE #2509: Yaml - Array Syntax vs String Syntax broken
- Deployer version: 7beta20
- Deployment OS: GitHub Actions

It appears after a [discussion](https://github.com/deployphp/deployer/discussions/2508) that using the array notation for scripts.

```
  yarn:run:prod:
    script:
      - 'cd {{release_or_current_path}} && yarn run prod'
```

Does not run properly vs something like:

```
  yarn:run:prod:
    script: 'cd {{release_or_current_path}} && yarn run prod'
```

lorisleiva:
When providing an array of string to the `script` key inside a YAML task, each line goes through a regex that identifies whether its a `cd` command or not. That way, the following script array:

```yaml
  yarn:run:prod:
    script:
      - 'cd {{release_or_current_path}}'
      - 'yarn run prod'
```

becomes:

```php
cd('{{release_or_current_path}}');
run('yarn run prod');
```

However, the regex did not take into account that multiple commands can be run within the same line like so.

```yaml
  yarn:run:prod:
    script:
      - 'cd {{release_or_current_path}} && yarn run prod'
```

Thus, the script array above became:

```php
cd('{{release_or_current_path}} && yarn run prod');
```

Making the task no execute anything.

This PR fixes this by adding an extra regex that looks for `\b&&\b` within the script line before wrapping it inside a `cd` function.

**EDIT**: I forgot to mention that the default `deploy.yaml` generated by the `dep init` command generates a task following this exact format which can cause newcomers to be confused when they don't see their tasks being executed.

---

- [x] Bug fix #2509
- [ ] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [x] Changelog updated?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@
 - Fixed "dep run" suggestion in ACL error message. [#2501]
 - TypeError, port is int on escapeshellarg. [#2503]
 - .env.local.php in Symfony recipe. [#2506]
+- Fixed regex identifying "cd" commands in YAML scripts. The regex was not taking into account that multiple commands can be executed within the same line — e.g. "cd {{release_path}} && npm run prod". [#2509]
 
 ### Removed
 - Removed the `artisan:public_disk` task. Use the `artisan:storage:link` task instead. [#2488]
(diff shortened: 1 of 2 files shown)


antonmedv:
Glad it was found and fixed before the release of v7. 

🎉 @lorisleiva

PR #2513, opened 2021-04-17
Isolated application tester for deployer E2E tests

bsacharski:
- [ ] Bug fix #…?
- [x] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [x] Changelog updated?

      $ php bin/changelog

I've created new class (`ConsoleApplicationTester`) to isolate the Deployer process from test executer process.

This class is responsible for allowing to run the Deployer inside separate process, but at the same time exposing methods responsible for interacting with STDIN and getting output from STDOUT, STDERR and getting Deployer process exit code.

This should allow for easier implementation of E2E tests, as they no longer will require workarounds for both tester and app being tested running in same space.


antonmedv:
Hi, thanks for you work again. Just remember what this was we cannot collect code coverage for e2e tests. Is there a way to support if here as well?

bsacharski:
> Hi, thanks for you work again. Just remember what this was we cannot collect code coverage for e2e tests. Is there a way to support if here as well?

I will look into this. I know that it's possible to force all php processes to dump code coverage information, so it would probably be just a matter of taking dumped codecov files, processing them and rendering a new report.

bsacharski:
> I will look into this. I know that it's possible to force all php processes to dump code coverage information, so it would probably be just a matter of taking dumped codecov files, processing them and rendering a new report.

And it's done in #2533

PR #2533, opened 2021-05-01
Added support for code coverage in E2E tests

bsacharski:
- [ ] Bug fix #…?
- [x] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [x] Changelog updated?

The E2E docker-compose stack has been modified to include code
coverage for all scripts that are kept in /project and are not
in the exclusion list.

The start command for `deployer` container has been replaced
with a `start-e2e-test.sh` script, which will execute E2E tests,
generate a Clover report and then return the exit code of E2E test
process to make sure that CI is aware of the test result.

Whole idea for collecting E2E coverage has been inspired by the
[this][1] article. All of scripts in `deployer` container will
now start with additional wrapper-like PHP code, which checks
if the `PHP_CCOV_START_FILE` variable is present. If not,
then nothing happens, but if the variable points to a PHP file,
then that file is being included.

The `docker-compose.yml` file has `PHP_CCOV_START_FILE` configured
to point to a file, which sets up code coverage, that starts counting
which instructions in `/project` directory have been called. Once
`php` interpreter starts shutting down, the coverage report is being
dumped into temporary files.

Once all tests have been processed, a script for merging partial
coverage reports is executed and finally a Clover code coverage file
is stored in the path defined in the `PHP_CCOV_OUTPUT_FILE` env var.

[1]: https://tarunlalwani.com/post/php-code-coverage-web-selenium/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 - Added tasks `artisan:key:generate` and `artisan:passport:keys` to the Laravel recipe.
 - Added the following artisan tasks: `artisan:route:clear`, `artisan:route:list`, `artisan:horizon`, `artisan:horizon:clear`, `artisan:horizon:continue`, `artisan:horizon:list`, `artisan:horizon:pause`, `artisan:horizon:purge`, `artisan:horizon:status`, `artisan:event:list`, `artisan:queue:failed`, `artisan:queue:flushed`. [#2488]
 - Isolated console application runner for E2E tests.
+- Support for code coverage in E2E tests.
 
 ### Changed
 - Refactored executor engine, up to 2x faster than before.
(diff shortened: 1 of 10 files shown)


antonmedv:
Awesome! Can you also write a short doc on how to use it?

bsacharski:
Im not sure there's much to describe:
- PHP is configured to automatically start coverage for ALL PHP processes and to end when PHP shuts down
- Once all tests are run, another script (`coverage-report.php`) takes all coverage data and generates `clover.xml`

I guess what would be now needed is exposing the `clover.xml` as a build artifact for further processing and having some next step that would take both `clover.xml` files from unit and E2E and generate final report.

PR #2534, opened 2021-05-02
Add Publish horizon assets

ahmadmayahi:
- [ ] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [ ] Changelog updated?

      $ php bin/changelog

### DIFF
--- a/recipe/laravel.php
+++ b/recipe/laravel.php
@@ -207,6 +207,8 @@ task('artisan:horizon:status', artisan('horizon:status', ['showOutput']));
 desc('Terminate the master supervisor so it can be restarted');
 task('artisan:horizon:terminate', artisan('horizon:terminate'));
 
+desc('Publish horizon assets');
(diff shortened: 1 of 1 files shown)


lorisleiva:
Hi there, can I ask why you’d need to generate assets from your deployment flow? This is the sort of command you typically run locally before committing changes. If you add that to your deployment process, the generated assets will not be committed to your repository.

lorisleiva:
Hi Adriaan 👋 

Thank you for sharing your use-case, I understand better now why these tasks could be beneficial to a deployment process.

However, like you said, there are quite a lot of packages and tools that require or allow us to publish their assets and so it could be tricky for Deployer to decide where to draw the line.

I feel it'd be more appropriate for these tasks to be defined as custom tasks as the use-case is also less common than some of the other support artisan tasks.

PR #2537, opened 2021-05-04
Add repository_dirs option to select diretories from repository to de…

leocsilva:
…ploy

- [ ] Bug fix #…?
- [ x ] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [ ] Changelog updated?

      $ php bin/changelog

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -50,6 +50,9 @@ set('keep_releases', 5);
 // Repository to deploy.
 set('repository', '');
 
+// Repository directories to deploy. Default [] (all directories)
+set('repository_dirs', []);
+
 // List of dirs what will be shared between releases.
 // Each release will have symlink to those dirs stored in {{deploy_path}}/shared dir.
 // ```php
--- a/recipe/deploy/update_code.php
+++ b/recipe/deploy/update_code.php
@@ -22,6 +22,9 @@ task('deploy:update_code', function () {
     $repository = get('repository');
     $branch = get('branch');
     $git = get('bin/git');
+    $repository_dirs = get('repository_dirs');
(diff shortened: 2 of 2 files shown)


antonmedv:
Hi, please, clarify why this code is needed. And what is doing, use cases.

leocsilva:
Hi, @antonmedv !

Consider that I want to deploy this repository: https://github.com/cirosantilli/test-git-partial-clone-big-small

However, I want to choose only the "small" directory to deploy.

With this proposed change, I would choose only the directories that I want to deploy. So my recipe would look like this:

<? php
namespace Deployer;

require 'recipe / common.php';

set ('application', 'test');
set ('deploy_path', '/home/test');
set ('repository', 'https://github.com/cirosantilli/test-git-partial-clone-big-small');
set ('repository_dirs', ['small']);
...

then, only the "small" directory will be deployed.

I hope it has become clearer. Thanks!

Baltazardoung:
👍 (with previous releases, I use git filter_branch for simulate this feature)


👍 (on my case with previous releases of 7.x, I use git filter_branch for simulate this feature)

holmey:
Hi @antonmedv I think it would be very helpful to exclude files and directories from deployments. This is currently already possible with the exclude option in the rsync deploy method. Each of our repositories contains quality control tools, configurations for local versioning (e.g. Dockerfiles), source files needed to build the application, the .git directory, documentation, etc.
All these files are not required in the directory where the current release is located. Apart from the fact that a lot of resources may be used to store and move the files back and forth, sometimes these files may not even be there.

PR #2553, opened 2021-05-20
add .gitattributes - exclude tests, docs from archives

iamvar:
Just added .gitattributes to exclude tests, docs from archives

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,8 @@
+/.gitattributes export-ignore
+/.github/ export-ignore
+/.gitignore export-ignore
+/docs/ export-ignore
+/phpcs.xml export-ignore
+/phpstan.neon export-ignore
+/phpunit.xml export-ignore
+/tests/ export-ignore


iamvar:
> What purpose for this in deployer repo?

As I mentioned above - to exclude tests and docs from archives.
Will make releases a little smaller and will be faster downloaded with composer install

antonmedv:
I think phar is better for Deployer =)

iamvar:
Yes, but those who use it as composer dependency for the dev environment will get benefits from this change.
Tell me your concerns why you do not want to include it? I do not understand.

You can read about .gitattributes here https://php.watch/articles/composer-gitattributes

iamvar:
Hello, so what do you think?
I do not see any negative effect from adding .gitattributes, but benefits for those who use deployer as composer dependency

iamvar:
Hello, so what do you think?
I do not see any negative effect from adding .gitattributes, but benefits for those who use deployer as composer dependency

Schrank:
Sounds to me likesomething useful. I use phar as well. But as long as the maintenance for "using deployer with composer" isn't killing us (this needs to be answered by @antonmedv), I think improving the experience is good.

PR #2554, opened 2021-05-23
Add Minimum PHP Version to Composer File

### LINKED ISSUE #2538: require php in composer.json
I don't see "php" in composer.json, why is that ?

https://github.com/spatie/package-skeleton-php/blob/master/composer.json#L18

amayer5125:
Closes #2538

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,6 +46,7 @@
         "bin/dep"
     ],
     "require": {
+        "php": ">=7.3",
         "marcj/topsort": "^2.0",
         "psr/http-message": "^1",
         "react/http": "^1",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "b35ae79e1110869b883a0ae6f3961566",
+    "content-hash": "72204ad7ad233b943ddfbe0296d93fea",
     "packages": [
         {
             "name": "evenement/evenement",
@@ -4433,7 +4433,9 @@
     "stability-flags": [],
     "prefer-stable": false,
     "prefer-lowest": false,
-    "platform": [],
+    "platform": {
+        "php": ">=7.3"
+    },
     "platform-dev": [],
-    "plugin-api-version": "1.1.0"
+    "plugin-api-version": "2.0.0"
 }


Schrank:
Im nur sure we want to change composer api version? @amayer5125

Schrank:
Wrong start, sorry, too early 🙈

thanks for the PR! 🥰

antonmedv:
But it will duplicate php check. In composer and in bin/dep.

amayer5125:
> Im not sure we want to change composer api version? @amayer5125

I added that without thinking about it. Probably should have brought it up in the PR. Composer 1.x [is being deprecated](https://blog.packagist.com/deprecating-composer-1-support/).

> But it will duplicate php check. In composer and in bin/dep

Should we remove the check in bin/dep or close this PR? In #2538 it seemed like you were saying we should add the php version back to the composer file.

antonmedv:
I'm not sure. Let's don't rush. We have a few other types of installation: phar and dist. And both of them even more popular than composer. So we definitely need check in bin/dep. But what purpose of adding such check in composer?

amayer5125:
Besides warning people when they are doing `composer require --dev deployer/deployer` that their php version isn't compatible I don't believe it does much else.

antonmedv:
Question: is Deployer in the master branch truly supports 7.3?)

amayer5125:
If I replace the named parameters in [tests/joy/recipe/named_arguments.php](https://github.com/deployphp/deployer/blob/master/tests/joy/recipe/named_arguments.php), all tests run green. Depending on the coverage of the tests, I would say that is a pretty good indicator that 7.3 is the minimum version.

smknstd:
https://getcomposer.org/doc/faqs/why-are-unbound-version-constraints-a-bad-idea.md

I think it would be good practice to change for something like :

```
"require": {
   "php": "^7.3|^8.0",
}
```

smknstd:
@antonmedv the bin/dep check is also unbound. I think it should also check for upper bound. For example, v6 should warn it doesn't work on php8, don't you think ? https://github.com/deployphp/deployer/blob/6.x/bin/dep#L13

antonmedv:
It will be cool to do. But it will require some effort on releasing. On v7 release process is automated.

kkrieger85 (approved):

Tobion (approved):

Schrank:
@antonmedv I think if people want to install deployer via composer (which is a horrible idea) we should help them by telling them what PHP version to use :-)

If you tell me what to do, I'm happy to take care of it.

antonmedv:
@Schrank it will be really cool.

antonmedv:
Added.

Issue #1786, opened 2019-01-08
Check for changes in directory


sweebee:
I'm using deployer for a lot of Laravel web applications. I also build the front-end with ```npm ci; npm run prod```. The problem with this is that it can take quite some time while most of the time, I did not change anything in the front-end code that needs to be builded.

So I found a solution for this, why not check for changes in specific folders and only build if there are any. This simple solution works perfectly.

It checks if there are changes in a specific folder based on the previous commit.

```php
// Which folders to check if need to build front-end
set('asset_dirs', [
	'resource/assets'
]);

// Build assets
task('npm', function () {
	$should_build = false;
	foreach(get('asset_dirs') as $asset){
		if(exec('git diff --name-only HEAD~1..HEAD ' . $asset)){
			$should_build = true;
		}
	}
	if($should_build) {
		run( 'npm ci; npm run prod' );
	}
});

```

Would be a nice feature to include into the core so you don't have to add this code manually.

antonmedv:
Awesome! Will think about it more. But `git diff --name-only HEAD~1..HEAD ` part maybe difficult. Why only one commit `~1`?

sweebee:
Well i think it depends on how you deploy. I deploy on every commit and for production on a merge request from dev > prod.

But i think it will be saver to look at some more commits. Also make sure you deploy the assets to a shared directory.

kanevbg:
I will add there this addition there should be a general-purpose check enabled by default - if the last release uses the same commit as the one requested for deployment than skip updating (if not in forced mode)

antonmedv:
I find this is really difficult to implement. If someone whats to implement in as contrib recipe - your are well come.

sbrow:
What makes this so difficult to implement? The way I see it, all that's needed is to add a property to tasks that contains the files to watch, and before each task is run, check all the listed files against the previous release. If none of them changed, skip the task. 

e.g.
```php
// src/Task/Task.php
<?php
namespace Deployer\Task;

class Task
{
   // An array of globs to check before running the task.
    public array $watch = [ '*' ];

    // ...
}

Schrank:
@sbrow Thanks for your comment and I think if you put all the configuration on the user, you are right. One needs to make sure, that files which are not on git are ignored for example.

I can only repeat Anton: Feel free to volunteer and implement it! Happy to review your PR and if you have questions feel free to ask!

PR #2595, opened 2021-06-30
Feature: Skip tasks if files unchanged.

sbrow:
This PR Allows tasks to be skipped if their dependencies haven't changed, As mentioned in issue #1786.

Examples:
- [ ] Skip `deploy:vendor` if `composer.json` and `composer.lock` haven't changed.
- [ ] [Hard] Skip `yarn:build` if `package.json`, `yarn.lock`, files have changed, also `webpack.config.js`, and files referenced in `package.json`.

Todos:
- [x] Add `$watch` property to `Task`, containing list of globs for files that should be watched.
- [ ] Before running a `Task`, check that none of the `$watch`ed files have changed since the last release.
- [ ] Never skip tasks during the first release.
- [ ] Add text/icon indicator to tasks in CLI, to inform the user that a Task was skipped.
  - [ ] Also show the file(s) that triggered the change?

### DIFF
--- a/src/Executor/Worker.php
+++ b/src/Executor/Worker.php
@@ -35,7 +35,11 @@ class Worker
 
             $context = new Context($host, $this->deployer->input, $this->deployer->output);
             $context->setIsLocal($task->isLocal());
-            $task->run($context);
+            if ($task->shouldSkip($context)) {
+                $task->skip($context);
+            } else {
+                $task->run($context);
+            }
 
             if ($task->getName() !== 'connect') {
                 $this->deployer->messenger->endOnHost($host);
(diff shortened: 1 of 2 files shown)


Schrank:
Awesome! Small hint, after reading the description - normally we don't add vendor to the git repo, so composer install needs to run always, at least, that is best composer practice

benoitchantre:
I think Make would be the appropriate tool for that. You need to redefine deploy:vendor task to use make.

antonmedv:
Can such check be on top of task function body?

Schrank:
Good idea. It can be implemented as an extension to task()->select()

Schrank:
Or maybe better another method, to not puzzle everyone with the syntax 😅

Issue #2568, opened 2021-06-05
`download()` rsync options not being passed on


bilogic:
- Deployer version: Deployer 7.0.0-beta.23
- Deployment OS: centos-release-8.2-2.2004.0.2.el8.x86_64

```PHP
    download('{{ deploy_path }}/shared', '', [
        'exclude' => [
            '**/storage/app/queue/**',
        ],
    ]);
```
The `exclude` is not being passed onto `rsync`

> https://deployer.org/docs/api.html#download

bilogic:
The docs are not clear on this aspect.

`download()` and `upload()` relies on `src/Utility/Rsync.php` and `options` require the `--` prefix

e.g. `$config = ['options' => ['--copy-links']];`

This should not be confused with `contrib/rsync.php` which requires that the `--` prefix be omitted.

bilogic:
I will find some time to update the docs... let me sort out my deployment code first :)

PR #2601, opened 2021-07-06
docs making clear there are 2 different rsync

bilogic:
as discussed on #2568
Is there a way to link from `api.md` to `rsync.md`?
I would like to link `upload()` and `download()`.

### DIFF
--- a/docs/rsync.md
+++ b/docs/rsync.md
@@ -1,0 +1,23 @@
+# rsync
+
+[Source](/src/Utility/Rsync.php)
+
+## IMPORTANT
+
+This is the deployer's built-in rsync and must not be confused with `/contrib/rsync.php`. Their configuration options are also very different, read carefully below.
+
+## Configuration options
+
+- **rsync**: Accepts an array with following options (all are optional and defaults are ok):
+    - *timeout*: accepts an *int* defining timeout for rsync command to run locally.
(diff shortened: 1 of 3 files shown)


docs/rsync.md
```diff
@@ -0,0 +1,23 @@
+# rsync
+
+[Source](/src/Utility/Rsync.php)
+
+## IMPORTANT
+
+This is the deployer's built-in rsync and must not be confused with `/contrib/rsync.php`. Their configuration options are also very different, read carefully below.
+
+## Configuration options
```
antonmedv, line 9: Better to put it next to upload/download func.

bilogic: ok, what is the syntax to link from `api.md` to `rsync.md`?

bilogic: @antonmedv any idea? thanks

Schrank: [TEXT](LINK TARGET)
e.g.[Source](/src/Utility/Rsync.php)

Does this help?

antonmedv: Also I don’t like wording. It’s not an external docs. “Their” not fit here.

bilogic: @Schrank I'm not sure, I could not find any references around, but let me try
@antonmedv Let me find some time to fix it all up

PR #2603, opened 2021-07-14
Fixed process args not being escaped in Master

### LINKED ISSUE #2450: Cannot deploy from working directory with space in path name
- Deployer version: v7.0.0-beta.13
- Deployment OS: Linux 4.4.0-18362-Microsoft (WSL)

When the current working directory contains a space in the path name (e.g. `/mnt/c/Foo Bar`), deploy fails during connection with:

>In Master.php line 165:

>  Could not open input file: /mnt/c/Foo

This was not an issue with version 6.8.x.

Moreover, if PHP binary itself is on a path with a space in the filename, it will fail in a similar fashion but with slightly different output. For example, when PHP is installed at `c:\prog\PHP 8.0`, deploy fails with:

>'C:\prog\PHP' is not recognized as an internal or external command,
operable program or batch file.
>In Master.php line 165:

>  [Deployer\Exception\ConnectException]

Again, this is a v7 exclusive regression.

Bilge:
* Restored `write()` function.

- [x] Bug fix #2450 
- [ ] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [ ] Changelog updated?

## Todo

@antonmedv I left a few things out, mostly relating to things `Stringify::options` did. Hope you don't mind picking those up:

* Forwarding verbosity (not sure why this is needed since there's already a `-v` switch added in the calling method?)
* Forwarding `--decorated` option.
* Adding test for `Arrayify::options`.
* Removing `Stringify::options` and accompanying test (no longer used).

### DIFF
--- a/src/Support/Arrayify.php
+++ b/src/Support/Arrayify.php
@@ -1,0 +1,23 @@
+<?php
+declare(strict_types=1);
+
+namespace Deployer\Support;
+
+use Symfony\Component\Console\Input\InputInterface;
+
+final class Arrayify
+{
+    public static function options(InputInterface $input): array
+    {
+        $options = array_filter($input->getOptions());
+
+        return iterator_to_array(
+            \iter\flatten(
+                \iter\zip(
+                    \iter\map(fn ($name) => "--$name", \iter\keys($options)),
+                    \iter\values($options)
+                )
+            )
+        );
+    }
+}
--- a/src/Support/Arrayify.php
+++ b/src/Support/Arrayify.php
@@ -1,6 +1,4 @@
-<?php
-declare(strict_types=1);
-
+<?php declare(strict_types=1);
 namespace Deployer\Support;
 
 use Symfony\Component\Console\Input\InputInterface;
--- a/src/Support/Arrayify.php
+++ b/src/Support/Arrayify.php
@@ -2,6 +2,10 @@
 namespace Deployer\Support;
 
 use Symfony\Component\Console\Input\InputInterface;
+use function iter\flatten;
+use function iter\map;
+use function iter\mapWithKeys;
+use function iter\toArray;
 
 final class Arrayify
 {
@@ -9,11 +13,13 @@ final class Arrayify
     {
         $options = array_filter($input->getOptions());
 
-        return iterator_to_array(
-            \iter\flatten(
-                \iter\zip(
-                    \iter\map(fn ($name) => "--$name", \iter\keys($options)),
-                    \iter\values($options)
+        return toArray(
+            flatten(
+                mapWithKeys(
+                    fn ($v, $k) => is_array($v)
+                        ? map(fn ($v) => ["--$k", $v], $v)
+                        : ["--$k", $v],
+                    $options
                 )
             )
         );
--- a/src/functions.php
+++ b/src/functions.php
@@ -600,7 +600,7 @@ function warning(string $message): void
 /**
  * Writes a message to the output.
  */
-function write(string|iterable $message, int $options = 0): void
+function write(string $message, int $options = 0): void
 {
     output()->write(parse($message), false, $options);
 }
@@ -608,7 +608,7 @@ function write(string|iterable $message, int $options = 0): void
 /**
  * Writes a message to the output and adds a newline at the end.
  */
-function writeln(string|iterable $message, int $options = 0): void
+function writeln(string $message, int $options = 0): void
 {
     $host = currentHost();
     output()->writeln("[$host] " . parse($message), $options);
--- a/src/functions.php
+++ b/src/functions.php
@@ -598,10 +598,17 @@ function warning(string $message): void
 }
 
 /**
- * Writes a message to the output and adds a newline at the end.
- * @param string|array $message
+ * Writes a message to the output.
  */
-function writeln($message, int $options = 0): void
+function write(string|iterable $message, int $options = 0): void
+{
+    output()->write(parse($message), false, $options);
+}
+
+/**
+ * Writes a message to the output and adds a newline at the end.
+ */
+function writeln(string|iterable $message, int $options = 0): void
 {
     $host = currentHost();
     output()->writeln("[$host] " . parse($message), $options);
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -2,9 +2,8 @@ name: check
 
 on:
   push:
-    branches: [ master ]
   pull_request:
-    branches: [ master ]
+  workflow_dispatch:
 
 jobs:
   phpstan:
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -2,7 +2,9 @@ name: check
 
 on:
   push:
+    branches: [ master ]
   pull_request:
+    branches: [ master ]
   workflow_dispatch:
 
 jobs:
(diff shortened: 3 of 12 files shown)


src/functions.php
```diff
@@ -597,11 +597,18 @@ function warning(string $message): void
     }
 }
 
+/**
+ * Writes a message to the output.
+ */
+function write(string|iterable $message, int $options = 0): void
```
antonmedv, line 603: It was removed as it only redirects call to output() and don’t add any helpers beside parse.

src/functions.php
```diff
@@ -597,11 +597,18 @@ function warning(string $message): void
...
+{
+    output()->write(parse($message), false, $options);
+}
+
 /**
  * Writes a message to the output and adds a newline at the end.
- * @param string|array $message
  */
-function writeln($message, int $options = 0): void
+function writeln(string|iterable $message, int $options = 0): void
```
antonmedv, line 611: Iterable is wrong. Parse don’t take irerable.

src/Support/Arrayify.php
```diff
@@ -0,0 +1,23 @@
...
+use Symfony\Component\Console\Input\InputInterface;
+
+final class Arrayify
+{
+    public static function options(InputInterface $input): array
+    {
+        $options = array_filter($input->getOptions());
+
+        return iterator_to_array(
+            \iter\flatten(
```
antonmedv, line 15: Is is possible to implement without extra dependencies?

antonmedv (changes requested):

src/Support/Arrayify.php
Bilge: Yes, of course, and it would be more verbose, inelegant and harder to maintain.

src/functions.php
Bilge: [Iterable is correct](https://github.com/symfony/symfony/blob/b5d4b33d5e65d956194c170fda5f8f40bc6a921a/src/Symfony/Component/Console/Output/OutputInterface.php#L36).

src/functions.php
Bilge: `write()` is more convenient and readable than `output()->write()`. It does not make sense to me to break backwards compatibility here.

antonmedv: writeln also adds host info now, but write not.

src/Support/Arrayify.php
antonmedv: Let’s do it)

src/functions.php
Bilge: Yes, but it can't, because the _whole point_ is that it doesn't _necessarily_ start a new line. I agree we might want a different solution here. Maybe we need two functions, like `writeStart()` and `writeContinue()`, where only _start_ adds the host info and _continue_ does not, but neither adds a trailing newline. Open to ideas.

src/Support/Arrayify.php
Bilge: If you want to change this, I'll leave it to you.

src/functions.php
```diff
@@ -597,11 +597,18 @@ function warning(string $message): void
...
+ * Writes a message to the output.
+ */
+function write(string $message, int $options = 0): void
+{
+    output()->write(parse($message), false, $options);
+}
+
 /**
  * Writes a message to the output and adds a newline at the end.
- * @param string|array $message
```
Bilge, line 602: I don't know that `parse()` "doesn't support" `iterable`. If you pass `iterable`, `parse()` just ignores it because `parse()` only deals with strings, but other types pass through unchanged. Moreover, this was previously typed as `array`. I don't know if it's OK for iterables (including arrays) to pass through `parse()` unchanged because I don't understand what `parse()` does, but if it just fiddles around with newlines, perhaps this doesn't apply to iterables anyway, since the underlying Symfony library will handle this.

Bilge:
@antonmedv Can you build on this or do you require me to develop the entire feature to your satisfaction?

antonmedv:
I fix it.

Bilge:
@antonmedv Any ETA on this?

antonmedv:
Currently no eta. Will try to take a look next week.

Bilge:
@antonmedv I understand you are busy. Would you perhaps reconsider adding @nikic's [iter](https://github.com/nikic/iter/) library anyway, considering it's just a couple of files with a few dozen [useful functions](https://github.com/nikic/iter/blob/master/src/iter.php)? Then we could just get this merged.

antonmedv:
Ok, let's just make tests pass.

antonmedv (approved):

Bilge:
@antonmedv I fixed most of the tests, although it is quite hard to check since I am not allowed to run workflows. There is just [one unit and one e2e failing now](https://github.com/Bilge/deployer/actions/runs/1121870128) but I don't know how to fix them. Can you help?

Bilge:
@antonmedv Any insight into these failures, or do I need to figure these out on my own?

antonmedv:
Well, I can take a look (then I find a free time). Also, you can try to fix it as in master everything works, so it's something with the tests.

Bilge:
I understand it's my fault. It's just that these tests in particular are very hard to follow and to figure out how they are ever expected to pass, even though I understand they do on master.

Bilge:
@antonmedv I've spent a long time trying to debug why `ParallelTest::testOption` fails to emit `Hello, prod!`. I assumed it would be something to do with the way I'm passing options to the process, since that's the only significant thing I've changed, but it's not that.

The `bash -s` process *is* created with input `"echo Hello, prod!"` and it runs fine and the expected output is generated. The problem is that the output goes nowhere. `ProcessRunner` returns the output to the `run()` function.

https://github.com/deployphp/deployer/blob/363f7ed6a15c4a7a967f38a29bf357a82784196f/src/Component/ProcessRunner/ProcessRunner.php#L66

The `run()` function returns the output to the `parallel.php` task file...

https://github.com/deployphp/deployer/blob/363f7ed6a15c4a7a967f38a29bf357a82784196f/src/functions.php#L381

...but `parallel.php` does nothing with that output.

https://github.com/deployphp/deployer/blob/363f7ed6a15c4a7a967f38a29bf357a82784196f/tests/joy/recipe/parallel.php#L24

So it's unclear to me how the output is ever expected to appear in the console's output stream in the first place. It seems assumed that the output pipe of the sub-process is somehow connected to the parent process's output stream (the console) but it's not. And I don't know how it ever was the case that it was. I checked the source code to see if pipes are somehow handled differently when using `new Process()` instead of `Process::fromShellCommandline()`, but they're not. So I just don't know how it ever worked and therefore what the correct way to fix it would be. Do you know?

antonmedv:
Thanks for efforts. Will rake a look myself on it. 
in verbose mode output of subprocess is printed to stdout.

Bilge:
Ah yes, most likely it is something to do with that: I am probably not propagating the verbose flags properly.

Bilge:
@antonmedv I fixed all the tests except the Laravel E2E. That seems to be failing just because PHP is now at 8.0.

Bilge:
@antonmedv I think I fixed everything 😁

Bilge:
`write()` is kind of useless now because if you `write('foo'); run(); write('bar')`, it will appear as
>foo
  bar

I cannot figure out why this is. I thought maybe the spinner was emitting newlines, but I commented it out and it still happens.

antonmedv:
Fixed in f3152a5264d0fd96316ff4ad5a01cb3024710907. Thanks for your work, but I prefer to merge smaller PRs and don't add separate dependencies.

PR #2604, opened 2021-07-15
Cachetool improvements

Tobion:
- remove outdated cachetool apc command (not apcu) that does not exist since 4.x,
- document web adapter example
- use release_or_current_path to be able to execute tasks outside of deployment (#2486)


antonmedv:
Please, regenerate docs.

Tobion:
Done

PR #2610, opened 2021-07-23
Fixed sw:plugin:upgrade:all task in Shopware recipe

nexxome:
- [x] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [x] Changelog updated?

      $ php bin/changelog


peterjaap:
@nexxome in what scenario would this task run on a current release?

nexxome:
It seems that this task isn't used anywhere during deployment at the moment. So I thought it could be useful to run this task manually.

peterjaap:
@nexxome ah yes, you are right. That was because we didn't want to automatically include it in the deploy. This seems a reasonable alternative :)

PR #2626, opened 2021-08-18
Use `cleanup_use_sudo` in task `deploy:release`

PATROMO:
This task failed every time because wie need `sudo` to clean. The code ist 1:1 copy from task `Cleaning up old releases` https://github.com/deployphp/deployer/blob/6.x/recipe/deploy/cleanup.php#L31.

- [x] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [ ] Changelog updated?

### DIFF
--- a/recipe/deploy/release.php
+++ b/recipe/deploy/release.php
@@ -96,14 +96,15 @@ set('release_path', function () {
 
 desc('Prepare release. Clean up unfinished releases and prepare next release');
 task('deploy:release', function () {
+    $sudo = get('cleanup_use_sudo') ? 'sudo' : '';
     cd('{{deploy_path}}');
 
     // Clean up if there is unfinished release
     $previousReleaseExist = test('[ -h release ]');
 
     if ($previousReleaseExist) {
-        run('rm -rf "$(readlink release)"'); // Delete release
-        run('rm release'); // Delete symlink
+        run("$sudo rm -rf \"$(readlink release)\""); // Delete release
+        run("$sudo rm release"); // Delete symlink
     }
 
     // We need to get releases_list at same point as release_name,


Schrank:
@PATROMO Thanks for your contribution, I like! <3

@antonmedv any idea why no checks ran?

Schrank:
Ah version 6 🤦  @antonmedv any reason to make a difference to version 7? If not I would merge this :-)

antonmedv:
I already don’t remember why it was deleted 😜🙃

let’s bring it back.

PATROMO:
@antonmedv can you merge this and create a new release? :)

PATROMO:
hmm... @antonmedv why revert?

antonmedv:
Reverted. Was too quick to merge. 

@Schrank we did a poor review. How can we improve?

@PATROMO a few issues:
* cleanup_use_sudo is designed especially for deploy:cleanup task, not deploy:release.
* currently changes only accepting in the master branch, not 6.x. You still can use alder versions on 6.x, but maintenance now only focused on 7.x.

PATROMO:
If you need `sudo` to delete old releases, then you will also need `sudo` to delete an unfinished release. So I thought it makes sense to use the same constant for this. After all, that's what it does: `cleanup_use_sudo`!  Defining an extra new constant for this I thought was overkill.
Too bad that a bug is not fixed in a young version.

antonmedv:
What about fixing it in 7.x as it is still in beta? For this, we can find a new config name that will be best for both tasks?

Schrank:
I second that, although „cleanup“ broken release is imho ok. I like the PR and the config name (my 2 cents)

antonmedv:
Hmmm, but it was for 6.x still) As you can see there was no releases of 6.x in a long time.

antonmedv:
Let’s find something what will work for 7.x?

tehplague:
Could you re-consider this patch for 7.x? Using `cleanup_use_sudo` for this purpose as well IMHO totally makes sense, for reasons @PATROMO already pointed out.

antonmedv:
Let's create a separate issue to discuss the problem.

PATROMO:
@tehplague my own **_dirty_** solution: 

```php
task('fix_permissions', function () {
    run('sudo chown -R deploy:deploy {{release_path}}');
});

after('deploy:failed', 'fix_permissions');
```

PR #2644, opened 2021-09-01
Allow overriding deployer root directory globally

### LINKED ISSUE #2628: Allow overriding DEPLOYER_ROOT or default process runner cwd
For issue [1774](https://github.com/deployphp/deployer/issues/1774) a [DEPLOYER_ROOT constant](https://github.com/deployphp/deployer/blob/88f6e423688bfe812c125a023bfbc024cd097118/bin/dep#L104) was introduced, which is used in the [ProcessRunner](https://github.com/deployphp/deployer/commit/29a643b62389542394c7b254f6c913f7191132ba#diff-b50c325c16c55b4ce886df4de55eb0f3aff1c44677a1a80f35c4fa96e8df3cfbR44) to set the default cwd. When commands are executed from the root of a project this would mean all run commands have to be configured to override the cwd.

Unfortunately there is no way to override the constant or cwd option globally. Therefore I'd like to suggest making the constant, default cwd option or both overridable, either via a setting or an environment variable. Something like:

bin/dep
```php
define('DEPLOYER_ROOT', getenv('DEPLOYER_ROOT') !== false ? getenv('DEPLOYER_ROOT') : dirname($deployFile));
```

ProcessRunner::run()
```php
'cwd' => has('deployer_root') ? get('deployer_root') : (defined('DEPLOYER_ROOT') ? DEPLOYER_ROOT : null),
```

kjkooistra-youwe:
Use an environment variable or setting to override the default deployer root
directory used for the runLocally() default cwd configuration.
Resolves #2628

- [ ] Bug fix #…?
- [x] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [ ] Changelog updated?

### DIFF
--- a/bin/dep
+++ b/bin/dep
@@ -101,5 +101,5 @@ exit(1);
 run:
 set_include_path($includePath . PATH_SEPARATOR . get_include_path());
 define('DEPLOYER_DEPLOY_FILE', $deployFile);
-define('DEPLOYER_ROOT', dirname($deployFile));
+define('DEPLOYER_ROOT', getenv('DEPLOYER_ROOT') !== false ? getenv('DEPLOYER_ROOT') : dirname($deployFile));
 \Deployer\Deployer::run(DEPLOYER_VERSION, $deployFile);
--- a/docs/UPGRADE.md
+++ b/docs/UPGRADE.md
@@ -54,6 +54,10 @@
    ``` 
 8. Rename success task in ~~success~~ to `deploy:success`.
 9. Verbosity function (`isDebug()`, etc) deleted. Use `output()->isDebug()` instead.
+10. runLocally() commands are executed relative to the recipe file directory. This behaviour can be overridden via an environment variable:
+    ```bash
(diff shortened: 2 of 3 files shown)


src/Component/ProcessRunner/ProcessRunner.php
```diff
@@ -36,7 +39,9 @@ public function run(Host $host, string $command, array $config = []): string
         $defaults = [
             'timeout' => $host->get('default_timeout', 300),
             'idle_timeout' => null,
-            'cwd' => defined('DEPLOYER_ROOT') ? DEPLOYER_ROOT : null,
+            'cwd' => has('deployer_root') ? get('deployer_root') : (
```
antonmedv, line 42: Lets keep only env. Right now root can be modified two ways but will affected in different places.

antonmedv (changes requested):

antonmedv (approved):

Issue #2656, opened 2021-09-17
Replace rsync with cp in copy_dirs.php


gjuric:
Is there a reason while rsync is used in the default `deploy:copy_dirs` task?

It seems to me that it would be simpler to replace:

```php
    run("mkdir -p {{release_path}}/$dir");
    run("rsync -av {{previous_release}}/$dir/ {{release_path}}/$dir");
````

with

```php
    run("cp -r {{previous_release}}/$dir {{release_path}}/");
```

and it wouldn't require installation of rsync on the host.

antonmedv:
Actually there is no reason. Found this: https://github.com/deployphp/deployer/pull/1072

If you want you can work on PR for this (include checks for tailing slash, -rtf, and other disscussed stuff from #1072) and it will be included to v7 release!

It will be cool.

gjuric:
I've created a PR and tested if manually, copying nested directories worked.

PR #2659, opened 2021-09-17
Use cp instead of rsync to copy directories

gjuric:
Since rsync is not installed on every host it's better to use standard tools.
Previosly rsync was introduced ( #1072 ) to enable copying of subdirectories, but
this is doable with cp as well.

- [ ] Bug fix #…?
- [x] New feature #2656
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [x] Changelog updated?


recipe/deploy/copy_dirs.php
```diff
@@ -7,7 +7,7 @@
         foreach (get('copy_dirs') as $dir) {
             if (test("[ -d {{previous_release}}/$dir ]")) {
                 run("mkdir -p {{release_path}}/$dir");
-                run("rsync -av {{previous_release}}/$dir/ {{release_path}}/$dir");
```
antonmedv, line 10: Should we check what `$dir` does not contain `/` at the end? 

Will be nice to throw an exception with explanation of why to remove `/` from copy_dirs.

gjuric: I can also just trim `/` from the provided strings, in that case we don't need to throw an exception, what do you think?

antonmedv: I think better to throw exception with explanation. In future it will help.

gjuric: Done.

gjuric:
I had to push a fix that strips the last name of the folder in the destination path for things to work properly.

antonmedv:
Please check errors.

gjuric:
It looks like an error that is not in the code I wrote, but it was fixed in https://github.com/deployphp/deployer/pull/2668

recipe/deploy/copy_dirs.php
```diff
@@ -5,9 +5,16 @@
...
+                throw new \RuntimeException('Entries in config parameter "copy_dirs" must not end with "/"');
+            }
             if (test("[ -d {{previous_release}}/$dir ]")) {
+                $destinationDir = '';
+                if (strpos($dir, '/') !== false) {
+                    $destinationDir = substr($dir, 0, strrpos($dir, '/') + 1);
+                }
                 run("mkdir -p {{release_path}}/$dir");
-                run("rsync -av {{previous_release}}/$dir/ {{release_path}}/$dir");
+                run("cp -rpf {{previous_release}}/$dir {{release_path}}/$destinationDir");
```
Bilge, line 17: This should probably be `cp -af`.

PR #2664, opened 2021-09-22
Fixed git submodule command requires checked out worktree.

sun:
Problem
- The [common solution for handling git submodules](https://stackoverflow.com/a/31627058/811306) does not work with Deployer, because Git isn't able to reference the main repo from the archive.

Proposed solution
1. Check out an actual work tree from the bare git repository, so that subsequent deployment commands requiring the repository data can operate.

Example
```yaml
tasks:
  build_app:
    script:
    - cd {{release_path}}
    - git --git-dir={{deploy_path}}/.dep/repo --work-tree=. submodule update --init --recursive
    - npm run install

after:
  deploy:update_code: build_app
```

- [x] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [ ] Tests added?
- [ ] Changelog updated?

      $ php bin/changelog

### DIFF
--- a/recipe/deploy/update_code.php
+++ b/recipe/deploy/update_code.php
@@ -73,7 +73,7 @@ task('deploy:update_code', function () {
 
     // Copy to release_path.
     run("$git remote update 2>&1");
-    run("$git archive $at | tar -x -f - -C {{release_path}} 2>&1");
+    run("$git --work-tree={{release_path}} checkout $at");
 
     // Save revision in releases log.
     $rev = run("$git rev-list $at -1");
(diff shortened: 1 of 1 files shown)


antonmedv:
Merged! Now available with update_code_strategy.

sun:
Thanks for the prompt response and commit https://github.com/deployphp/deployer/commit/d2c4521b7a3736ca7d1d3360b0a278dbe73cd30b - turning it into an option makes a lot of sense 👌

PR #2672, opened 2021-09-24
Fix SshCommand not retrieving the host by its alias

### LINKED ISSUE #2528: [7.0-beta] Aliases don't work as expected
- Deployer version: 7.0.beta-23

Hello there,
It looks like we can't ssh to a specific host with its label anymore.

Example: `deploy.php`:

```php
namespace Deployer;

$foo = host('127.0.0.1');
$foo->set('alias', 'example.com');

$bar = host('127.0.0.2');
$bar->set('alias', 'example.org');
```

Capture d’écran 2021-04-27 à 17 06 22

Running `php vendor/bin/dep ssh 127.0.0.1` works, but:
- It requires to know the server's IP
- Sometimes you just don't know it (I dynamically fetch a list of servers matching a specific tag, from DigitalOcean API, for instance)
- It used to work in 6.8.

Thank you,
Ben

TheFakeCake:
This PR enforces the use of host aliases by the ssh command.

I'm not sure if this is the intended behaviour though, so let me know if modifications should be made.

I also don't know if this is a breaking change since I never used v6.8 (I'm new to Deployer)

- [x] Bug fix #2528
- [ ] New feature?
- [ ] BC breaks?
- [ ] Deprecations?
- [x] Tests added?
- [ ] Docs updated?
- [ ] Changelog updated?

### DIFF
--- a/src/Command/SshCommand.php
+++ b/src/Command/SshCommand.php
@@ -47,7 +47,7 @@ class SshCommand extends Command
         $this->telemetry();
         $hostname = $input->getArgument('hostname');
         if (!empty($hostname)) {
-            $host = $this->deployer->hosts->get($hostname);
+            $host = $this->deployer->hosts->findOneByAlias($hostname);
         } else {
             $hostsAliases = [];
             foreach ($this->deployer->hosts as $host) {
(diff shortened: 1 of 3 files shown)


TheFakeCake:
I corrected the code to pass phpcs and phpstan checks.

antonmedv:
https://github.com/deployphp/deployer/issues/2528#issuecomment-927259310

PR #2715, opened 2021-10-24
feat: add support for hidden keys in yaml files

kevinreniers:
As discussed in https://github.com/deployphp/deployer/discussions/2709, I'd like to add back support for root level hidden keys in YAML files. 

I've chosen to remove the hidden keys completely after parsing because their references are already resolved during loading, so there is no point in keeping them after that. That's also the reason why I didn't change the JSON schema file that validates the contents of the input data.

- [x] New feature?
- [x] Tests added?


kevinreniers:
Ok, done.

PR #2726, opened 2021-11-03
Increase min-version of react/http

luca-rath:
- [x] Bug fix?
- [ ] New feature?
- [x] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

`Deployer\Executor\Server` uses `React\Http\HttpServer`, but that class is only available in `react/http:^1.5`, therefore I changed the min-requirement

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,7 @@
         "ext-json": "*",
         "justinrainbow/json-schema": "^5.2",
         "psr/http-message": "^1",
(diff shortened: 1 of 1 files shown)


antonmedv:
Looks like .lock file is out of sync. 
Updated myself. Thanks!

luca-rath:
Sorry, did edit the `composer.json` directly here in github ... Thanks for fixing!

PR #2734, opened 2021-11-07
Make Magento 2 recipe suitable for push strategy deployments

peterjaap:
- [ ] Bug fix #…?
- [x] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

When refactoring our Deployer 6 recipes for Magento 2 to Deployer 7, I was eventually left with this code block in all our deployment configurations;

```php
// Override default recipe tasks since we use a push-approach instead of a pull-approach
task('deploy', [
    'deploy:prepare',
//    'deploy:vendors', // We run composer install in the CI pipeline
    'deploy:clear_paths',
    'deploy:magento',
    'deploy:publish',
]);
task('deploy:magento', [
    //'magento:compile', // We run bin/magento setup:di:compile in the CI pipeline
    //'magento:deploy:assets', // We run bin/magento setup:static-content:deploy in the CI pipeline
    'magento:config:import',
    'magento:upgrade:db',
    'magento:cache:flush'
]);
task('deploy:prepare', [
    'deploy:info',
    'deploy:setup',
    'deploy:lock',
    'deploy:release',
    //`deploy:update_code`, // We do not push or pull the Git repo
    'deploy:shared',
    'deploy:writable',
]);
```

This is because we don't compile DI and generate static assets on the production server but instead in our CI/CD pipeline. I know for a fact a large number of other Magento 2 developers also do this, so they will also need to override in some way what we did above.

To facilitate the push (instead of pull) deployment strategy, I have introduced a setting `push_strategy` which by default is set to `false`. This will make sure the recipe is backwards compatible.

However, if the `push_strategy` is set to true (or, non-false), the following tasks will not be run;

- `deploy:vendors`
- `magento:compile`
- `magento:deploy:assets`
- `deploy:update_code`

Technically, the `deploy:update_code` will run, but the task is set to do nothing. This is because this task lives in the common recipe, so we can't add a Magento-specific check there. This will only be a problem in edge cases where somebody configured an after hook on `deploy:update_code` which will assume the code is actually updated.

Merging this PR will make our `deploy.php` file a lot cleaner and easier to grasp. It will also make upgrading to future versions easier/less error-prone since there is less code overridden.

Edit:

Using `disabled()` makes this a lot easier. This PR is now updated to move the build steps into its own group, so we can disable the pull approach by adding 3 lines;

```php
task('deploy:vendors')->disabled();
task('deploy:update_code')->disabled();
task('magento:build')->disabled();
```

### DIFF
--- a/recipe/magento2.php
+++ b/recipe/magento2.php
@@ -19,6 +19,12 @@ add('recipes', ['magento2']);
 // in you deployer script.
 set('static_content_locales', 'en_US');
 
+// By default, the deployment strategy for this recipe is
+// pull instead of push, to make it backwards compatible with Deployer 6.
+// If you change this to push, you will have to push the generated & compiled
+// files yourself, for example by adding a rsync hook after deploy:release
+set('push_strategy', false);
+
 set('content_version', function () {
     return time();
 });
@@ -171,8 +177,6 @@ task('magento:cache:flush', function () {
 
 desc('Magento2 deployment operations');
 task('deploy:magento', [
-    'magento:compile',
-    'magento:deploy:assets',
     'magento:config:import',
     'magento:upgrade:db',
     'magento:cache:flush'
@@ -181,10 +185,19 @@ task('deploy:magento', [
 desc('Deploy your project');
 task('deploy', [
     'deploy:prepare',
-    'deploy:vendors',
     'deploy:clear_paths',
     'deploy:magento',
     'deploy:publish',
 ]);
+
+if (get('push_strategy') === false) {
+    after('deploy_prepare', 'deploy:vendors');
+    before('magento:config:import', 'magento:compile');
+    before('magento:config:import', 'magento:deploy:assets');
+} else {
+    task('deploy:update_code', function () {
+        // Do not update code when using push strategy
+    });
+}
 
 after('deploy:failed', 'magento:maintenance:disable');
--- a/recipe/magento2.php
+++ b/recipe/magento2.php
@@ -19,12 +19,6 @@ add('recipes', ['magento2']);
 // in you deployer script.
 set('static_content_locales', 'en_US');
 
-// By default, the deployment strategy for this recipe is
-// pull instead of push, to make it backwards compatible with Deployer 6.
-// If you change this to push, you will have to push the generated & compiled
-// files yourself, for example by adding a rsync hook after deploy:release
-set('push_strategy', false);
-
 set('content_version', function () {
     return time();
 });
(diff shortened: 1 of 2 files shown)


recipe/magento2.php
```diff
@@ -181,10 +185,19 @@
 desc('Deploy your project');
 task('deploy', [
     'deploy:prepare',
-    'deploy:vendors',
     'deploy:clear_paths',
     'deploy:magento',
     'deploy:publish',
 ]);
 
+if (get('push_strategy') === false) {
```
antonmedv, line 193: But this code executed right away. How to override it in users recipe?

antonmedv:
I think this is common to all recipes. Every recipe is designed with the assumption that deploy and build hosts are one.

And to introduce CI and push strategy every recipe requires some tweaking. I also was thinking about it a lot. So I decided to take this approach: 
 - Design recipes for one build and deploy host.
 - Make it easy to introduce CI server or build a server.
 - Write good docs on how to do it.

recipe/magento2.php
```diff
@@ -181,10 +185,19 @@
...
+    after('deploy_prepare', 'deploy:vendors');
+    before('magento:config:import', 'magento:compile');
+    before('magento:config:import', 'magento:deploy:assets');
+} else {
+    task('deploy:update_code', function () {
```
antonmedv, line 198: Now we can use:
```php
task('deploy:update_code')->disabled();
```

recipe/magento2.php
```diff
@@ -171,8 +177,6 @@
 
 desc('Magento2 deployment operations');
 task('deploy:magento', [
-    'magento:compile',
-    'magento:deploy:assets',
```
antonmedv, line 175: What about separating it into "build" task what is enabled by default and add doc on how to disable this step?

recipe/magento2.php
peterjaap: Cool!

recipe/magento2.php
peterjaap: Good point. Using the `disabled()` method this isn't needed because we can do this in our custom recipe;

```
task('deploy:vendors')->disabled();
task('deploy:update_code')->disabled();
task('magento:compile')->disabled();
task('magento:deploy:assets')->disabled();
```

recipe/magento2.php
peterjaap: Ok I did just that.

Now we can disable 'pull' strategy by adding only 3 lines;

```
task('deploy:vendors')->disabled();
task('deploy:update_code')->disabled();
task('build')->disabled();
```

antonmedv:
Awesome!

carlos-reynosa:
I've been confused and trying to work with deployer for the first time and i can only imagine that most people who try to use it are not looking to build and deploy code on the same server. So that is to say i think the conclusion of this ticket is very important. 

So what are the step to setup a simple deploy with a seperate build step? 

## Disable build tasks
task('deploy:vendors')->disabled();
task('deploy:update_code')->disabled();
task('build')->disabled();

1. Then what? What would be the process for initiating the static content deploy locally and di compile locally? Do we have to have a localhost host setup and then call the tasks on it? 
2. How should the deploy script be set up to upload the current build to the current release?

PR #2752, opened 2021-11-14
Fix timeout for heavy tasks in provision recipe

ramundomario:
Testing deployer & the new provision recipe I noticed timeout for heavy tasks such as "Upgrades all packages" and "Installs packages".
The timeout for this tasks now is 15 minutes, WDYT?

Screenshot from 2021-11-14 18-31-45
Screenshot from 2021-11-14 18-40-08

- [X] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

      $ php bin/docgen

### DIFF
--- a/recipe/provision.php
+++ b/recipe/provision.php
@@ -100,7 +100,7 @@ task('provision:update', function () {
 
 desc('Upgrades all packages');
 task('provision:upgrade', function () {
-    run('apt-get upgrade -y', ['env' => ['DEBIAN_FRONTEND' => 'noninteractive']]);
+    run('apt-get upgrade -y', ['env' => ['DEBIAN_FRONTEND' => 'noninteractive'], 'timeout' => 900]);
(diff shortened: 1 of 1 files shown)


recipe/provision.php
```diff
@@ -100,7 +100,7 @@
 
 desc('Upgrades all packages');
 task('provision:upgrade', function () {
-    run('apt-get upgrade -y', ['env' => ['DEBIAN_FRONTEND' => 'noninteractive']]);
+    run('apt-get upgrade -y', ['env' => ['DEBIAN_FRONTEND' => 'noninteractive']], 900);
```
antonmedv, line 103: Let’s better to put it in second arg as timeout => 900

ramundomario: Done! :wink:

Issue #2754, opened 2021-11-15
Temporary workaround in shopware recipe uses constant path


phizab:
- Deployer version: 7.0.0-rc2
- Deployment OS: Ubuntu 20.04

The following line in the shopware recipe fixes absolute paths in theme config files. But the given absolute path can vary from deployment to deployment:

`runLocally('sed -i "" -E \'s/\\\\\/var\\\\\/www\\\\\/htdocs\\\\\/releases\\\\\/[0-9]+\\\\\///g\' files/theme-config/*');`

In my case the project is stored in an other path than "/var/www/htdocs". There is currently no way to change this path.

Btw: if I'm right the workaround is no longer needed, since it was fixed in Shopware 6.4.6.0: https://issues.shopware.com/issues/NEXT-17720. Maybe the workaround can simply be deleted.

antonmedv:
I didn’t get it) Can you make a pr to discuss it?

phizab:
Created PR #2755 for removal of the workaround.

peterjaap:
The PR mentioned causes deployments on 6.4.6.0 not to fail, but still won't work for <6.4.6.0 where the release path is not exactly the same as `/var/www/htdocs/releases`. So that needs to be fixed.

peterjaap:
Example file;

```json
[
    {
        "storefront": {
            "path": "Resources\/app\/storefront\/src",
            "entryFilePath": "Resources\/app\/storefront\/src\/main.js",
            "webpack": null,
            "styleFiles": [
                "\/var\/www\/htdocs\/releases\/20211124115627\/vendor\/shopware\/storefront\/Resources\/app\/storefront\/src\/scss\/base.scss",
                "\/var\/www\/htdocs\/releases\/20211124115627\/vendor\/shopware\/storefront\/Resources\/app\/storefront\/src\/scss\/skin\/shopware\/_base.scss",
                "@Plugins"
            ]
        }
    }
]
```

Run `sed -E 's#\\/var\\/www\\/htdocs\\/releases\\/[0-9]+\\/##g' test.json`;

```
[
    {
        "storefront": {
            "path": "Resources\/app\/storefront\/src",
            "entryFilePath": "Resources\/app\/storefront\/src\/main.js",
            "webpack": null,
            "styleFiles": [
                "vendor\/shopware\/storefront\/Resources\/app\/storefront\/src\/scss\/base.scss",
[...]

Issue #2710, opened 2021-10-21
Implement sorter for Shopware


antonmedv:
Hi, @Schrank 

I had trouble with dependency for sorting. Ended up deleting it. Please reimplement sorter here:

https://github.com/deployphp/deployer/blob/33bdaacfea8d8ecea41ec299d3556bfcdac10861/recipe/shopware.php#L121-L122

Thanks.

Schrank:
> But why topological sort is needed?

Because the shopware plugins are composer packages and can have different dependencies, therefore we need to take care of the right order to activate and install them :-/

antonmedv:
@Schrank lets create two recipes? :)

antonmedv:
I’ll refactor current recipe to simple recipe. You more complex is needed, consider adding it too? It it okay with everyone?

antonmedv:
Updated in b40c641ee21ef1e4cc5921a5dc704d21d8ffd7b5. Please verify in beta.39.

UlrichThomasGabor:
@antonmedv Is the documentation generated by hand or automatically? I would change `This task is group task which contains next tasks:` to `This task is a group task which contains the following tasks:`

I'll try the recipe out but on the first glance it looks good!

antonmedv:
Yes, it’s generated. See src/Documentation.

PR #2755, opened 2021-11-15
Fix temporary workaround in Shopware recipe

phizab:
If I'm right the workaround is no longer needed, since it was fixed in Shopware 6.4.6.0: https://issues.shopware.com/issues/NEXT-17720. Can anyone confirm this please?

Can anyone confirm this please?


phizab:
@UlrichThomasGabor You mentioned this workaround in https://github.com/deployphp/deployer/issues/2710#issuecomment-954821426. Can you confirm that it got fixed in 6.4.6.0?

Schrank:
Thanks for the PR!

Can we please be very careful with this? If I understand it correctly, it kills deployment for SW < 6.4.6

Schrank:
Hey @phizab 👋👋👋

UlrichThomasGabor:
Was fixed in 6.4.6
https://github.com/shopware/platform/commit/01c8ff86c7d8d3bee1888a26c24c9dc9b4529cbc
https://issues.shopware.com/issues/NEXT-17720

I would not remove it though, because the potential downside of replacing something wrong is basically non-existent, but for old version it will just break without the command.

UlrichThomasGabor:
> Was fixed in 6.4.6 [shopware/platform@01c8ff8](https://github.com/shopware/platform/commit/01c8ff86c7d8d3bee1888a26c24c9dc9b4529cbc) https://issues.shopware.com/issues/NEXT-17720

I would add those thinks as comments though.

phizab:
I just followed the discussion in #2710 and thought the way there was to update the recipe along shopware updates and to get a clean recipe in Deployer 7. But it's okay for me to keep the workaround for a certain time.

phizab:
Hey @Schrank 😉

Schrank:
> I just followed the discussion in #2710 and thought the way there was to update the recipe along shopware updates and to get a clean recipe in Deployer 7. But it's okay for me to keep the workaround for a certain time.

You know what: You are right. We already decided to have two recipes, so let us have one for the latest version we can achieve with a recipe 👍

UlrichThomasGabor:
Well, I think there is a difference between working around a bug, which has no down-sides of being left in place, and a recipe being unable to be used with current versions because it just does not fit the current deployment process anymore.

But I don't care that much, because this can be easily fixed by extending the recipe. Feel free if a workaround causing no harm annoys you ;)

phizab:
It doesn't bother me, too. I was only struggling with the path used in the sed command, which doesn't fit to the path in my environment (see #2754) and I saw that the bug the workaround is for got fixed in Shopware. So I thought about discussing to remove it, but don't want to break any deployments with this change.

But I still think the path "/var/www/htdocs" should be customizable, shouldn't it? Isn't it {{deploy_path}} or something like this?

@UlrichThomasGabor I reverted the commit and changed the comment. What do you think?

@Schrank Since there is only this recipe yet I and I don't want to break anything I would keep it for the moment :shrug:

Schrank:
I would love to create the second, but...

UlrichThomasGabor:
> But I still think the path "/var/www/htdocs" should be customizable, shouldn't it? Isn't it {{deploy_path}} or something like this?

Yes, this should be replaced, but I was too lazy and annoyed by the escaping to make it right.... feel free ;)

peterjaap:
I wanted to fix this but I don't have an example file to test the `sed` command against since I've upgraded my Shopware installs. Could someone supply me with one (with absolute paths)?

@phizab in the meantime, could you update the PR to append `|| true` ? That way, the `sed` command will fail silently for >= 6.4.6 installs.

So;

```
runLocally('sed -i "" -E \'s/\\\\\/var\\\\\/www\\\\\/htdocs\\\\\/releases\\\\\/[0-9]+\\\\\///g\' files/theme-config/* || true');
```

phizab:
@peterjaap Thanks for taking over and the hotfix. I've updated the `sed` command and the docs. Unfortunately I upgraded Shopware recently, too, so I don't have an example file.

peterjaap:
@antonmedv I'm fine with merging this for now.

Let's keep https://github.com/deployphp/deployer/issues/2754 open so somebody (me?) can fix the hardcoded path. But I'd need an example file for that. @Schrank ?

UlrichThomasGabor:
I can take a look at that hard-coded path when I clean up my recipe here and rebase it onto the Deployer recipe. But will take a few weeks I guess, sorry

peterjaap:
@UlrichThomasGabor I will fix it, all I need is an example file with absolute paths!

UlrichThomasGabor:
It seems I don't have one anymore.

But I guess you can easily break a shopware instance by removing [this single line](https://github.com/shopware/platform/blob/01c8ff86c7d8d3bee1888a26c24c9dc9b4529cbc/src/Storefront/Theme/ThemeFileResolver.php#L74), clear the cache `rm -Rf var/cache/*` and then run `./bin/console theme:dump` and look into `files/theme-config/*`.

Or just edit the JSON file manually:
```javascript
"storefrontEntryFilepath":"custom\/plugins\/SwagExample\/src\/Resources\/app\/storefront\/src\/main.js"
```
was something like
```javascript
"storefrontEntryFilepath":"\/var\/www\/releases\/105\/custom\/plugins\/SwagExample\/src\/Resources\/app\/storefront\/src\/main.js"
```

Because the search string is that specific, I did not bother with parsing the JSON, but just replaced every occurrence of the string.

Schrank:
```
        "storefront": {
            "path": "Resources\/app\/storefront\/src",
            "entryFilePath": "Resources\/app\/storefront\/src\/main.js",
            "webpack": null,
            "styleFiles": [
                "\/var\/www\/customer.stage.winkelwagen.de\/releases\/20211124115627\/vendor\/shopware\/storefront\/Resources\/app\/storefront\/src\/scss\/base.scss",
                "\/var\/www\/customer.stage.winkelwagen.de\/releases\/20211124115627\/vendor\/shopware\/storefront\/Resources\/app\/storefront\/src\/scss\/skin\/shopware\/_base.scss",
                "@Plugins"
            ]
        }
    },
```

peterjaap:
I continued working on this here; https://github.com/deployphp/deployer/issues/2754#issuecomment-982950659

PR #2767, opened 2021-11-19
Add fields support in Slack notification

peterjaap:
The old Slack recipe had support for fields, which looked like this;

image

See the code here; https://github.com/deployphp/recipes/blob/5f25c5e931ec35d724663b6fa710162e4fd8e3b9/recipe/slack.php#L51

This PR adds support for those fields back to the recipe.


contrib/slack.php
```diff
@@ -125,6 +125,7 @@ function checkSlackAnswer($result)
         'title' => get('slack_title'),
         'text' => get('slack_success_text'),
         'color' => get('slack_success_color'),
+        'fields' => get('slack_fields'),
```
antonmedv, line 128: We should define it before using. Or it will fail for used without slack_fields.
Add 
```php
/* Some examples
 * ```php
 * ...
 * ```
 */
set('slack_fields', []);
```

peterjaap: Wouldn't this be nicer;

```
'fields' => get('slack_fields') ?? [],
```

?

antonmedv: This will throw an exception, so get(…, []);

But better to dectare separately as we can add comment with an example and it will be added to docs by bin/docgen

peterjaap: Ok done

huanes:
I know this PR is quite old and merged long time ago, but why wasn't the `slack_fields` support added to `slack:notify`, `slack:notify:failure` and `slack:notify:rollback`?   
Seems quite useful to add some data about host/path or whatever information to all of those notifications.

PR #2785, opened 2021-11-25
deployer should exit with error status code on exception

fliespl:
Currently if exception happens inside deploy.php CI will mark job as succeeded.

This PR adds exit code to run command in case it fails with an exception.

- [X] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

### DIFF
--- a/src/Deployer.php
+++ b/src/Deployer.php
@@ -315,6 +315,8 @@ class Deployer extends Container
                 $output->setVerbosity(OutputInterface::VERBOSITY_DEBUG);
             }
             self::printException($output, $exception);
+            
+            exit(1);
         }
     }
 


src/Deployer.php
```diff
@@ -315,6 +315,8 @@ public static function run(string $version, ?string $deployFile)
                 $output->setVerbosity(OutputInterface::VERBOSITY_DEBUG);
             }
             self::printException($output, $exception);
+            
+            exit(1);
```
antonmedv, line 319: Let’s use Symfony h onds fir this.

antonmedv: Methods

antonmedv (changes requested):

src/Deployer.php
fliespl: Do you mean Command constants? Cause your current console implementation can't rely on "returning" status code.

antonmedv: Oh, I see.

antonmedv (approved):

Issue #2759, opened 2021-11-18
[v7] Deploy cleanup doesn't work as expected after failed deployments


mortenscheel:
- Deployer version:  v7.0.0-rc.2
- Deployment OS: AlmaLinux (Centos 8)

Steps to reproduce:
Config: `set('keep_releases', 5)`
1. Perform 5 succesful deployments
2. Perform 5 deployments that fail
3. Perform 5 succesful deployments

Expected: The releases folder should contain the 5 latest releases.
Actual: All 10 succesful releases remain.

The issue seems to be the way `releases_list`  works. It reads the last  `$keepReleases + 5` entries in .dep/releases_log and returns those that exist in the releases folder. So it doesn't notice the first 5 releases because they aren't mentioned in the last 10 lines of the log file. This could probably be fixed by increasing the number of log lines it reads. Or by comparing the actual release folders with the log file, in stead of the other way around.

PR #2788, opened 2021-11-29
Refactor deploy:cleanup Command

amayer5125:
Remove $runOpts since it's not actually used
Move release symlink deletion to beginning of method. This will allow the release symlink to be deleted even if deleting the old releases fails.


Schrank:
Thanks for the PR! @antonmedv does it make sense to delete everything, EXCEPT the last five releases? (making sure the list of the kept releases has really 5 unique entries?

https://github.com/deployphp/deployer/issues/2759

recipe/deploy/cleanup.php
```diff
@@ -4,26 +4,17 @@
 // Use sudo in deploy:cleanup task for rm command.
 set('cleanup_use_sudo', false);
 
-desc('Cleanups old releases');
+desc('Cleanup old releases');
```
antonmedv, line 7: Let’s keep an s at the end. To be in descriptive form.

antonmedv:
@Schrank yes, I think this makes sense. Can @amayer5125 you do it as well in this pr? More context it #2759

recipe/deploy/cleanup.php
peterjaap: But that's not correct English? You probably mean "Cleans up old releases"

antonmedv: Yes! I’m not native English bearer and will need proofread of docs after I finish. 😌

recipe/deploy/release.php
```diff
@@ -18,21 +18,13 @@
...
-    $releasesLog = [];
-    foreach (explode("\n", $data) as $line) {
-        $metainfo = json_decode($line, true);
-        if (!empty($metainfo)) {
-            $releasesLog[] = $metainfo;
-        }
-    }
-    return $releasesLog;
+    return array_filter($releaseLogs, function ($release) {
+        return ! empty($release);
```
Schrank, line 26: No need for this (I guess?) because array_filter by defaults removes everything which evaluates to false

Schrank:
I like all the array functions, but most people are not used to them. Can you please comment  them? 😅

amayer5125:
@Schrank I added comments to the `array_*` functions.

recipe/deploy/release.php
amayer5125: Updated as suggested.

recipe/deploy/release.php
```diff
@@ -44,33 +36,31 @@
         return [];
     }
 
-    // Will list only dirs in releases.
-    $ll = explode("\n", run('cd releases && ls -t -1 -d */'));
-    $ll = array_map(function ($release) {
+    // retrieve all release directories
```
antonmedv, line 39: Why this change here? Let’s separate into two PR.

antonmedv (changes requested):

recipe/deploy/release.php
amayer5125: > @Schrank yes, I think this makes sense. Can @amayer5125 you do it as well in this pr? More context it #2759

Didn't you ask me to look at it here? Maybe I didn't need to refactor releases_list, but just releases_log to fix the issue? I'll separate it out into another PR.

recipe/deploy/cleanup.php
amayer5125: Is this okay the way it is, or did you still want me to revert it?

recipe/deploy/release.php
amayer5125: I moved the changes you mentioned to another PR (#2845).

antonmedv: Thanks, got out of context. Will review it today. 

Thanks for your contributions!

PR #2789, opened 2021-11-29
Laravel: Add Route Cache to Deploy Tasks

amayer5125:
Caching Routes has been recommended since the early 5.x days to improve performace or production apps.
https://laravel.com/docs/8.x/deployment#optimizing-route-loading

Reorder optimization commands to same order as Laravel docs.

### DIFF
--- a/docs/recipe/laravel.md
+++ b/docs/recipe/laravel.md
@@ -376,8 +376,9 @@ This task is group task which contains next tasks:
 * [deploy:prepare](/docs/recipe/common.md#deployprepare)
 * [deploy:vendors](/docs/recipe/deploy/vendors.md#deployvendors)
 * [artisan:storage:link](/docs/recipe/laravel.md#artisanstoragelink)
-* [artisan:view:cache](/docs/recipe/laravel.md#artisanviewcache)
 * [artisan:config:cache](/docs/recipe/laravel.md#artisanconfigcache)
+* [artisan:route:cache](/docs/recipe/laravel.md#artisanroutecache)
+* [artisan:view:cache](/docs/recipe/laravel.md#artisanviewcache)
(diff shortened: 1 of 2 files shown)


peterjaap:
Weird that the tasks were there but they weren't called?

lorisleiva:
It was always available for you to use, it simply wasn't part of the default Laravel deployment task.

Whilst I agree it's nice to have it by default, you should avoid relying on the default `deploy` task for your deployment flow anyway. Even if you're 100% happy with the default `deploy` task, I'd recommend copy/pasting it in your own `deploy.yaml` or `deploy.php` file so you have control over them. 

If you don't, you take the risk of your deployment flow changing next time you upgrade Deployer. This PR is a good example of how this could happen.

peterjaap:
You should lock the Deployer version in your composer.json so no unexpected
updates happen. No need to copy paste the recipe in its entirety.

lorisleiva:
You can of course do that too but I don't think this is good advice.

It's better for your deployment flow to be independent of Deployer's version as newer versions could bring new features and bug fixes.

peterjaap:
But if you lock Deployer to a specific version (or maybe even commit) that
would never happen without you knowing. That is the whole reasoning behind
composer(.lock).

antonmedv:
I think both approaches are valid.

lorisleiva:
@peterjaap Yeah I get that. ☺️

I just think it's a shame to lock your dependency to a specific version and therefore not relying on Semver to provide little enhancements and bug fixes if you need them.

But if that works for you, great! 👍

peterjaap:
You can always lock it with ^ and then check the composer diff to see what
changed. We use Renovate Bot to automate that process.

PR #2793, opened 2021-11-29
Sentry recipe fix regarding `git_version_command` with `update_code_strategy` set to `archive`

DamienHarper:
Makes use of `git_version_command` (if defined) even when `update_code_strategy` is set to `archive`


dpfaffenbauer:
💪

DamienHarper:
@antonmedv does this PR look good for you?

PR #2811, opened 2021-12-04
Add dry-run option that outputs commands instead of running them

midweste:
- [ ] Bug fix #…?
- [x] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [x] Docs added?

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -431,9 +431,9 @@ function run(string $command, ?array $options = [], ?int $timeout = null, ?int $
  */
 function runLocally(string $command, ?array $options = [], ?int $timeout = null, ?int $idle_timeout = null, ?string $secret = null, ?array $env = null, ?string $shell = null): string
 {
-    $dryRun = Deployer::get()->console->getDefinition()->hasOption('dry-run');
+    $dryRun = get('dry-run', false) || input()->getOption('dry-run');
     if ($dryRun) {
-        writeln($command);
+        writeln('[<info>dry-run</info>] ' . $command);
         return $command;
     }
 
--- a/src/functions.php
+++ b/src/functions.php
@@ -339,6 +339,12 @@ function within(string $path, callable $callback)
  */
 function run(string $command, ?array $options = [], ?int $timeout = null, ?int $idle_timeout = null, ?string $secret = null, ?array $env = null, ?bool $real_time_output = false, ?bool $no_throw = false): string
 {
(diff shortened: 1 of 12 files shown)


src/functions.php
```diff
@@ -339,6 +339,12 @@ function within(string $path, callable $callback)
  */
 function run(string $command, ?array $options = [], ?int $timeout = null, ?int $idle_timeout = null, ?string $secret = null, ?array $env = null, ?bool $real_time_output = false, ?bool $no_throw = false): string
 {
+    $dryRun = Deployer::get()->console->getDefinition()->hasOption('dry-run');
+    if ($dryRun) {
+        writeln($command);
+        return $command;
```
antonmedv, line 345: What if tasks depends on output? For example 
if (test(….)) 
 Do something 
else
  Do another thing

midweste: Thanks for the feedback and pointing that out.  I'm still trying to get my head around moving from 6 to 7.  I'll take a look at test and testLocally to see if there's an elegant solution there.  In the current state the pull request is less like a dry run and more like a command list.

midweste: Ideally this command would be completely non-destructive.  I'm not sure if allowing the test commands to run would be a good idea, what do you think here?  Because the commands could be anything, what if users wrote tests that modify files/etc.

What's your thoughts?

antonmedv: Tasks can also contain php code. It can rm files for example. Better solution will be to analyse php code by AST.

midweste: Just to understand a bit better and to keep within your project guidelines, are you thinking to analyze the behavior of the test command and limit anything that modifies, and allow pure tests to execute as usual?  

In effect:
- allow standard bash tests to run if matching a regex of test patterns ( [ -d "file.txt"  ] or double bracket tests [[ ( -d "$HOME" ) && ( -w "$HOME" ) ]] )
- use AST to analyze php code and limit functions that could modify files such as file_put_contents, unlink, etc

antonmedv: Something like this. But this is really hard and not looks like feasible. Rake a look at deploy:release task for example.

midweste: Command is still useful in development, but its not going to work across the board with testing being allowed to run any code.  For flexibility it makes sense that it can, but it presents some real issues with ever being able to test complex deployments.

I'm going to close this until a better solution presents itself.  Thank you for the feedback, I appreciate it.

PR #2820, opened 2021-12-08
Issue 2570: Add an option to deploy only a sub-folder of the repository

### LINKED ISSUE #2570: Add an option to deploy only a sub-folder of the repository
In some projects, we want to deploy only a sub-folder of a git repository (as a deployment root).

Until now, I was using a hook like this (with all the problems associated with the use of "git filter-branch")

```php
task('repository:filter_branch', function () {
    if (has('deploy_repository_path')) {
        run("cd {{release_path}} && {{bin/git}} filter-branch --subdirectory-filter {{deploy_repository_path}} > /dev/null");
    }
})
```

With the new version `7`, this can be simplified by simply adding this option within the "`git archive`" command of `deploy:update_code` :

```php
run("$git archive $at:{{deploy_repository_path}} | tar -x -f - -C {{release_path}} 2>&1");
```

If you are interested, I can try to push a PR.

matsn0w:
- [ ] Bug fix #…?
- [x] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [x] Docs added?

Fixes #2570 

This feature adds an option to specify a sub directory in the repository of the project. This can be useful when you only want to deploy a single folder in the repo.

An example use case would be a repository with a separate frontend and backend, each living in their own folder.

When no `sub_directory` is specified, the entire repository will be deployed. Works only when `update_code_strategy` is set to `archive`, which is the default option.

### DIFF
--- a/docs/recipe/deploy/update_code.md
+++ b/docs/recipe/deploy/update_code.md
@@ -43,7 +43,7 @@ Can be one of:
 
 
 ### sub_directory
-[Source](https://github.com/deployphp/deployer/blob/master/recipe/deploy/update_code.php#L25)
+[Source](https://github.com/deployphp/deployer/blob/master/recipe/deploy/update_code.php#L31)
 
 Specifies a sub directory within the repository to deploy.
 Works only when [`update_code_strategy`](#update_code_strategy) is set to `archive` (default).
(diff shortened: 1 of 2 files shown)


matsn0w:
@antonmedv does this feature need additional tests?

PR #2822, opened 2021-12-08
Don't add remote user to setfacl command, if it doesn't exist as an os user

ElAberino:
Only add the the remote user to setfacl command if it exists as an os user.
Based on this discussion:
https://github.com/deployphp/deployer/discussions/2787

- [x] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

      $ php bin/docgen

### DIFF
--- a/recipe/deploy/writable.php
+++ b/recipe/deploy/writable.php
@@ -109,7 +109,7 @@ task('deploy:writable', function () {
             $setFaclUsers = "-m u:\"$httpUser\":rwX";
             // Check if remote user exists, before adding it to setfacl
             $remoteUserId = run("id -u $remoteUser &>/dev/null 2>&1 || exit 0");
-            if (is_int($remoteUserId)) {
+            if (!empty($remoteUserId)) {
                 $setFaclUsers .= " -m u:$remoteUser:rwX";
             }
             if (empty($sudo)) {
--- a/recipe/deploy/writable.php
+++ b/recipe/deploy/writable.php
@@ -106,6 +106,12 @@ task('deploy:writable', function () {
             run("$sudo chmod +a \"$httpUser allow delete,write,append,file_inherit,directory_inherit\" $dirs");
             run("$sudo chmod +a \"$remoteUser allow delete,write,append,file_inherit,directory_inherit\" $dirs");
         } elseif (commandExist('setfacl')) {
+            $setFaclUsers = "-m u:\"$httpUser\":rwX";
+            // Check if remote user exists, before adding it to setfacl
+            $remoteUserId = run("id -u $remoteUser &>/dev/null 2>&1 || exit 0");
+            if (is_int($remoteUserId)) {
+                $setFaclUsers .= " -m u:$remoteUser:rwX";
+            }
             if (empty($sudo)) {
                 // When running without sudo, exception may be thrown
                 // if executing setfacl on files created by http user (in directory that has been setfacl before).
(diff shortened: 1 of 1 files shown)


ElAberino:
@antonmedv is there anything I need to do in order to get approval for the workflows?

antonmedv:
Sorry for the delay. Will check tomorrow

antonmedv:
The whole check is to verify dirs will be writable to remote user. 

Task should fail in remote user not found.

ElAberino:
@antonmedv Why does the remote user have to exist an an acl user? With my hoster that is not the case and this leads to the error described in https://github.com/deployphp/deployer/discussions/2787

antonmedv:
Ah, yes. I confused it with www-data user.

recipe/deploy/writable.php
```diff
@@ -106,6 +106,12 @@
             run("$sudo chmod +a \"$httpUser allow delete,write,append,file_inherit,directory_inherit\" $dirs");
             run("$sudo chmod +a \"$remoteUser allow delete,write,append,file_inherit,directory_inherit\" $dirs");
         } elseif (commandExist('setfacl')) {
+            $setFaclUsers = "-m u:\"$httpUser\":rwX";
+            // Check if remote user exists, before adding it to setfacl
+            $remoteUserId = run("id -u $remoteUser &>/dev/null 2>&1 || exit 0");
```
antonmedv, line 111: Lets use test() func here.

antonmedv (changes requested):

recipe/deploy/writable.php
```diff
@@ -108,8 +108,8 @@
         } elseif (commandExist('setfacl')) {
             $setFaclUsers = "-m u:\"$httpUser\":rwX";
             // Check if remote user exists, before adding it to setfacl
-            $remoteUserId = run("id -u $remoteUser &>/dev/null 2>&1 || exit 0");
-            if (!empty($remoteUserId)) {
+            $remoteUserExists = test("id -u $remoteUser &>/dev/null 2>&1 || exit 0");
```
antonmedv, line 111: Let's also delete `exit 0` and use 
```
run('...', ['no_throw' => true]);
```

antonmedv (changes requested):

recipe/deploy/writable.php
ElAberino: @antonmedv Does that mean I shall replace `test()` with `run()` again, as `test()` doesn't have any additional options? 

If I just remove `|| exit 0` and use it with `test()`, it keeps working. I still get `true` if the user exists and `false` if it doesn't.

antonmedv: Ohh, I see, sorry for misleading. Let me think a bit. Maybe test() should also have options?

ElAberino: @antonmedv 
I am not sure if that's necessary.  I tried all variants to see how it works:

If the user does exist
1. `run("id -u $remoteUser &>/dev/null 2>&1")`: Returns empty string -> the deployment continues
2. `run("id -u $remoteUser &>/dev/null 2>&1", ['no_throw' => true])`: Returns empty string -> the deployment continues
3. `test("id -u $remoteUser &>/dev/null 2>&1")`: Returns true -> the deployment continues

If the user doesn't exist
1. `run("id -u $remoteUser &>/dev/null 2>&1")`: Throws exception -> the deployment is abortet with exit code 1.
2. `run("id -u $remoteUser &>/dev/null 2>&1", ['no_throw' => true])`: Returns empty string -> the deployment continues
3. `test("id -u $remoteUser &>/dev/null 2>&1")`: Returns false -> the deployment continues

So i'd say test() works as expected, but I am not aware of side effects you might see.

antonmedv (approved):

PR #2837, opened 2021-12-14
Fix cachetool PHP 7 compatibility, fixes #2836

### LINKED ISSUE #2836: cachetool breaks PHP 7.4 Compatibility
- Deployer version: v7.0.0-beta.24
- Deployment OS: Ubuntu 20.04

If you add the following to your `deploy.php` and try to deploy a PHP 7.4 project, you will get an error.

      after('deploy:symlink', 'cachetool:clear:opcache');

The error looks like this:

    [my.host.com] run if [ -f /my/path/releases/55/cachetool.phar ]; then echo +right; fi
    [my.host.com] run cd /my/path/releases/55 && curl -sLO https://github.com/gordalina/cachetool/releases/latest/download/cachetool.phar
    [my.host.com] run cd /my/path/releases/55 && /usr/bin/php7.4 /my/path/releases/55/cachetool.phar opcache:reset --fcgi=/var/run/my-socket.sock
    [my.host.com] Composer detected issues in your platform:
    [my.host.com] Your Composer dependencies require a PHP version ">= 8.0.0". You are running 7.4.26.
    [my.host.com] PHP Fatal error:  Composer detected issues in your platform: Your Composer dependencies require a PHP version ">= 8.0.0". You are running 7.4.26. in phar:///my/path/releases/55/cachetool.phar/vendor/composer/platform_check.php on line 24
    [my.host.com]  error  in cachetool.php on line 80:
    [my.host.com] exit code 255 (Unknown error)

The issue is that the cachetool is always used in the latest version:

https://github.com/deployphp/deployer/blob/2d16b6d2a808df72d7c4c9067637d10067ee99e8/contrib/cachetool.php#L57

Unfortunately, the latest version dropped support for PHP 7, see https://github.com/gordalina/cachetool#compatibility.

I could not find any official information which PHP versions Deployer officially supports, but I think it should at least support all still maintained PHP versions (even if that PHP version only gets security fixes). So my suggested fix for now would be to fix the cachetool to version 7 and not use the latest version. Happy to send a PR.

What do you think?

sprankhub:
- [X] Bug fix #2836?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

      $ php bin/docgen

### DIFF
--- a/contrib/cachetool.php
+++ b/contrib/cachetool.php
@@ -51,10 +51,11 @@ http://gordalina.github.io/cachetool/
 namespace Deployer;
 
 set('cachetool', '');
+set('cachetool_url', 'https://github.com/gordalina/cachetool/releases/download/7.0.0/cachetool.phar');
 set('cachetool_args', '');
 set('bin/cachetool', function () {
     if (!test('[ -f {{release_or_current_path}}/cachetool.phar ]')) {
-        run("cd {{release_or_current_path}} && curl -sLO https://github.com/gordalina/cachetool/releases/download/7.0.0/cachetool.phar");
+        run("cd {{release_or_current_path}} && curl -sLO {{cachetool_url}}");
     }
     return '{{release_or_current_path}}/cachetool.phar';
 });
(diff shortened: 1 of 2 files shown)


Schrank:
I added a configuration, would you mind, test that? :-)

sprankhub:
I did not test it, but it looks perfectly fine to me :+1: Thanks!

PR #2845, opened 2021-12-17
Refactor deploy:releases_list

amayer5125:
No functional change, but I feel this change makes the code easier to understand.


antonmedv:
This part of the code is not really well tested. So it's dangerous to make changes, even refactoring and clean code only. 

Thanks for effort, but let's not merge or this part of code need more tests.

PR #2851, opened 2021-12-19
Contao recipe

richardhj:
This PR adds the recipe for the [Contao CMS](https://contao.org/en/) which is based on the Symfony recipe.


Schrank:
Thanks @richardhj! Are you done? If yes I would take a view on it.

recipe/contao.php
```diff
@@ -0,0 +1,73 @@
...
+desc('Run Contao migrations');
+task('contao:migrate', function () {
+    run('{{bin/php}} {{bin/console}} contao:migrate {{console_options}}');
+});
+
+// Optional task to download the `contao-manager.phar.php` in the public path. Make sure to
+// set a password on first access!
+desc('Download the Contao Manager');
+task('contao:manager:download', function () {
+    run('curl -LsO https://download.contao.org/contao-manager/stable/contao-manager.phar && mv contao-manager.phar {{release_or_current_path}}/{{public_path}}/contao-manager.phar.php');
```
Schrank, line 47: I have no clue, what this file is, but if it lives in public, maybe delete it after deployment?

richardhj: This file is definitely there to stay :-) Can be reached under https://example.org/contao-manager.phar.php.

richardhj:
@Schrank I would be more than happy if you review this PR :-)

PR #2852, opened 2021-12-20
Fix for #2850

### LINKED ISSUE #2850: `whoami` fails on some hosters
- Deployer version: 7.0.0-rc3
- Deployment OS: I don't know...

I have a hoster where `whoami` fails, but it is required here:
https://github.com/deployphp/deployer/blob/cfcb963ead5f993157d20478c8332c0c93908337/recipe/deploy/writable.php#L68

When I set `remote_user` it is used for the ssh connection as well, but the ssh username is a different name than the one owning the files...

Actually, I have set `writable_mode` to `chmod` because I expected some kind of problem and the `chmod` part of this task does not even use `$remoteUser`.

I propose setting `$remoteUser` in the acl-block:
https://github.com/deployphp/deployer/blob/cfcb963ead5f993157d20478c8332c0c93908337/recipe/deploy/writable.php#L101-L132
And change the message here to some generic text:
https://github.com/deployphp/deployer/blob/cfcb963ead5f993157d20478c8332c0c93908337/recipe/deploy/writable.php#L96

UlrichThomasGabor:
This fixes #2850.

I removed the `catch` block completely, because it is untypical (all other calls of `run` do not have one) and the message can be misguiding (`$sudo`).

### DIFF
--- a/recipe/deploy/writable.php
+++ b/recipe/deploy/writable.php
@@ -63,11 +63,6 @@ task('deploy:writable', function () {
     $recursive = get('writable_recursive') ? '-R' : '';
     $sudo = get('writable_use_sudo') ? 'sudo' : '';
 
-    $remoteUser = get('remote_user', false);
-    if (empty($remoteUser)) {
-        $remoteUser = run('whoami');
-    }
-
     if (empty($dirs)) {
         return;
     }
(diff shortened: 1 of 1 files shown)


recipe/deploy/writable.php
```diff
@@ -87,18 +82,17 @@
         // -L   traverse every symbolic link to a directory encountered
         run("$sudo chown -L $recursive $httpUser $dirs");
     } elseif ($mode === 'chgrp') {
-        try {
```
antonmedv, line 90: Why this block was removed?

antonmedv (changes requested):

recipe/deploy/writable.php
UlrichThomasGabor: Because I do not consider it really helpful and it's not common in the codebase to give hints why a command failed.

Basically no other command gives feedback when it fails, so why here?

And, the commands can fail for many reasons (lost connection, wrong group, broken server, wrongly configured sudo, …), so why assume that the problem is that the user is not part of the group? What about all the other cases where the message may be just misleading, because the problem has nothing to do with the groups of the user?

antonmedv: Make sense.

antonmedv (approved):

PR #2877, opened 2022-01-03
set `phar.readonly=true` because building will fail without it anyway

UlrichThomasGabor:


bin/build
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env php
+#!/usr/bin/env php -d phar.readonly=false
```
antonmedv, line 1: Unfortunately extra args not supported in some env implementations.

antonmedv (changes requested):

bin/build
antonmedv: Possibly to detect missing option and rerun command with this args. Same as composer does.

UlrichThomasGabor: Ah. ... well I don't see an easy variant to call oneself again including parameters (`PHP_BINARY` is the executable, forwarding all pipes may be done by `proc_open`, remains constructing the rest of the options for the executable...). This looks like it can be done wrong in various ways. I would vote for that this problem does not need to be fixed in the script, but can be fixed by hand (modify the shebang or change php.ini). I changed the code to check for the setting and give some hints for a fix.

PR #2879, opened 2022-01-03
Implemented basic version of replacing tasks, fixes #2878

### LINKED ISSUE #2878: Overwriting existing tasks silently drops before/after hooks
- Deployer version: 7.0.0-rc3

Overwriting an existing task by using `task()`, for example:
```php
task('deploy:update_code', static function () {
        // whatever
    ]);
});
```
results in removing all before/after hooks because the collection just replaces the whole task:
https://github.com/deployphp/deployer/blob/10a2f43b0a0b625c661741154230fa50a4abb5c1/src/functions.php#L198

I think this is unexpected and caused a silent issue on my side.

Can this be changed to either use `Task::setCallback()` when a task already exists or at least print a warning when an existing task is overwritten?

UlrichThomasGabor:
- [x] Bug fix #2878?
- [ ] New feature?
- [x] BC breaks? Behavior changed (hooks are preserved) and replacing a `Task` with a `GroupTask` (or other way around) is now forbidden. Could be fixed, but I think it is cleaner to require removing a task and then re-add it.
- [ ] Tests added?
- [ ] Docs added?

### DIFF
--- a/src/Collection/Collection.php
+++ b/src/Collection/Collection.php
@@ -44,6 +44,15 @@ class Collection implements Countable, IteratorAggregate
         $this->values[$name] = $object;
     }
 
+    public function remove(string $name): void
+    {
+        if ($this->has($name)) {
+            unset($this->values[$name]);
+        } else {
+            $this->throwNotFound($name);
+        }
+    }
+
     public function count(): int
     {
         return count($this->values);
--- a/src/Task/GroupTask.php
+++ b/src/Task/GroupTask.php
@@ -43,4 +43,9 @@ class GroupTask extends Task
     {
         return $this->group;
     }
+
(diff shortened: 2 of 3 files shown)


UlrichThomasGabor:
@antonmedv Something like that?

antonmedv:
Let’s try to fix overriding with group task as well.

UlrichThomasGabor:
Then functionality is needed to create a new `GroupTask` object from an existing `Task` object (and the other way around), i.e. extend both classes with static methods to create an instance from a given object.

If any of the classes is modified and further attributes are added, these methods must be changed as well (and forgetting to adapt the methods will lead to weird bugs).

This is what you want to have? Or do you have something else in mind?

PR #2880, opened 2022-01-04
Update release.php

luca-rath:
- [x] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [x] Docs added?

I'm not sure if this is really a bug or not, but if this line is missing, deployer always tries to deploy as release `1` and doesn't take the contents of the `.dep/latest_release` file into account

### DIFF
--- a/docs/recipe/deploy/release.md
+++ b/docs/recipe/deploy/release.md
@@ -16,35 +16,35 @@ The name of the release.
 
 
 ### releases_log
-[Source](https://github.com/deployphp/deployer/blob/master/recipe/deploy/release.php#L14)
+[Source](https://github.com/deployphp/deployer/blob/master/recipe/deploy/release.php#L16)
 
 Holds releases log from `.dep/releases_log` file.
 
 
 
 ### releases_list
-[Source](https://github.com/deployphp/deployer/blob/master/recipe/deploy/release.php#L39)
+[Source](https://github.com/deployphp/deployer/blob/master/recipe/deploy/release.php#L41)
 
 Return list of release names on host.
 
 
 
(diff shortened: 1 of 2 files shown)


Schrank:
Hey, thanks for your contribution! 

May I ask how you found this problem? Because in theory while checking for the release number you should be in the directory 😅

luca-rath:
Hi, thanks for your very quick response! I don't know why I'm not in that directory ... I guess I should be in that directory because `get('releases_list')` is executed directly before `get('release_name')`, right? Do you have any tips for me on how to further debug that problem?

luca-rath:
Just changed the implementation of `release_name` to the following:

```php
set('release_name', function () {
    $pwd = run('pwd');
    output()->writeln("$pwd");

    cd('{{deploy_path}}');

    $latest = run('cat .dep/latest_release || echo 0');
    return strval(intval($latest) + 1);
});
```

This outputs `/home/deploy`

luca-rath:
Btw, using deployer `v7.0.0-rc.3`

Schrank:
Something changes the current directory for you, this must happen before release_name runs. But better safe than sorry. If all test are green, I'm happy to merge it. Thanks for the docgen update!

antonmedv:
Let’s use within() here as it may affect different tasks.

luca-rath:
@antonmedv changed it

luca-rath:
@antonmedv I don't want to bother you, but it would be really great if you could have a look at this again - thanks 🙂

luca-rath:
thanks a lot! 😊

PR #2886, opened 2022-01-07
CakePHP recipe, use the configured PHP version

steinkel:
- [ ] Bug fix #…?
- [X] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

Tweak the CakePHP recipe to use the configured PHP version instead of the target system default. Useful when multiple PHP versions installed in the deploy target.

### DIFF
--- a/recipe/cakephp.php
+++ b/recipe/cakephp.php
@@ -25,15 +25,15 @@ set('shared_files', [
  * Create plugins' symlinks
  */
 task('deploy:init', function () {
-    run('{{release_or_current_path}}/bin/cake plugin assets symlink');
+    run('{{bin/php}} {{release_or_current_path}}/bin/cake.php plugin assets symlink');
 })->desc('Initialization');
 
 /**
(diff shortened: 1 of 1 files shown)


recipe/cakephp.php
```diff
@@ -25,15 +25,15 @@
  * Create plugins' symlinks
  */
 task('deploy:init', function () {
-    run('{{release_or_current_path}}/bin/cake plugin assets symlink');
+    run('{{bin/php}} {{release_or_current_path}}/bin/cake.php plugin assets symlink');
```
antonmedv, line 28: .php?

steinkel: yes, this file https://github.com/cakephp/app/blob/4.x/bin/cake.php called directly instead of the wrapper shell provided. It would allow you to use the right version of PHP configured in {{bin/php}} instead of relying on the shell to locate the default configured php version in the target env.

PR #2891, opened 2022-01-10
Allow for symfony 6

lsv:
- composer.lock has been ignored with .gitignore - so it doesnt get updated because of a different php version.
- composer.json updated to allow for symfony/* ^5|^6
- workflow updated so it run tests with composer prefer-lowest and prefer-dist, so both ^5 and ^6 gets tested.
- symfony ^6 - requires PHP 8, therefor platform in composer.json is removed

Nothing changed in docs, no BC breaks - all tests passes with symfony/* 5.0 and symfony/* 6.0.

No changed to docs.


antonmedv:
Please use deployer/dist

PR #2906, opened 2022-01-17
Fix Magento version detection, fixes #2905

### LINKED ISSUE #2905: M2: Config Import Not Executed For Patch Versions
- Deployer version: master
- Deployment OS: Ubuntu 20.04

The Magento version detection does not detect patch versions correctly:

https://github.com/deployphp/deployer/blob/b6d646d8e34f19df10e9b8edfa1c3a8c497809d3/recipe/magento2.php#L62

This will detect `Magento CLI 2.4.3-p1` incorrectly as `1` due to the `-p1` suffix. Hence, no config import will be executed:

https://github.com/deployphp/deployer/blob/b6d646d8e34f19df10e9b8edfa1c3a8c497809d3/recipe/magento2.php#L109-L112

This may lead to an error after the deployment.

I am working on a fix right now.

sprankhub:
- [X] Bug fix #2905
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

      $ php bin/docgen

### DIFF
--- a/recipe/magento2.php
+++ b/recipe/magento2.php
@@ -59,8 +59,8 @@ set('clear_paths', [
 set('magento_version', function () {
     // detect version
     $versionOutput = run('{{bin/php}} {{release_or_current_path}}/bin/magento --version');
-    preg_match('/(\d+\.?)+$/', $versionOutput, $matches);
-    return $matches[0] ?? "2.0";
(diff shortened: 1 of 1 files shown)


Schrank:
Thanks!

PR #2928, opened 2022-01-30
Separate upgrades pages

kicaj:
- [ ] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [x] Docs added?

Proposal for separating pages about upgrades.


antonmedv:
Unforcentally this can be done as Google already indexed UPGRADE page. So let's not change URL.

PR #2946, opened 2022-02-04
Prepend the relevant bin/php

trsteel88:
At the moment if bin/php is modified and composer exists at a global level, it is executed with the global php. We need it to execute with the modified bin/php version.

### DIFF
--- a/recipe/deploy/vendors.php
+++ b/recipe/deploy/vendors.php
@@ -14,7 +14,7 @@ set('bin/composer', function () {
     }
 
     if (commandExist('composer')) {
-        return which('composer');
+        return '{{bin/php}} '.which('composer');
     }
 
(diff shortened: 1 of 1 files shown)


trsteel88:
Also, just looking at this:

```php
    return '{{bin/php}} {{deploy_path}}/.dep/composer.phar';
}
````

Should deployer update composer to ensure it's the latest version? Otherwise if it's installed at the end of this function it will become outdated.

e.g.

```php
    run('{{bin/php}} {{deploy_path}}/.dep/composer.phar self-update');

    return '{{bin/php}} {{deploy_path}}/.dep/composer.phar';
}
```

antonmedv:
Fixed in https://github.com/deployphp/deployer/pull/2967

PR #2950, opened 2022-02-07
Adding PATCH to Httpie

CaptainFalcon92:
Hello.

I've quickly added PATCH method to Httpie.
I need this to update messages in webhooks integrations (discord).

I currently respected the existing scheme, but i do think it would be better to allow more (any) methods by some refactoring.
Still i'd rather have it discussed here upfront before going further.

### DIFF
--- a/src/Utility/Httpie.php
+++ b/src/Utility/Httpie.php
@@ -43,6 +43,14 @@ class Httpie
         $http->url = $url;
         return $http;
     }
+    
+    public static function patch(string $url): Httpie
+    {
+        $http = new self;
+        $http->method = 'PATCH';
+        $http->url = $url;
+        return $http;
+    }
 
     public function query(array $params): Httpie
     {


CaptainFalcon92:
It would probably be more efficient to refactor into a single method.

```php
public function __construct(string $url, string $method)
{
    //    
    // cURL check...
    //

   $this->url = $url;
   $this->method = $method;
}

public static function get(string $url) : self
{
    return new self($url, 'GET');
}

public static function post(string $url) : self
{
    return new self($url, 'POST');
}

public static function put(string $url) : self
{
    return new self($url, 'PUT');
}

public static function patch(string $url) : self
{
    return new self($url, 'PATCH');
}
```

Let me know if your are willing to accept such evolution, and i'd commit to do the change.

Thank you, bye.

CaptainFalcon92:
But there is one sure thing - currently i can't do but _get_ or _post_ as the property's private and there is no setter.

antonmedv:
Let’s try to keep compatibility as much as possible. Adding new methods is ok. Opening private fields is also ok, but check out new function called fetch.

CaptainFalcon92:
I will check `fetch` before. Thank you.

PR #2967, opened 2022-02-17
Make sure to use the correct PHP binary with existing Composer

Toflar:
Not sure if this is correct but otherwise, the wrong PHP binary will be used.

### DIFF
--- a/recipe/deploy/vendors.php
+++ b/recipe/deploy/vendors.php
@@ -14,7 +14,7 @@ set('bin/composer', function () {
     }
 
     if (commandExist('composer')) {
-        return '{bin/php}}' . which('composer');
+        return '{{bin/php}}' . which('composer');
     }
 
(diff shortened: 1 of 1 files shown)


Schrank:
@antonmedv you changed this three months ago, but I would agree with @Toflar that we want to run composer with the correct php version, which can be configured by using bin/php.

https://github.com/deployphp/deployer/commit/45e4708aad31931805e37387ccb7940cae9c9149

Toflar:
Actually the change 3 months ago caused my deployment to suddenly fail ;)

Schrank:
Good to know, you only need it every three months 🙈

Schrank:
@Toflar You know for a quick-fix you can just overwrite the task?

clxmstaab:
duplicate of https://github.com/deployphp/deployer/pull/2946 ?

PR #2994, opened 2022-03-03
Push to v8

antonmedv:


src/functions.php
```diff
@@ -194,8 +194,27 @@ function task(string $name, $body = null): Task
...
+    if ($deployer->tasks->has($name)) {
+        // If task already exists, try to replace.
+        $existingTask = $deployer->tasks->get($name);
+        if (get_class($existingTask) !== get_class($task)) {
+            // There is no "up" or "down"casting in PHP.
+            throw new \Exception('Tried to replace a Task with a GroupTask or vice-versa. This is not supported. If you are sure you want to do that, remove the old task `Deployer::get()->tasks->remove(<taskname>)` and then re-add the task.');
+        }
+        if ($existingTask instanceof Task) {
+            $existingTask->setCallback($body);
+        } elseif ($existingTask instanceof GroupTask) {
```
luigel, line 206: I'm having a problem with this. According to phpstan the elseif is unreachable even though the `$existingTask` is `GroupTask` it will be true with `$existingTask instanceof Task`

PR #3010, opened 2022-03-09
Make the importer extendable

qzminski:
Although it seems to be a new feature, I think this could be treated as a bugfix.

I wanted to extend (not override) the `shared_dirs` configuration from YAML file, but it occurred that the `Importer` uses the `set()` function instead of `add()`. Thus, I wanted to extend the class, because the methods are protected and not private, but it was not possible because it uses `self::` instead of `static::`.

```yaml
config:
    shared_dirs:
        - my_extra_shared_folder
```

This pull requests makes the `Importer` class extendable, for example:

```php
class ExtendedImporter extends \Deployer\Importer\Importer
{
    protected static function config(array $config)
    {
        foreach ($config as $key => $value) {
            add($key, $value);
        }
    }
}

ExtendedImporter::import('deploy-hosts.yml');
```

### DIFF
--- a/src/Importer/Importer.php
+++ b/src/Importer/Importer.php
@@ -63,9 +63,9 @@ class Importer
                     }
                 });
             } else if (preg_match('/\.ya?ml$/i', $path)) {
-                static::$recipeFilename = basename($path);
-                static::$recipeSource = file_get_contents($path);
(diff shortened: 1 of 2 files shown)


antonmedv:
Cool. A test would be nice.

qzminski:
@antonmedv unit test added :+1:

PR #2864, opened 2021-12-28
Added Symfony 6 support

vudaltsov:
- [ ] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

PR #3050, opened 2022-03-17
Widen support to Symfony 5 and Symfony 6

janmartendeboer:
- Add support for Symfony 6 components
   - `symfony/console`
   - `symfony/process`
   - `symfony/yaml`
- Update lock file for `symfony/console`, `symfony/process` and
`symfony/yaml`


Tobion:
Duplicate of #2918, #2891 and #2864.
But I'm still in favor of doing this. If there is no extra work required to support symfony 5 and 6, then there is also no reason to forbid it.

PR #3072, opened 2022-03-26
Add --with-deletes flag to contao:migrate task

qzminski:
- [x] New feature?

Since Contao 4.13 the `contao:migrate` performs the automatic database backup task, so I think we should enable `--with-deletes` flag by default.

/cc @richardhj

### DIFF
--- a/docs/recipe/contao.md
+++ b/docs/recipe/contao.md
@@ -47,6 +47,16 @@ Run Contao migrations.
 
 This task updates the database. A database backup is saved automatically as a default.
 
+#### --with-deletes
+
+To automatically drop the obsolete database structures, you can override the task as follows:
+
+```php
+task('contao:migrate', function () {
+    run('{{bin/php}} {{bin/console}} contao:migrate --with-deletes {{console_options}}');
+});
+```
+
 
 ### contao:manager:download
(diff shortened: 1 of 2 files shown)


richardhj:
This is a no-go 🙅 in production for me 😊

qzminski:
Why not? If you'd like to keep the database backups, simply add `var/backups` directory to the shared dirs. It's incorrect to keep the old and irrelevant structures in your database…

@Toflar what do you think?

richardhj:
It happened to me that relevant data was deleted, either because 
- you wrongfully removed an extension and re-install later,
- you removed a field from the DCA and realized afterward that some part was missing,
- the Doctrine mapping in Contao was defective thus removing all entity-related data (or MetaModels similar).

Each case happened to me like this or similar.

Deleting data from the database should be opt-in, not opt-out. Even the install tool does not default to deletes. You can overwrite the task :-)

Toflar:
Should definitely be opt-out, otherwise it would be the default in the command itself. So 👎 on this as well. Might add a configuration flag for that in order not to require replacing the whole task, though.

qzminski:
I have reverted the recipe command, but added the snippet in the docs. I hope that's a reasonable consensus 🙂

richardhj:
I feel like that the Contao documentation should be the place for such extensions :)

PR #3078, opened 2022-03-29
Improve vendors recipe

nlemoine:
- [x] New features
Sometimes, project may need to install vendors from multiple directories. Although it's quite easy to add them in your `deploy.php` file, I think it would be easier if it was included in the vendors recipe.

This PR also includes some code to self update composer from time to time (only when using phar version, the duration can be defined and this feature can be disabled) because once composer is downloaded, it will never upgrade.

- [ ] BC breaks?

None that I'm aware of. Default config includes "{{release_or_current_path}}".

- [ ] Tests added?

No, could find tests against recipes.

- [ ] Docs added?

No, will do if you feel this PR is valuable.


richardhj:
Please split the different features into distinct PRs for better review.

recipe/deploy/vendors.php
```diff
@@ -1,15 +1,33 @@
...
+
 set('composer_options', '--verbose --prefer-dist --no-progress --no-interaction --no-dev --optimize-autoloader');
 
-// Returns Composer binary path in found. Otherwise try to install latest
+// Returns Composer binary path if found. Otherwise try to install latest
 // composer version to `.dep/composer.phar`. To use specific composer version
 // download desired phar and place it at `.dep/composer.phar`.
 set('bin/composer', function () {
-    if (test('[ -f {{deploy_path}}/.dep/composer.phar ]')) {
+    $composer_phar_exists = test('[ -f {{deploy_path}}/.dep/composer.phar ]');
```
richardhj, line 18: Is there a later usage of the variable or why did you introduce it?

recipe/deploy/vendors.php
```diff
@@ -1,15 +1,33 @@
...
+    if ($composer_phar_exists) {
+        // If composer.phar is older than `composer_self_update` days, run self update
+        if(
+            get('composer_self_update')
```
richardhj, line 22: Using `0` as second parameter would make this operator superfluous

recipe/deploy/vendors.php
```diff
@@ -1,15 +1,33 @@
...
+            && get('composer_self_update') > 0
+            && $composer_mtime = run('stat -c %Y {{deploy_path}}/.dep/composer.phar')
+        ) {
+            if(strtotime(sprintf('+%d days', (int) get('composer_self_update')), $composer_mtime) <= time()) {
```
richardhj, line 26: If-statements should be merged

recipe/deploy/vendors.php
```diff
@@ -28,5 +46,9 @@
     if (!commandExist('unzip')) {
         warning('To speed up composer installation setup "unzip" command with PHP zip extension.');
     }
-    run('cd {{release_or_current_path}} && {{bin/composer}} {{composer_action}} {{composer_options}} 2>&1');
+    $paths = get('vendor_paths');
+    foreach ($paths as $path) {
+        $commands[] = "cd $path && {{bin/composer}} {{composer_action}} {{composer_options}} 2>&1";
+    }
+    run(implode('; ', $commands));
```
richardhj, line 53: Just run `run()` inside the loop, you save a variable and iteration then :)

recipe/deploy/vendors.php
```diff
@@ -28,5 +46,9 @@
     if (!commandExist('unzip')) {
         warning('To speed up composer installation setup "unzip" command with PHP zip extension.');
     }
-    run('cd {{release_or_current_path}} && {{bin/composer}} {{composer_action}} {{composer_options}} 2>&1');
+    $paths = get('vendor_paths');
```
richardhj, line 49: No need to introduce this variable

recipe/deploy/vendors.php
```diff
@@ -1,15 +1,33 @@
 <?php
 namespace Deployer;
 
+set('vendor_paths', ['{{release_or_current_path}}']);
```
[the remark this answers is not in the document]
nlemoine: But does it hurt? Except adding a setting, that's a feature with no impact that will be completely transparent to all users not interested by this.

To be honest, this one is the main reason of this PR.

recipe/deploy/vendors.php
nlemoine: Do not remember, maybe from a first iteration

recipe/deploy/vendors.php
nlemoine: done!

recipe/deploy/vendors.php
nlemoine: done!

recipe/deploy/vendors.php
nlemoine: Taken from the `clear_paths` recipe:
https://github.com/deployphp/deployer/blob/cf569530f5f030177b324a2d276b6df62dfbba87/recipe/deploy/clear_paths.php#L20-L23

nlemoine:
Second thoughts about this:
- Should the self update feature be enabled for installed `composer` (e.g. system wide) via sudo?
- It would also be nice to have a feature/flag to force usage of one (composer.phar) over the other (system composer), or vice versa **without** having to put a `composer.phar` in `{{deploy_path}}/.dep/composer.phar`. I've had issues with some hosting providers renaming `composer` v2 to `composer2` (`composer` being v1) and thus failing to install dependencies.

recipe/deploy/vendors.php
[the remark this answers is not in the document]
antonmedv: It definitely make code less readable. Also it’s new config to already big list of configs. 

Tasks in Deployer easily overriden, for a reason.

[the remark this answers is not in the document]
nlemoine: > It definitely make code less readable

Sorry, I can't agree with this.

Any developer can read this, this is pretty self explanatory: 
```php
foreach (get('vendor_paths', []) as $path) {
    run("cd $path && {{bin/composer}} {{composer_action}} {{composer_options}} 2>&1");
}
```

> Also it’s new config to already big list of configs.

Although I'm not happy with it, I can understand much more this one. But the feature we're talking about does not force you to set the added option because it has a default behavior that will satisfy 99% of users.

> Tasks in Deployer easily overriden, for a reason.

And Deployer offers a really nice API for this, thanks! Even if I'm a bit frustated to have to override a whole task/recipe for a tiny missing option.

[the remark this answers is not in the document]
antonmedv: > Sorry, I can't agree with this.

Agree. I mean more stuff to ask question while reading code. Someone may encounter this loop and wondering WTF?

> Although I'm not happy with it, I can understand much more this one. But the feature we're talking about does not force you to set the added option because it has a default behavior that will satisfy 99% of users.

I'm looking on it from the other side: only small percentage of users (~1%) gonna benefit from it. What I'm thinking is not try to cover every possible usage scenario, but create "a framework" for user to implement all they want. 

For example you can simple override only update_code task:
```php
    foreach (get('vendor_paths', []) as $path) {
        run("cd $path && {{bin/composer}} {{composer_action}} {{composer_options}} 2>&1");
    }
});
```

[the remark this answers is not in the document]
nlemoine: Thanks for the snippet, that's pretty much what I'm doing already. I thought that a bit more flexibility over this wouldn't hurt but I understand your point of view.

I desperetaly tried to find an issue where someone asked for this and didn't found any 😅 
I'm mostly working with WordPress/Timber and that kind of things are often needed. If anyone stumble upon this comment, please raise your 🙌  to help me grow that 1% use case 😄

I'll remove that `vendor_paths` part soon from the PR.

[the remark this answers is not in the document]
antonmedv: As more requests will come - I'll rereview my view)

Please, create separate PR for other features. Will close this PR for now.

PR #3089, opened 2022-04-09
[httpie] Delete Content-Length

joanhey:
- [ ] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

      $ php bin/docgen

Curl add it automatically. ~And also change it if it isn't correct~.
If we add the `Content-Length` will replace the internal one created by Curl.

### DIFF
--- a/src/Utility/Httpie.php
+++ b/src/Utility/Httpie.php
@@ -91,7 +91,6 @@ class Httpie
         $http->body = $body;
         $http->headers = array_merge($http->headers, [
             'Content-Type' => 'application/json',
-            'Content-Length' => strlen($http->body),
         ]);
         return $http;
     }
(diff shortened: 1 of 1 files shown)


antonmedv:
But why?

joanhey:
Curl receive the body via 
`curl_setopt($ch, CURLOPT_POSTFIELDS, $this->body);`

And it add the `Content-Length`.
If we set an incorrect `Content-Length` will be a malformed request.

joanhey:
But the important thing is that Curl add it.
So less code in PHP.

antonmedv:
I see. didn't know about curl_setopt($ch, CURLOPT_POSTFIELDS, $this->body);

joanhey:
Another PR will change this line
`curl_setopt($ch, CURLOPT_POSTFIELDS, $this->body);`

to:
```
if($this->body) {
 curl_setopt($ch, CURLOPT_POSTFIELDS, $this->body);
}
```
Now when is a GET add an extra header that is not necessary.

joanhey:
Joined in #3098

PR #3090, opened 2022-04-09
[httpie] Delete unnecessary clone

joanhey:
- [ ] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

      $ php bin/docgen

### DIFF
--- a/src/Utility/Httpie.php
+++ b/src/Utility/Httpie.php
@@ -73,49 +73,44 @@ class Httpie
 
     public function query(array $params): Httpie
     {
-        $http = clone $this;
-        $http->url .= '?' . http_build_query($params);
-        return $http;
+        $this->url .= '?' . http_build_query($params);
+        return $this;
     }
 
     public function header(string $header, string $value): Httpie
     {
-        $http = clone $this;
-        $http->headers[$header] = $value;
-        return $http;
+        $this->headers[$header] = $value;
+        return $this;
     }
 
     public function body(string $body): Httpie
     {
-        $http = clone $this;
-        $http->body = $body;
-        $http->headers = array_merge($http->headers, [
+        $this->body = $body;
+        $this->headers = array_merge($http->headers, [
             'Content-Type' => 'application/json',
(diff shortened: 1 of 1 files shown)


antonmedv:
This is a big breaking change. Why?

joanhey:
This change don't break anything. All works the same.

It is NOT necessary to clone this object.
We change directly the attributes in the received instance.

joanhey:
When I have time I'll add tests to this class.
And it will be easier to make changes.

antonmedv:
Clone is essential part. One can create a "boilerplate" and copy it via the method. Probably somebody relays on it already.

joanhey:
But this is the contrary.
Clone it when you need it. And not httpie.

`return Httpie::get($url)->header(....)->send()
`
Here we don't needed.

`$response1 = Httpie::get($url)->header(....);`
Here is the same if it's cloned or not;

`$response2 = $response1->jsonBody(...);`
Here response2 is a copy of $reponse1, but like you said we changed $reponse1 too.

And the logical is that the developer use the clone if they want it. Not automatically Httpie.
```
$response2 = clone $response1;
$response2->jsonBody(....);
```

joanhey:
But you are who will decide every PR we send you.

joanhey:
Also it's possible to add a clone() method. And that will be semantically correct.

joanhey:
A good example is the `fetch()` function
It's working without clone, but we can simplify the code, and use less instances.

Now:
```php
$http = $http->nothrow($nothrow);
foreach ($headers as $key => $value) {
        $http = $http->header($key, $value);
}
if ($body !== null) {
        $http = $http->body($body);
}
return $http->send($info);

```

After:
```php
$http->nothrow($nothrow);
foreach ($headers as $key => $value) {
        $http->header($key, $value);
}
if ($body !== null) {
        $http->body($body);
}
return $http->send($info);
```

antonmedv:
Merged!

PR #3091, opened 2022-04-09
[httpie] Add encoding

joanhey:
If an empty string, '', is set, a header containing all supported encoding types is sent

https://curl.se/libcurl/c/CURLOPT_ACCEPT_ENCODING.html

- [ ] Bug fix #…?
- [x] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

      $ php bin/docgen


joanhey:
Joined in #3098

PR #3094, opened 2022-04-12
Allowance for symfony 6

lsv:
- [ ] Bug fix #…?
- [x] New feature?
- [ ] BC breaks?
- [x] Tests added?
- [ ] Docs added?

- Allowed for using symfony 6 components
- Github action now tests for the lowest (--prefer-lowest) and stable (--prefer-stable) composer requirements.


antonmedv:
Please use rc8

PR #3095, opened 2022-04-12
[httpie] Make __contruct()  private

joanhey:
- [ ] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

      $ php bin/docgen

### DIFF
--- a/src/Utility/Httpie.php
+++ b/src/Utility/Httpie.php
@@ -37,7 +37,7 @@ class Httpie
      */
     private $nothrow = false;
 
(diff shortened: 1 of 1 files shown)


joanhey:
All the attributes are private.

So this is impossible:

```php
$http = new Httpie();
$http->method = 'POST';
$http->url = '.....';
```

Need to be always instanced by a static method.

joanhey:
Joined in #3098

PR #3098, opened 2022-04-14
Httpie refactor

joanhey:
- [x] Bug fix #…?
- [x] New feature?
- [x] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

      $ php bin/docgen

# Refactor Httpie Class

#### public function body(string $body, string $type = 'application/json'): Httpie
Now we can pass the `Content-Type`. By default `application/json`, like before.
So it's possible to send: xml, csv, html, ...

#### public function getJson(bool $associative = true)
Now automatically add header `Acept: application/json`.
And We can choose the return type. By default associative array, like before.

#### Catch errors in `json_encode()` and `json_decode()`
New way to catch errors.
Before return an `HttpieException` in `json_decode()` errors. Now also with `json_encode()`.

#### Duplicate query()
https://github.com/deployphp/deployer/pull/3098#issuecomment-1099370073
#### Changed `__construct()` to private
It's a very small break, that will affect no code.
This also permit to change later the `__construct()` without BC.

All the attributes are private.
So this is impossible:

```php
$http = new Httpie();
$http->method = 'POST';
$http->url = '.....';
```

Need to be always instanced by a static method. (Factory Method)

Happy coding !!!
This info is in progress...

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -3043,35 +3043,30 @@
         },
         {
             "name": "phpstan/phpdoc-parser",
-            "version": "1.2.0",
+            "version": "1.4.2",
             "source": {
                 "type": "git",
                 "url": "https://github.com/phpstan/phpdoc-parser.git",
-                "reference": "dbc093d7af60eff5cd575d2ed761b15ed40bd08e"
+                "reference": "4cb3021a4e10ffe3d5f94a4c34cf4b3f6de2fa3d"
             },
             "dist": {
                 "type": "zip",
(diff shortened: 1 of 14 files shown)


joanhey:
The others PRs will need that you approve it or not.

joanhey:
Fix `query()` before always add to the `url`.
Example of fails:

```php
$http = Htppie::get(....)->query(['hola']);

$http = $http->query[('bye']);

```
The URL will repeat the query '...?hola?bye'.
Is the same if it's cloned or not.

joanhey:
@antonmedv 
I'm sending the changes in small commits, so you can review it and comment in any change.
It's OK for now ?

PR #3112, opened 2022-04-26
deploy:release : Only delete unfinished symlink, not whole release

### LINKED ISSUE #3111: Deployer deletes the current release If the 'release' and 'current' point to the same directory
- Deployer version: 6.8.0
- Deployment OS: Ubuntu 20.4

We recently had a production deployment which failed because of a database timeout. We were able to just finish the last steps manually and symlink the new release.

Unbeknownst to us, we did not realize that by not also removing the `release` Symlink after the Deployment was finished, Deployer started removing the contents of the directory the link points to (and thus the current version) rendering the current production system unusable.

While this can probably considered a user error, I would personally also consider it a bug that Deployer does not check whether or not `release` and `current` point to the same directory.

I apologize if this is already addressed in new Deployer versions, I skimmed through the changelogs since 6.8.0 but didn't see it.

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/deployphp) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/deployphp/deployer/issues/3111">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/deployphp/deployer/issues/3111/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

theCapypara:
This fixes #3111 by just not removing the release directory. This is supposedly save to do, since Deployer keeps track of good releases. 

- [x] Bug fix #3111

### DIFF
--- a/docs/recipe/deploy/release.md
+++ b/docs/recipe/deploy/release.md
@@ -63,7 +63,7 @@ Clean up unfinished releases and prepare next release
 
 
 ### releases
-[Source](https://github.com/deployphp/deployer/blob/master/recipe/deploy/release.php#L149)
+[Source](https://github.com/deployphp/deployer/blob/master/recipe/deploy/release.php#L148)
 
 Shows releases list.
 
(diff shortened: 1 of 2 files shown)


antonmedv:
Merged.

PR #3147, opened 2022-05-30
Fix website provisioning with wrong remote user

### LINKED ISSUE #3146: Provision:website - Wrong remote user
- Deployer version:  7.0.0-rc.8
- Deployment OS: Mac OS 12.4 - M1

Default `deploy.php` with `remote_user` = `ubuntu`

Error:
```
task provision:website
[production]  error  ....
[production] run [ -d /home/deployer/app ] || mkdir /home/deployer/app
[production] Please login as the user "ubuntu" rather than the user "root".
```

The error occurs because the remote user is set statically.
I think this was done because of the command `run("echo 'import $deployPath/Caddyfile' >> /etc/caddy/Caddyfile");`.
The mentioned command will fail if the user is not root.
See my pull request, which fixes this.

fschirinzi:
- [X] Bug fix #3146?

### DIFF
--- a/recipe/provision/website.php
+++ b/recipe/provision/website.php
@@ -11,8 +11,6 @@ set('public_path', function () {
 
 desc('Provision website');
 task('provision:website', function () {
-    set('remote_user', 'deployer');
-
     run("[ -d {{deploy_path}} ] || mkdir {{deploy_path}}");
 
     $domain = get('domain');
@@ -68,9 +66,8 @@ EOF;
         run("echo $'$caddyfile' > Caddyfile");
     }
 
-    set('remote_user', 'root');
     if (!test("grep -q 'import $deployPath/Caddyfile' /etc/caddy/Caddyfile")) {
-        run("echo 'import $deployPath/Caddyfile' >> /etc/caddy/Caddyfile");
+        run("echo 'import $deployPath/Caddyfile' | sudo tee -a /etc/caddy/Caddyfile");
     }
     run('service caddy reload');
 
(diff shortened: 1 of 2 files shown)


recipe/provision/website.php
```diff
@@ -68,9 +66,8 @@
         run("echo $'$caddyfile' > Caddyfile");
     }
 
-    set('remote_user', 'root');
     if (!test("grep -q 'import $deployPath/Caddyfile' /etc/caddy/Caddyfile")) {
-        run("echo 'import $deployPath/Caddyfile' >> /etc/caddy/Caddyfile");
+        run("echo 'import $deployPath/Caddyfile' | sudo tee -a /etc/caddy/Caddyfile");
```
antonmedv, line 70: We can't use this as this will require a password on some hosts.

antonmedv (changes requested):

recipe/provision/website.php
fschirinzi: I chose to do it like that because I found the same concept here: https://github.com/deployphp/deployer/blob/68a983a8ca52e6ffd44ad08f139d3dc94fb570da/recipe/provision.php#L94

It would be consistent with the other code.

fschirinzi: I removed "sudo" since it works without it when using `-o become=root`.

orbziiharry:
This is still valid. The hard coded value or remote_user breaks this tool on EC2 instances.

dvdinbox:
This  is preventing me from deploying to aws ec2. 

Where can I change code to hack my way through this issue? 

I'm trying to make changes in my local ./vendor/deployer directory, but it seems as though deployer executes code on the remote server. Unfortunately I can't find where I can update code and have the updated code execute.

PR #3203, opened 2022-07-19
Fix default value of `shell` in docs

sforward:
This PR fixes the docs showing an incorrect default value of the `shell` parameter. The code uses `bash -s`, see [ProcessRunner](https://github.com/deployphp/deployer/blob/master/src/Component/ProcessRunner/ProcessRunner.php#L48).

### DIFF
--- a/docs/hosts.md
+++ b/docs/hosts.md
@@ -109,7 +109,7 @@ Default is `true`.
 
 ### `shell` 
 
-Default is `bash -ls`.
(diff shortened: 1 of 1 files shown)


antonmedv:
It's more complicated https://github.com/deployphp/deployer/blob/85b771fb9fb778e6fa774656da6ce21748ce74ac/src/Deployer.php#L125

PR #3212, opened 2022-07-23
Pimcore recipe fixes

gizamichal:
Pimcore database configuration is located in database.yaml file.

### DIFF
--- a/docs/recipe/pimcore.md
+++ b/docs/recipe/pimcore.md
@@ -13,7 +13,7 @@
 ## Tasks
 
 ### pimcore:rebuild-classes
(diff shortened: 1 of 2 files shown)


gizamichal:
This second commit fixes the following problem:

IOException
HTTP 500 Internal Server Error
Cannot rename "/tmp/url_matching_routes.phpJJWiht" to "/path/to/www/releases/X/var/cache/dev/url_matching_routes.php": rename(/tmp/url_matching_routes.phpJJWiht,/path/to/www/releases/X/var/cache/dev/url_matching_routes.php): Permission denied

I also added cache_clear task. Some developers may push content of this directory to repository, despite it isn't necessary and good practice.

Issue #3193, opened 2022-07-13
remove/replace default `--no-suggest` composer option on 6.8 branch


arderyp:
- Deployer version: 6.8.0
- Deployment OS: Linux Ubuntu 20.04 / RHEL 8.6

https://github.com/deployphp/deployer/blob/6.x/recipe/common.php#L90

This is causing the following composer error:

```
[MY-ENV] > cd /path/to/my/deployment/releases/34 && /usr/local/bin/composer install --verbose --prefer-dist --no-progress --no-interaction --no-dev --optimize-autoloader --no-suggest
[app-prod.ce.berkeley.edu] < [error] You are using the deprecated option "--no-suggest". It has no effect and will break in Composer 3.
```

I realize you are working on a major 7 release, but I think this warrants a quick and easy patch to the 6.x branch with a new release.

Thanks for the awesome project!

antonmedv:
Sorry but v6 is no longer maintained. You can override this config in your own recipe.

PR #3215, opened 2022-07-24
[6.x] Remove deprecated option --no-suggest

lnfel:
- [x] Bug fix

This PR aims to fix an issue #3193  where deprecated `--no-suggest` option in composer is being called.


antonmedv:
Sorry but v6 is no longer maintained. You can override this config in your own recipe.

Issue #3198, opened 2022-07-15
Implement default_selector


antonmedv:
@antonmedv Not sure if this would work but around here in the `SelectCommand`
https://github.com/deployphp/deployer/blob/42734c5f352e04279097fde6f4cc9182e342755b/src/Command/SelectCommand.php#L56
Could we have a hook to provide a default selector? I'd sure enough set it to `stage=staging`

_Originally posted by @joelpittet in https://github.com/deployphp/deployer/issues/2676#issuecomment-1050429787_

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/deployphp) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/deployphp/deployer/issues/3198">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/deployphp/deployer/issues/3198/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/deployphp/deployer/issues/3198/pledge.svg">
</picture>
</a>

juouy:
Honestly i cannot understand why there's still no "default_stage" from 6.x replacement as of 7.x.
Why should it be implemented? 

By default deployer connects to all defined hosts and attempts to begin deployment to all of them. 
Imagine you have dev and prod and it'll begin deployment everywhere just because you did not specify your label. In certain situations this can lead to dangerous mistakes and default_stage was the security switch, where you could define dev environment as default one.

PR #3247, opened 2022-08-03
Add default_selector and improve documentation on host labels and selection

marvinhinz:
- [ ] Bug fix #…?
- [x] New feature #3198
- [ ] BC breaks?
- [ ] Tests added?
- [x] Docs added?

The default_selector was pretty quick to implement. I also added some documentation. 

Would like to get some feedback on this!

congrats on the v7 release 🔥


joelpittet:
This looks pretty solid solution @marvinhinz 👏

joelpittet:
The documentation updates are a great touch too. Maybe could be shorter but they do describe the flexibility

marvinhinz:
Thanks! In the meantime i thought about another comfort feature. In addition to a default_selector, we could use selector maps or aliases. I'm a fan easy to remember commands, so for a big project setup with many servers a deployment to production can be a bit complicated as you might need to enter something like `dep deploy stage=prod&role=web,role=special` everytime.

```
add('selector_map', [
    "prod" => "stage=prod&role=web,role=special",
    "staging" => "stage=staging",
]);
```
could then result in:

`dep deploy prod`

@antonmedv whats you opinion on this?

marvinhinz:
@antonmedv seems like in the meantime you added documentation about the selectors itself, would you consider the default_selector functionality to be merged when i update this pr to the new documentation?

joelpittet:
@antonmedv poking this as I still think it's a good idea but interested in your opinion/take as maintainer.

antonmedv (approved):

joelpittet:
Whoa thanks @antonmedv for the quick turn around on the poke.

PR #3258, opened 2022-08-08
ci/release: Do not omit the autoload section from composer.json

tdgroot:
Ever since https://github.com/deployphp/deployer/pull/3051, many keys in the `composer.json` file are omitted when releasing a new Deployer version in order to have no version constraints for the package. This is all fine when just using deployer as-is, but becomes problematic when building on top of Deployer and depend on the namespaces and classes to be available.

With this change, external applications can make use of the internal Deployer classes, while the requirements are not being propagated through Composer.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -54,7 +54,7 @@ jobs:
         shell: bash
         run: |
           set -x
-          cat composer.json | jq 'del(.autoload) | del(.scripts) | del(.require) | del(."require-dev") | setpath(["bin"]; "dep")' > composer-new.json
+          cat composer.json | jq 'del(.scripts) | del(.require) | del(."require-dev") | setpath(["bin"]; "dep")' > composer-new.json
           mv composer-new.json composer.json
           git add composer.json
 


antonmedv:
I’m thinking about creating a separate tag on release which will be holding all needed dependencies and autoload.

tdgroot:
Yes that would be even better! Please let me know if there's anything I can help you with :).

antonmedv:
Let’s research the idea. 

 - what tag name? -dev?
 - Is composer allow to use something like composer-dev.json

dcanaud:
Hello, yes, composer is allowed to use a custom composer.json, I actually created a pull request to add this functionality to the vendors recipe: https://github.com/deployphp/deployer/pull/3259

staabm:
> Let’s research the idea.
> 
>     * what tag name? -dev?
> 
>     * Is composer allow to use something like composer-dev.json

rector and phpstan have a similar problem and I don't know yet what makes it work correctly for these tools



> Hello, yes, composer is allowed to use a custom composer.json, I actually created a pull request to add this functionality to the vendors recipe

thats a different thing and is unrelated to the problem mentioned initially

antonmedv:
One possible solution is to bundle all Symfony deps into Deployer namespce.

antonmedv:
Let’s just release with a special tag.

PR #3259, opened 2022-08-14
Allow setting of composer variables

dcanaud:
- [ ] Bug fix #…?
- [x] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [x] Docs added?

We deploy different composer.json file based on the environment we want to deploy on, this change will allow the user to set the composer.json based on the host. The default is '' which will just use the default composer.json. Sample usage:

```
host('test')
    ->set('composer_env_vars', 'COMPOSER=composer-test.json')
    ->set('deploy_path', '/var/www/html);

host('dev')
    ->set('composer_env_vars', 'COMPOSER=composer-dev.json')
    ->set('deploy_path', '/var/www/html);
```

### DIFF
--- a/docs/recipe/deploy/vendors.md
+++ b/docs/recipe/deploy/vendors.md
@@ -28,12 +28,25 @@
 ```
 
 
-### bin/composer
-[Source](https://github.com/deployphp/deployer/blob/master/recipe/deploy/vendors.php#L11)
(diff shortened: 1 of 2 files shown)


dcanaud:
Reference: https://getcomposer.org/doc/03-cli.md#composer

PR #3302, opened 2022-09-14
Added the laravel default folder storage/framework/cache to writable folders array

juliangarcess:
- [ ] Bug fix #…?
- [X] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [X] Docs added?

      $ php bin/docgen

### DIFF
--- a/docs/recipe/laravel.md
+++ b/docs/recipe/laravel.md
@@ -92,6 +92,7 @@ Overrides [writable_dirs](/docs/recipe/deploy/writable.md#writable_dirs) from `r
     'storage/app/public',
     'storage/framework',
     'storage/framework/cache',
+    'storage/framework/cache/data',
     'storage/framework/sessions',
     'storage/framework/views',
     'storage/logs',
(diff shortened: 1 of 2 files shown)


peterjaap:
In what situation would omitting this cause an issue? I've never seen this being a problem and I don't think adding this is necessary so I'd love to hear why you think this is needed in the general recipe instead of just for your installation?

juliangarcess:
As storage/framework/cache/data is a default folder of fresh laravel instalation is not affected by umask cause it comes directly from git repository, I think that this folder must be specifically mentioned in writable_dirs array.

In my personal experience I have problems with the recipe until I added that folder to writable array.

peterjaap:
Curious to hear from more people having problems with this. We've been using this recipe for years and haven't run into this issue yet.

juliangarcess:
what`s the problem with this PR? why it is not merged??

Schrank:
Because @peterjaap doesn't have this issue and the question is, is this a generic problem or one of your deployment

peterjaap:
I think the problem lies here, from your comment;

> not affected by umask cause it comes directly from git repository

This folder shouldn't be in any git repository, since it's a cache folder. So I'd advise you to remove it from your git by doing `git remote --cached storage/framework/cache`. That way, the webserver will create the folder and you won't run into problems.

juliangarcess:
If you install a fresh laravel aplication that folder came by default, then it must be mentioned on the recipe

Issue #3311, opened 2022-09-21
[IMPROVEMENT] allow to split Magento2 static file deployment in frontend and adminhtml area


mautz-et-tong:
We have a case where we have 30 frontend languages but only 2 adminhtml languages.

That causes a big overhead when deploying static files.

My 5 minute idea hack for this would be the following improvement:

in the magento2.php replace ll 97.

```
desc('Deploys assets');
task('magento:deploy:assets', function () {

    $themesToCompile = '';
    if (count(get('magento_themes')) > 0) {
        foreach (get('magento_themes') as $theme) {
            $themesToCompile .= ' -t ' . $theme;
        }
    }

    run("{{bin/php}} {{release_or_current_path}}/bin/magento setup:static-content:deploy --content-version={{content_version}} {{static_content_locales}} $themesToCompile -j {{static_content_jobs}}");
});
```

with:
```
// Deploy frontend and adminhtml together as default
set('split_static_deployment', false);

desc('Deploys assets');
task('magento:deploy:assets', function () {

    $themesToCompile = '';
    if (get('split_static_deployment')) {
        invoke('magento:deploy:assets:adminhtml');
        invoke('magento:deploy:assets:frontend');
    } elseif (count(get('magento_themes')) > 0 ) {
        foreach (get('magento_themes') as $theme) {
            $themesToCompile .= ' -t ' . $theme;
        }
        run("{{bin/php}} {{release_or_current_path}}/bin/magento setup:static-content:deploy --content-version={{content_version}} {{static_content_locales}} $themesToCompile -j {{static_content_jobs}}");
    }
});

desc('Deploys assets for backend only');
task('magento:deploy:assets:adminhtml', function () {

    $themesToCompile = '';
    if (count(get('magento_themes_backend')) > 0 && get('split_static_deployment')) {
        foreach (get('magento_themes_backend') as $theme) {
            $themesToCompile .= ' -t ' . $theme;
        }
    }

    run("{{bin/php}} {{release_or_current_path}}/bin/magento setup:static-content:deploy --area=adminhtml --content-version={{content_version}} {{static_content_locales_backend}} $themesToCompile -j {{static_content_jobs}}");
});

desc('Deploys assets for frontend only');
task('magento:deploy:assets:frontend', function () {

    $themesToCompile = '';

    if (count(get('magento_themes_frontend')) > 0 && get('split_static_deployment')) {
        foreach (get('magento_themes_frontend') as $theme) {
[...]

PR #3324, opened 2022-09-30
Adds possibility to execute shopware commands as another user

UlrichThomasGabor:
- [ ] Bug fix #…?
- [X] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

      $ php bin/docgen

### DIFF
--- a/recipe/shopware.php
+++ b/recipe/shopware.php
@@ -19,8 +19,9 @@ require_once __DIR__ . '/common.php';
 
 add('recipes', ['shopware']);
 
-
 set('default_timeout', 3600); // Increase when tasks take longer than that.
+set('use_sudo_for_shopware_commands', false);
+set('use_sudo_user_for_shopware_commands', 'root');
 
 // These files are shared among all releases.
 set('shared_files', [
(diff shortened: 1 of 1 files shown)


peterjaap:
@UlrichThomasGabor this feels like a workaround for a deeper lying problem. What's your usecase exactly?

UlrichThomasGabor:
In our setup the deployment user is different from the user which runs the webserver-php. Executing tasks as deployment user generates files which are thus not owned by webserver-php, and vice-versa. Since the last update Shopware is performing some `chmod` operations. These calls fail if files are not owned by the user currently executing the command, which is either the deployment user or the webserver-PHP-user depending where its called.

I did not find a way to configure Shopware such that it does not execute the `chmod`. It seems this is not possible without changes. So it's either `chown`ing basically all files back and forth to the correct user via `writable_dirs`/`writable_files`, which is not nice since [it's not possible to `chmod` additionally](https://github.com/deployphp/deployer/blob/6d759ef372af18c2a8ec48d4bf5282aee0133b64/recipe/deploy/writable.php#L80-L84) (and actually I do not want every file to be owned by the webserver-php-user) and I cannot anticipate which files Shopware will generate in the future therefore any solution in this direction will fail sooner or later again. Or just executing the commands as the webserver-php-user such that all files Shopware wants to create are only owned be one user.

peterjaap:
@UlrichThomasGabor thanks for your explanation. I do however feel this is very specific for your situation. We like to keep the recipes applicable to as many people as people, which as few code as possible. So I'm leaning towards not merging this, since you could easily add it yourself by overriding the tasks in your local `deploy.php` or recipe.

peterjaap:
It's more of a philosophical question than it is about the actual code. When do we introduce code to support a smaller number of use-cases?

I have the same conundrum here; https://github.com/deployphp/deployer/issues/3311#issuecomment-1254687742

@antonmedv do you have any thoughts?

UlrichThomasGabor:
[Cleanup also provides a sudo option](https://github.com/deployphp/deployer/blob/a52ae3da6a4bcf10f7bf42f7fac7d982a62240c2/recipe/deploy/clear_paths.php#L8) which can only be necessary if the runner of the application is running under a different user, otherwise no situation can arise where the deployment-user loses access to his own files?

I also don't think that our situation is that special? Webservers are either running with a single user for everything because basically nobody cares in a containerized environment, then Deployers chown/chgrp functionality is not needed anyway because everything is owned and executed by the same user. Or the deployment user is not the one used for executing scripts, because nobody should allow an `www-data` user to login, then you need the chown/chgrp writable functionality of Deployer but then you also need a way to make sure that you can execute commands as a user other than the one used for deployment.

So, actually in my opinion this (have a "runner user" to execute application commands) is a functionality which should actually be present in Deployer itself and not in the recipe, but well, ... I don't know and this is at least larger at the problem I have right now.

peterjaap:
@UlrichThomasGabor ok good point about the cleanup task. I'd like to have @antonmedv chime in on this one.

antonmedv:
But we already have become. Why can’t we use it?

UlrichThomasGabor:
I think it would solve the problem too, but it's not the same. When one uses `become` (which is basically nowhere documented by the way?) everything is owned/executed by that user. In the solution I created only the necessary files are owned by that user.

Or, to be more specific, using `become` every file deployed will be owned and is thus changeable by `www-data` which is something one might not want (e.g. for security reasons); whereas in my solution the `www-data` user only owns and thus can write to the directories/files specified via `writable_dirs`/`writable_files` and the temporary files it creates when executing a command. It will not be able to alter is own scripts.

But I guess a more clean solution than the one I proposed would be to just split these two options:
https://github.com/deployphp/deployer/blob/85b771fb9fb778e6fa774656da6ce21748ce74ac/src/Component/Ssh/Client.php#L61
and
https://github.com/deployphp/deployer/blob/85b771fb9fb778e6fa774656da6ce21748ce74ac/src/Utility/Rsync.php#L68
?

Then I can leave out a `become`-user for the rsync, which is the current situation in our deployment, and use `become` for `run` to execute all other commands, which is the functionality I implemented directly in the Shopware recipe here.

antonmedv:
What cd about adding rsync_become option with fallback to become?

UlrichThomasGabor:
I've done it with `become` for now because the problem had to be solved today. If I encounter problems with that later, I'll open another ticket.

antonmedv:
Sounds good

PR #3325, opened 2022-09-30
Removed temporary workaround as it is for a version which is a year old

UlrichThomasGabor:
- [ ] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

      $ php bin/docgen

This was a temporary workaround and the version with the fix was released nearly a year ago. There were multiple security updates in between, therefore running the old version is negligent anyway. In my opinion this can be deleted.

### DIFF
--- a/recipe/shopware.php
+++ b/recipe/shopware.php
@@ -156,19 +156,6 @@ task('sw-build-without-db:get-remote-config', static function () {
 
         run('./bin/console theme:dump');
         download('{{deploy_path}}/current/files/theme-config', './files/');
-
-        // Temporary workaround to remove absolute file paths in Shopware <6.4.6.0
-        // See:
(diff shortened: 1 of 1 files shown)


peterjaap:
@UlrichThomasGabor do you have any documentation that says this version isn't supported anymore / EOL? I can't find it anywhere.

UlrichThomasGabor:
I don't think there is any. I have never seen that they released a bugfix/minor release for an older minor/major release after they have released a following minor/major one. I guess the maintenance plan is basically "stay up-to-date with each new release whatever major/minor/bugfix one it is or fall behind".

There is a security plugin for older versions though: https://store.shopware.com/swag136939272659f/shopware-6-sicherheits-plugin.html
I don't know how good the compatibility (and the provided security fixes) are though.

peterjaap:
Ok good enough for me!

PR #3343, opened 2022-10-11
add pub/static/_cache as shared_dir

torhoehn:
- [x] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [x] Docs added?

      $ php bin/docgen

`pub/static/_cache` should be declared as `shared_dir`, so it's possible to mount it between different app-servers
 in a multi host setup.

Example:
- run Magento with two app-servers (app-a / app-b)
- route all admin request to app-a
- cached css files will be generated on app-a and app-b
- observe 404 responses for files of `pub/static/_cache` directory


torhoehn:
@peterjaap Wanted to test behaviour before the change again, but when you already agree I'll mark it as ready. 😀

PR #3369, opened 2022-11-02
Magento: Do not disable maintenance mode after failed deployment

schmengler:
A failed deployment is likely not in a usable state. Also if it failed because of missing bin/magento, this command will also fail and the the process stops before deploy:unlock can be executed

- [x] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

      $ php bin/docgen

### DIFF
--- a/recipe/magento2.php
+++ b/recipe/magento2.php
@@ -218,4 +218,3 @@ task('deploy', [
     'deploy:publish',
 ]);
 
-after('deploy:failed', 'magento:maintenance:disable');


antonmedv:
As symlink is switched only at the last step, it's safe to disable mm.

schmengler:
Maintenance mode is achieved through a flag in `var` which is not shared, so that it is tied to the release. When the shop is in mainenance mode, the symlink has already been switched and database migrations are running. Regardless of the symlink, a failure during database migrations is likely to leave an unstable state

also this, which was my actual reason for the PR:
>  if it failed because of missing bin/magento, this command will also fail and the the process stops before deploy:unlock can be executed

antonmedv:
Make sense. Let’s merge.

Schrank:
@sprankhub @peterjaap Opinions?

peterjaap:
Oof this is hard one. It depends on _why_ the deployment failed, so you can't really confidently pick a good solution to it.

For example, a bad state would be where `bin/magento setup:upgrade` failed. In that case, removing the maintenance flag would probably (but not always) result in a 500 error.

In other cases, removing the maintenance flag would just show the 'current' deployment (not the new one).

sprankhub:
Exactly what @peterjaap says. I tend to keep the current process, because:

1. if it falls back to the last version, we should disable maintenance mode and
2. if the symlink has already been switched, there is no huge advantage in showing the maintenance page over showing a 500.

Your monitoring should inform you about 2 anyway ^^

antonmedv:
If I got it correctly we are closing this pr.

PR #3384, opened 2022-11-21
Cache key id prefix

valguss:
Reference Issue: https://github.com/davidalger/capistrano-magento2/issues/151

- [x] Bug fix https://github.com/davidalger/capistrano-magento2/issues/151
- [x] Docs added?


valguss:
Going to take this back and work some more on it actually

Issue #3386, opened 2022-11-21
Sulu recipe - Not all necessary shared folders are writable


AndreasA:
- Deployer version: 7.0.2
- Deployment OS: Ubuntu.

Using the "sulu.php" recipe not all necessary shared folders are writable. In combination with writable recursive being false - new default with 7.x - those folders might miss the necessary permissions.

Issue #3379, opened 2022-11-09
Change of default value of `writable_recursive` from `true` to `false` missing in upgrade docs


msarris:
In deployer 7 the `writable_recursive` was changed to be default `false`, while in 6 (and possibly before) is was default `true`. I noticed that this is not mentioned in the upgrade docs (https://deployer.org/docs/7.x/UPGRADE), and is probably important for a lot of deploys (our deployment was broken because of this for example).

So probably a good idea to add it to to the upgrade docs. Thanks!

(Here you can see it was true: https://github.com/deployphp/deployer/blob/6.x/recipe/common.php, and here it's now false: https://github.com/deployphp/deployer/blob/master/recipe/deploy/writable.php.)

PR #3387, opened 2022-11-21
Change writable dirs of sulu recipe to include all shared dirs.

AndreasA:
- [X] Bug fix #...? 
#3386
- [X] New feature? No
- [X] BC breaks? No
- [X] Tests added? Not needed.
- [X] Docs added? Not needed.

### DIFF
--- a/recipe/sulu.php
+++ b/recipe/sulu.php
@@ -7,7 +7,7 @@ add('recipes', ['sulu']);
 
 add('shared_dirs', ['var/indexes', 'var/sitemaps', 'var/uploads', 'public/uploads']);
 
-add('writable_dirs', ['public/uploads']);
+add('writable_dirs', ['var/indexes', 'var/sitemaps', 'var/uploads', 'public/uploads']);
 
 set('bin/websiteconsole', function () {
     return parse('{{bin/php}} {{release_or_current_path}}/bin/websiteconsole --no-interaction');


alexander-schranz (approved):
Sulu Core Team Member here. This change looks required as the update from Deployer 6 to 7 the Symfony `var` directory is not longer `writable_recursive` `true` like already reported in #3379. The other directories need also be set to be writable.

Alternative would be that the `symfony.php` recipe sets:

```php
set('writeable_dirs', ['var']);
set('writable_recursive', true);
```

Which would maybe something somebody would expect for a directory called `var`. But I think there was maybe a specific reasons not longer use `writable_recursive` by default @antonmedv?

antonmedv:
Can we add this to sulu recipe?

    add('writeable_dirs', ['var']);

AndreasA:
@antonmedv without `writable_recursive` set to `true` it will not make the other dirs in `var` writable. `var` is already part of the Symfony recipe, that is the base for this recipe.

alexander-schranz:
> Can we add this to sulu recipe?
> add('writeable_dirs', ['var']);

The `var` directory is already there as `sulu.php` uses the `symfony.php` recipe. The problem I think is mostly that somebody is copying things from another servers into the symfony applications `var` directory. And in deployer 6 with `writable_recursive` set to `true` you never run into any problems with this kind of sync processes as `writable_recursive` did follow symlinks to the shared dirs. And in deployer 7 with `writable_recursive` set to `false` we need to add all `shared_dirs` which are writeable again on `writeable_dirs`. Or as alternative add `writable_recursive` `true` again to `symfony.php` then all symfony depending recipes don't need again add all shared_dirs which are writeable again to writeable_dirs.

Hope that was not to confusing 🙈

antonmedv:
Let's use writable_recursive.

antonmedv:
Let's create a PR.

AndreasA:
@antonmedv done: https://github.com/deployphp/deployer/pull/3451

PR #3388, opened 2022-11-22
Magento2 cache key id prefix

valguss:
- [x] Bug fix https://github.com/davidalger/capistrano-magento2/issues/151
- [x] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [x] Docs added?

      $ php bin/docgen


antonmedv:
This code seems very odd to me. Let's find a nicer solution. 

Also we really need to add some e2e magento tests.

Issue #3399, opened 2022-12-08
deployer 7.0.2 does not run with php 8.2


chenrui333:
- Deployer version: 7.0.2
- Deployment OS: macOS

      Please, provide a minimal reproducible example of deploy.php

## How to reproduce

- Install php8.2
- Install deployer 7.0.2
- Run `dep init --no-interaction`

```
==> /opt/homebrew/Cellar/deployer/7.0.2/bin/dep init --no-interaction
Last 15 lines from /Users/rui/Library/Logs/Homebrew/deployer/test.01.dep:
2022-12-08 11:07:16 -0500

/opt/homebrew/Cellar/deployer/7.0.2/bin/dep
init
--no-interaction

 ErrorException  in DumpCompletionCommand.php on line 48:

  Using ${var} in strings is deprecated, use {$var} instead
```

relates to https://github.com/Homebrew/homebrew-core/pull/117647

neverlose-lv:
Same here

Schrank:
Have you turned on deprecation warnings? What happens if you turn them off?

Or in other words: Is the problem PHP or the error handler?

neverlose-lv:
> Have you turned on deprecation warnings? What happens if you turn them off?
> 
> Or in other words: Is the problem PHP or the error handler?

Don't mess up. It is not a deprecation warning. It is a deprecated syntax/approach in the new version of PHP 8.2., e.g. -> you are not allowed to do 'so' anymore in the new version.

sgrossberndt:
I am running deployer in a composer image

I had to add `echo 'error_reporting=8191' >> /usr/local/etc/php/php-cli.ini` in order to fix it:
```
    - composer global require deployer/deployer $COMPOSER_OPTIONS
    # Workaround for https://github.com/deployphp/deployer/issues/3399
    - echo 'error_reporting=8191' >> /usr/local/etc/php/php-cli.ini
    - /tmp/vendor/bin/dep deploy -vvv ${TARGET_HOSTS}
```

Really a pity such a thing makes deployments fail :-(
I think, I'll keep this setting even when the symfony/console issue is solved.

Schrank:
It should, because PHP 8.2 deprecated it only. It isn't removed.

https://www.php.net/releases/8.2/en.php
> # Deprecations and backward compatibility breaks
> - Deprecated ${} string interpolation.

Can you try with error_reporting=0 please? 🤔

sgrossberndt:
[...]

PR #3422, opened 2023-01-02
Make Deployer work with PHP 8.2

franmomu:
Replaces https://github.com/deployphp/deployer/pull/3401

Using @Konafets' work and updating the composer.lock file

Closes https://github.com/deployphp/deployer/issues/3399 and https://github.com/deployphp/deployer/pull/3401

- [x] Bug fix https://github.com/deployphp/deployer/issues/3399?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        php-versions: [ '7.3', '7.4', '8.0', '8.1' ]
+        php-versions: [ '7.3', '7.4', '8.0', '8.1', '8.2' ]
     steps:
       - uses: actions/checkout@v2
 
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,7 @@
         "justinrainbow/json-schema": "^5.2",
         "psr/http-message": "^1",
         "react/http": "^1.5",
-        "symfony/console": "^5",
+        "symfony/console": "^5.4.9",
         "symfony/polyfill-php80": "^1.22",
         "symfony/process": "^5",
         "symfony/yaml": "^5"
--- a/composer.lock
+++ b/composer.lock
@@ -2976,21 +2976,21 @@
         },
         {
             "name": "phpspec/prophecy",
-            "version": "v1.15.0",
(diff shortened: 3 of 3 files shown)


B4rb4ross4:
Isn't the composer.lock ignored when composer is required by another package? I always thought so. And I don't think I misread the documentation (https://getcomposer.org/doc/02-libraries.md#lock-file). So your change should in theory have no effect.

B4rb4ross4:
And the php 8.2 checks are missing
https://github.com/deployphp/deployer/blob/1f5cfb25c247506b8d2a12484c87f9b8b3fe3f9b/.github/workflows/test.yml#L14

franmomu:
> Isn't the composer.lock ignored when a library is required by a project? I always thought so. And I don't think I misread the documentation (https://getcomposer.org/doc/02-libraries.md#lock-file). So your change should in theory have no effect.

The thing here, if I'm not wrong, is that this change would have effect in the next release because of this job: https://github.com/deployphp/deployer/blob/1f5cfb25c247506b8d2a12484c87f9b8b3fe3f9b/.github/workflows/release.yml

When a release is created, [it executes `bin/build`](https://github.com/deployphp/deployer/blob/1f5cfb25c247506b8d2a12484c87f9b8b3fe3f9b/.github/workflows/release.yml#L18) which [uses `composer install`](https://github.com/deployphp/deployer/blob/1f5cfb25c247506b8d2a12484c87f9b8b3fe3f9b/bin/build#L28) and generates the `dep` binary, then [the `composer.json` file is modified to remove some sections included the `require` one](https://github.com/deployphp/deployer/blob/1f5cfb25c247506b8d2a12484c87f9b8b3fe3f9b/.github/workflows/release.yml#L38-L59).

If you see https://packagist.org/packages/deployer/deployer#v7.0.2 there are no requirements there.

maxhelias (approved):

DamienHarper (approved):

Konafets (approved):

UlrichThomasGabor:
@antonmedv Can you please release a new version with that PR merged?

antonmedv:
Will try to do it today.

Jamesking56 (approved):

PR #3435, opened 2023-01-11
Fix Node.js 12 actions are deprecated

benoitchantre:
This PR fixes the following warning

> Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout@v2, actions/cache@v2

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -10,7 +10,7 @@ jobs:
   phpstan:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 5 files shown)


benoitchantre:
I just pushed a commit to replace unmaintained satackey/push-prebuilt-action by a maintained fork.

I think I need to open a new PR.

PR #3438, opened 2023-01-12
Update composer to the latest version before using

trsteel88:
At the moment, if composer isn't available system wide, this task installs it into the .dep folder. However, on the next deploy it uses the previously installed composer.

This means, it's never updated. To ensure composer stays up to date this PR runs self-update to update it before using it.

Another option would be to always install composer.phar into the current release so it's always fresh.


antonmedv:
I think we don't need to download the latest as the default. 

First, deploy process may relay on a specific composer version.

Second, it easy to update the composer with just one command from user machine:

```sh
dep run '{{bin/composer}} self-update'
```

PR #3451, opened 2023-01-17
Use recursive writable for Symfony recipe.

AndreasA:
- [X] Bug fix #…? Yes. It fixes potentially missing permissions in the `var` folder for created directories by e.g. custom Symfony classes or Sulu etc.
- [ ] New feature? No
- [ ] BC breaks? If folders in the `var` Symfony directory have wrong permissions, it might require sudo to set writable for them correctly. It is most likely no breaking change but not 100% sure.
- [ ] Tests added?
- [ ] Docs added?

### DIFF
--- a/docs/recipe/symfony.md
+++ b/docs/recipe/symfony.md
@@ -88,39 +88,46 @@ Overrides [shared_files](/docs/recipe/deploy/shared.md#shared_files) from `recip
 ```
 
 
-### writable_dirs
+### writable_recursive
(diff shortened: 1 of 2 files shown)


Seb33300:
Symfony related documentation: https://symfony.com/doc/current/setup/file_permissions.html

AndreasA:
as there still has nothing happened here, I am closing the PR. let me know if I should reopen / recreate it

PR #3453, opened 2023-01-18
Magento 2: Cache key id prefix

valguss:
- [x] Bug fix #…?
- [x] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [x] Docs added?

      $ php bin/docgen

### DIFF
--- a/recipe/magento2.php
+++ b/recipe/magento2.php
@@ -12,6 +12,8 @@ use Deployer\Host\Host;
 const CONFIG_IMPORT_NEEDED_EXIT_CODE = 2;
 const DB_UPDATE_NEEDED_EXIT_CODE = 2;
 const MAINTENANCE_MODE_ACTIVE_OUTPUT_MSG = 'maintenance mode is active';
+const ENV_CONFIG_FILE_PATH = 'app/etc/env.php';
+const TMP_ENV_CONFIG_FILE_PATH = 'app/etc/env_tmp.php';
 
 add('recipes', ['magento2']);
 
@@ -433,6 +435,47 @@ task('deploy:additional-shared', function () {
     add('shared_files', get('additional_shared_files'));
     add('shared_dirs', get('additional_shared_dirs'));
 });
+
+/**
+ * Update cache ip_prefix on deploy so that you are compiling against a fresh cache
+ * Reference Issue: https://github.com/davidalger/capistrano-magento2/issues/151
+ **/
+desc('Update cache id_prefix');
+task('magento:set_cache_prefix', function () {
+    //get current env config
+    $envConfigString = run("cat {{deploy_path}}/shared/" . ENV_CONFIG_FILE_PATH);
+    $envConfig = eval('?>'.$envConfigString);
+    //set prefix to `alias_releasename_`
+    $prefixUpdate = get('alias') . '_' . get('release_name') . '_';
+
+    //update id_prefix to include release name
+    $envConfig["cache"]["frontend"]["default"]["id_prefix"] = $prefixUpdate;
+    $envConfig["cache"]["frontend"]["page_cache"]["id_prefix"] = $prefixUpdate;
+
+    //Generate configuration array as string
+    $envConfigStr = "<?php return " . var_export($envConfig, true) . ";";
+    // Touch tmp config
+    run("[ -f {{deploy_path}}/shared/" . TMP_ENV_CONFIG_FILE_PATH. " ] || touch {{deploy_path}}/shared/" . TMP_ENV_CONFIG_FILE_PATH);
+    run('echo $"' . $envConfigStr . '" > {{deploy_path}}/shared/' . TMP_ENV_CONFIG_FILE_PATH);
+    //delete the symlink for env.php
+    run("rm {{release_or_current_path}}/" . ENV_CONFIG_FILE_PATH);
+    //link the env to the tmp version
+    run("{{bin/symlink}} {{deploy_path}}/shared/" . TMP_ENV_CONFIG_FILE_PATH . " {{release_path}}/" . ENV_CONFIG_FILE_PATH);
+});
+after('deploy:shared', 'magento:set_cache_prefix');
+
+/**
+ * After successful deployment, move the tmp_env.php file to env.php ready for next deployment
+ */
+desc('Cleanup cache id_prefix env files');
+task('magento:cleanup_cache_prefix', function () {
+    run("rm {{deploy_path}}/shared/" . ENV_CONFIG_FILE_PATH);
+    run("rm {{release_or_current_path}}/" . ENV_CONFIG_FILE_PATH);
+    run("mv {{deploy_path}}/shared/" . TMP_ENV_CONFIG_FILE_PATH . " {{deploy_path}}/shared/" . ENV_CONFIG_FILE_PATH);
+    // Symlink shared dir to release dir
+    run("{{bin/symlink}} {{deploy_path}}/shared/" . ENV_CONFIG_FILE_PATH . " {{release_path}}/" . ENV_CONFIG_FILE_PATH);
+});
+after('deploy:magento', 'magento:cleanup_cache_prefix');
 
 
 desc('Prepares an artifact on the target server');
--- a/recipe/magento2.php
+++ b/recipe/magento2.php
@@ -437,30 +437,29 @@ task('deploy:additional-shared', function () {
 });
 
 /**
- * Update cache ip_prefix on deploy so that you are compiling against a fresh cache
+ * Update cache id_prefix on deploy so that you are compiling against a fresh cache
  * Reference Issue: https://github.com/davidalger/capistrano-magento2/issues/151
  **/
 desc('Update cache id_prefix');
 task('magento:set_cache_prefix', function () {
     //get current env config
-    $envConfigString = run("cat {{deploy_path}}/shared/" . ENV_CONFIG_FILE_PATH);
+    $envConfigString = run('cat {{deploy_path}}/shared/' . ENV_CONFIG_FILE_PATH);
     $envConfig = eval('?>'.$envConfigString);
     //set prefix to `alias_releasename_`
     $prefixUpdate = get('alias') . '_' . get('release_name') . '_';
 
     //update id_prefix to include release name
-    $envConfig["cache"]["frontend"]["default"]["id_prefix"] = $prefixUpdate;
-    $envConfig["cache"]["frontend"]["page_cache"]["id_prefix"] = $prefixUpdate;
-
+    $envConfig['cache']['frontend']['default']['id_prefix'] = $prefixUpdate;
+    $envConfig['cache']['frontend']['page_cache']['id_prefix'] = $prefixUpdate;
     //Generate configuration array as string
-    $envConfigStr = "<?php return " . var_export($envConfig, true) . ";";
+    $envConfigStr = '<?php return ' . var_export($envConfig, true) . ';';
     // Touch tmp config
-    run("[ -f {{deploy_path}}/shared/" . TMP_ENV_CONFIG_FILE_PATH. " ] || touch {{deploy_path}}/shared/" . TMP_ENV_CONFIG_FILE_PATH);
+    run('[ -f {{deploy_path}}/shared/' . TMP_ENV_CONFIG_FILE_PATH. ' ] || touch {{deploy_path}}/shared/' . TMP_ENV_CONFIG_FILE_PATH);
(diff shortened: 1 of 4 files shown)


valguss:
@antonmedv This is an updated version of this https://github.com/deployphp/deployer/pull/3388

It now does all the operations on the server and should be a bit easier to see what's going on

antonmedv:
Looks good to me. 👍🏻

antonmedv:
@peterjaap please review

antonmedv:
@peterjaap merging?

valguss:
Rebased docs again

peterjaap:
@valguss I'm still unsure what problem this exactly solves. I understand what it does, but I don't know which issues this fixes. So I'm hesitant to include new code for something that doesn't effect everyone.

valguss:
Ahh, The issue is based from here https://github.com/davidalger/capistrano-magento2/issues/151
Essentially, it's a way to run a deployed instance of magento on a totally fresh cache.
The main issue we see is that on high traffic sites where this is not implemented, the code generation part of magento `setup:di:compile` uses the currently deployed cache rather than a fresh cache which can cause it to incorrectly compile the site. When this happens, another redeploy is required to compile correctly

Hope that explains it more

Cheers

Tom

peterjaap:
@valguss so this is in the scenario where you run `setup:di:compile` on the target production machine instead of in a build pipeline?

valguss:
So it runs in the pipeline but because the cache is redis would be shared between the current version and the deploying version, the deploying version wouldn't necessarily generate all the classes required to run. We see this happening where we add new modules and the dependency injection is not created because the production cache dosen;t know about the new module.

peterjaap:
@valguss ok so it does run in a pipeline, but it runs on the production machine and not on an isolated build runner?

valguss:
That's correct. Most of the Magento commands run in the deployment folder on the deployment server

peterjaap:
In that case I'm not in favor of merging this into the default recipe since running those commands on the production server instead of a build pipeline isn't standard (or advised) practice. What do you think @schmengler / @Schrank ?

valguss:
Not sure I follow. All the magent2 commands in the recipe run on a target server before being pushed live? Maybe I've confused the situation

recipe/magento2.php
```diff
@@ -367,6 +369,47 @@
...
+desc('Update cache id_prefix');
+task('magento:set_cache_prefix', function () {
+    //get current env config
+    $envConfigString = run("cat {{deploy_path}}/shared/" . ENV_CONFIG_FILE_PATH);
+    $envConfig = eval('?>'.$envConfigString);
+    //set prefix to `alias_releasename_`
+    $prefixUpdate = get('alias') . '_' . get('release_name') . '_';
+
+    //update id_prefix to include release name
+    $envConfig["cache"]["frontend"]["default"]["id_prefix"] = $prefixUpdate;
```
peterjaap, line 385: Please use single quotes instead of double quotes.

recipe/magento2.php
```diff
@@ -367,6 +369,47 @@
...
+    $envConfig["cache"]["frontend"]["page_cache"]["id_prefix"] = $prefixUpdate;
+
+    //Generate configuration array as string
+    $envConfigStr = "<?php return " . var_export($envConfig, true) . ";";
```
peterjaap, line 389: Please use Symfony's VarDumper https://symfony.com/doc/current/components/var_dumper.html

recipe/magento2.php
```diff
@@ -367,6 +369,47 @@
...
+    // Touch tmp config
+    run("[ -f {{deploy_path}}/shared/" . TMP_ENV_CONFIG_FILE_PATH. " ] || touch {{deploy_path}}/shared/" . TMP_ENV_CONFIG_FILE_PATH);
```
peterjaap, line 391: Please use single quotes instead of double quotes.

recipe/magento2.php
```diff
@@ -367,6 +369,47 @@
...
+    run('echo $"' . $envConfigStr . '" > {{deploy_path}}/shared/' . TMP_ENV_CONFIG_FILE_PATH);
+    //delete the symlink for env.php
+    run("rm {{release_or_current_path}}/" . ENV_CONFIG_FILE_PATH);
```
peterjaap, line 394: Please use single quotes instead of double quotes.

recipe/magento2.php
```diff
@@ -367,6 +369,47 @@
...
+    //link the env to the tmp version
+    run("{{bin/symlink}} {{deploy_path}}/shared/" . TMP_ENV_CONFIG_FILE_PATH . " {{release_path}}/" . ENV_CONFIG_FILE_PATH);
```
peterjaap, line 396: Please use single quotes instead of double quotes.

recipe/magento2.php
```diff
@@ -367,6 +369,47 @@
...
+});
+after('deploy:shared', 'magento:set_cache_prefix');
+
+/**
+ * After successful deployment, move the tmp_env.php file to env.php ready for next deployment
+ */
+desc('Cleanup cache id_prefix env files');
+task('magento:cleanup_cache_prefix', function () {
+    run("rm {{deploy_path}}/shared/" . ENV_CONFIG_FILE_PATH);
```
peterjaap, line 405: Please use single quotes instead of double quotes.

recipe/magento2.php
```diff
@@ -367,6 +369,47 @@
...
+    run("rm {{release_or_current_path}}/" . ENV_CONFIG_FILE_PATH);
```
peterjaap, line 406: Please use single quotes instead of double quotes.

recipe/magento2.php
```diff
@@ -367,6 +369,47 @@
...
+    run("mv {{deploy_path}}/shared/" . TMP_ENV_CONFIG_FILE_PATH . " {{deploy_path}}/shared/" . ENV_CONFIG_FILE_PATH);
```
peterjaap, line 407: Please use single quotes instead of double quotes.

recipe/magento2.php
```diff
@@ -367,6 +369,47 @@
...
+    // Symlink shared dir to release dir
+    run("{{bin/symlink}} {{deploy_path}}/shared/" . ENV_CONFIG_FILE_PATH . " {{release_path}}/" . ENV_CONFIG_FILE_PATH);
```
peterjaap, line 409: Please use single quotes instead of double quotes.

recipe/magento2.php
```diff
@@ -367,6 +369,47 @@
...
+    $envConfigStr = "<?php return " . var_export($envConfig, true) . ";";
+    // Touch tmp config
+    run("[ -f {{deploy_path}}/shared/" . TMP_ENV_CONFIG_FILE_PATH. " ] || touch {{deploy_path}}/shared/" . TMP_ENV_CONFIG_FILE_PATH);
+    run('echo $"' . $envConfigStr . '" > {{deploy_path}}/shared/' . TMP_ENV_CONFIG_FILE_PATH);
+    //delete the symlink for env.php
+    run("rm {{release_or_current_path}}/" . ENV_CONFIG_FILE_PATH);
+    //link the env to the tmp version
+    run("{{bin/symlink}} {{deploy_path}}/shared/" . TMP_ENV_CONFIG_FILE_PATH . " {{release_path}}/" . ENV_CONFIG_FILE_PATH);
+});
+after('deploy:shared', 'magento:set_cache_prefix');
```
peterjaap, line 398: Please put this behind a toggle/configuration value and turn it off by default.

recipe/magento2.php
```diff
@@ -367,6 +369,47 @@
...
+ */
+desc('Cleanup cache id_prefix env files');
+task('magento:cleanup_cache_prefix', function () {
+    run("rm {{deploy_path}}/shared/" . ENV_CONFIG_FILE_PATH);
+    run("rm {{release_or_current_path}}/" . ENV_CONFIG_FILE_PATH);
+    run("mv {{deploy_path}}/shared/" . TMP_ENV_CONFIG_FILE_PATH . " {{deploy_path}}/shared/" . ENV_CONFIG_FILE_PATH);
+    // Symlink shared dir to release dir
+    run("{{bin/symlink}} {{deploy_path}}/shared/" . ENV_CONFIG_FILE_PATH . " {{release_path}}/" . ENV_CONFIG_FILE_PATH);
+});
+after('deploy:magento', 'magento:cleanup_cache_prefix');
```
peterjaap, line 411: Please put this behind a toggle/configuration value and turn it off by default.

peterjaap:
@valguss ok I misunderstood, this recipe does indeed run every command straight on the production server (I work on another only that does a push-only approach from a runner instance, hence the confusion).

I've added some comments.

schmengler:
@valguss what is the reason of manipulating env.php like this instead of using `bin/magento config:set --lock`? Looks overly complicated to me but I'm probably missing somethign

recipe/magento2.php
```diff
@@ -367,6 +369,47 @@
     add('shared_dirs', get('additional_shared_dirs'));
 });
 
+/**
+ * Update cache ip_prefix on deploy so that you are compiling against a fresh cache
```
schmengler, line 373: ip_prefix => id_prefix

valguss:
It's like this because the `env.php` is shared across deployments. If we set the config update during deployment then the currently deployed site can also use the updated value and pollute the cache

schmengler:
it could be set after the symlink has been switched

schmengler:
Oh, I should have read the whole conversation

> the code generation part of magento setup:di:compile uses the currently deployed cache rather than a fresh cache which can cause it to incorrectly compile the site. When this happens, another redeploy is required to compile correctly

I'm only using artifact deployment where setup:di:compile is not executed on the target server, so as long as this is not affected by default, I'm fine with whatever helps your use case

valguss:
Unfortunately, it needs to be set before then as the fresh cache is required during the Magento compile section. Without this change, the currently live and deploying codebases share the config and cache during `magento:compile`. It's specifically this section where having a fresh non-polluted cache is essential for the compilation to work correctly

recipe/magento2.php
schmengler: yes please, this task is also used by `artifact:deploy`, where the described issue is not relevant

https://github.com/deployphp/deployer/blob/master/recipe/magento2.php#L372-L382

recipe/magento2.php
```diff
@@ -367,6 +369,47 @@
...
+ * Reference Issue: https://github.com/davidalger/capistrano-magento2/issues/151
+ **/
+desc('Update cache id_prefix');
+task('magento:set_cache_prefix', function () {
+    //get current env config
+    $envConfigString = run("cat {{deploy_path}}/shared/" . ENV_CONFIG_FILE_PATH);
+    $envConfig = eval('?>'.$envConfigString);
```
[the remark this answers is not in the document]
valguss: Just been having a go at this and not sure it's possible to do what I need here. I'm trying to get the contents of a remote file and then manipulate it locally and then push up the results. `include` won't work here as far as I can tell.

The eval works as I get the text from the remote file and parse it to PHP locally.

Only other way I can think of at the moment is to download the file and include it then?

recipe/magento2.php
valguss: Added a toggle now

recipe/magento2.php
valguss: Might need a hand here too. Just couldn't;t work out how to use dump to export the array into a file

Any ideas?

recipe/magento2.php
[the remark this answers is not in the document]
valguss: Now updated to download the config file and then include it

valguss:
Now updated to use a config flag to enable

antonmedv:
@peterjaap @schmengler let's merge it and release?

schmengler (approved):

peterjaap:
@valguss could you move the if check `if (get('use_redis_cache_id')) {` to around where the hook is defined?

That way, the task itself doesn't start if `use_redis_cache_id` is not true. Now, it does start and finishes immediatly, showing up in the task log.

valguss:
Thanks @guvra i'll work on updating the repload keys as well for a future update

schmengler:
By the way, this is also helpful in edge cases if you don't compile on the target server, for example when you deploy an update where a cache type was removed:
```
[staging] run /usr/bin/php7.4 /.../releases/.../src/bin/magento cache:flush
[staging] err In ClassReader.php line 34:
[staging] err Class Vertex\Tax\Model\Cache\Type does not exist
```

PR #3456, opened 2023-01-23
Add an option to build magento artifact from the repository

guvra:
- [ ] Bug fix #…?
- [x] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [x] Docs added?

Currently, the Magento 2 recipe only allows to build the artifact file from the current working directory.

This works perfectly fine when building the artifact on a platform dedicated to deployments (e.g. gitlab CI), but it makes it nearly impossible to deploy from a local environment for the following reasons:

- A dev environment runs in developer mode.
- The vendor directory contains dev packages
- There is a possibility that the current working directory contains uncommitted files.

The goal of this PR is to add an option that works around this limitation, by cloning the project repository in a temporary folder, and creating the artifact file from this repository.

Example usage:

```
dep artifact:build localhost --option build_from_repo=1
```

```php
<?php

namespace Deployer;

require 'recipe/magento2.php';

localhost()
    ->set('local', true);

host('prod')
    ->setHostname('myprod')
    ->setRemoteUser('www')
    ->setDeployPath('/var/www/html');

set('repository', 'git@xxxxxxxxxxxxxxxx.git');
```

If no repository was specified (`repository` option), an error is displayed.

Advantages:

- No BC changes, it's just a new option.
- Very few changes to the current code of the recipe.
- Allows to easily deploy Magento projects from a local environment (more user friendly than having to create a build environment or setting up automatic deployment on a github/gitlab CI).


guvra:
docgen workflow was failing, I amended the commit to update the documentation.

Schrank:
@schmengler @peterjaap I'm not totally sure I understand what happens here. Can I have a second opinion? From my point of view, I'm happy to merge it.

peterjaap:
@Schrank I have 4 MR's to look at by the other Fabian :P @schmengler I'll try to get around to it tomorrow!

recipe/magento2.php
```diff
@@ -229,6 +229,7 @@
 set('artifact_file', 'artifact.tar.gz');
 set('artifact_dir', 'artifacts');
 set('artifact_excludes_file', 'artifacts/excludes');
+set('build_from_repo', false);
```
schmengler, line 232: A comment should be added here so that an explanation is included in the generated documentation. Also,  `set('repository', null)` with another comment. Otherwise this is going to be a secret hidden feature

schmengler (changes requested):
The feature looks fine, but documentation is missing

recipe/magento2.php
guvra: > A comment should be added here so that an explanation is included in the generated documentation. Also, `set('repository', null)` with another comment. Otherwise this is going to be a secret hidden feature

Seems right, I'll add these and push the modifications.

About the variable name (`build_from_repo`), does it seem right to you?
English is not my native language so I'm not good at naming things :sweat_smile: 

>  I'm not totally sure I understand what happens here

The current artifact mechanism only allows to build the artifact from the current working directory.
It works fine on CIs, because when a pipeline is executed, the repository is automatically cloned, and the artifact is built from there.

But if you need to deploy from a local environment (which runs in developer mode and contains lots of things that definitely shouldn't be bundled in the artifact), you have to manually clone the repository somewhere, install composer dependencies (to have access to vendor/bin/dep), then run the deployment.

This feature allows to tell Deployer to automatically clone the repo and build from it (instead of deploying from the current working directory).

About why you would want to deploy from a local environment:

- If a deployment fails, it's a lot easier to debug what happens from a local env
- Depending on the context of the project, you may not necessarily want to automate your deployments with a CI

guvra:
I added better documentation about the feature:

- A comment explain how to use `build_from_repo`
- A comment explains that `repository` must be overridden is `build_from_repo` is set to true.

schmengler:
@guvra question: How would one specify which branch should be built?
I imagine something like `dep artifact:build localhost --option build_from_repo=1 --option build_from_branch=release/1.0` would be useful

guvra:
I don't know if I should add a comment about recommending to set the `target` option (by default it uses `HEAD`, in most cases you need to deploy a tag or a specific branch instead).

schmengler:
> I don't know if I should add a comment about recommending to set the `target` option (by default it uses `HEAD`, in most cases you need to deploy a tag or a specific branch instead).

posted at the same time :D yes, please do! As you see I was confused

guvra:
@schmengler Actually in my project i set `build_from_repo` to true in deploy.php and specify `--option build_from_repo=0` in my CI config.

But it also works if you deploy with the command you provided.
For the branch, you must use the `target` option which is available by default in deployer.

recipe/magento2.php
schmengler: `build_from_repo` sounds good to me

guvra:
@schmengler I just thought of another way of doing it:

- If a target is set (with `--branch`, `--tag` or `--revision`), then the artifact is built from the repository
- Otherwise, it's built from the current working directory (current behavior)

It's more simple, but maybe also more confusing?

schmengler:
tbh, I wasn't aware of these CLI arguments, but that seems to be exactly the way they are intended to be used, like in the `update_code` recipe:

https://github.com/deployphp/deployer/blob/fe3874370b9204ec5a0a428f2f64a16dbba478b6/recipe/deploy/update_code.php#L73-L74

Since the magento recipe currently does not use the `repository` or `target` variables I don't see a real backwards compatibility issue

peterjaap:
LGTM!

peterjaap (approved):

PR #3458, opened 2023-01-24
Update Upgrade instructions

dkoether:
I noticed this sentence while upgrading a project and think it sounds not clear enough. What do you think?

### DIFF
--- a/docs/UPGRADE.md
+++ b/docs/UPGRADE.md
@@ -66,7 +66,7 @@
     ```
 11. Replace `local()` tasks with combination of `once()` and `runLocally()` func.
 12. Replace `locateBinaryPath()` with `which()` func.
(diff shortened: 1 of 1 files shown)


antonmedv:
Ok 👍

docs/UPGRADE.md
```diff
@@ -66,7 +66,7 @@
     ```
 11. Replace `local()` tasks with combination of `once()` and `runLocally()` func.
 12. Replace `locateBinaryPath()` with `which()` func.
-13. Configuration property `default_stage` is not supported.
+13. Configuration property `default_stage` is not supported anymore and has been dropped.
```
Bilge, line 69: *anymore* [isn't a word](https://dictionary.cambridge.org/grammar/british-grammar/any-more-or-anymore).

PR #3459, opened 2023-01-25
Magento: Fix artifact_path generation

schmengler:
- Fix 1. Add `-p` parameter to mkdir command for `{{artifact_path}}` could be configured with multiple folders such as build/artifacts
- Fix 2. `run()` -> `runLocally()` – The `{{artifact_dir}}` is used to keep artifacts archive file. It should be called at CI server only and not at target server. Then the task `artifact:upload` uploads this file at target server host under `{{release_path}}`. It is not needed to have `{{artifact_dir}}` at target server.

- [x] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

      $ php bin/docgen

### DIFF
--- a/docs/recipe/magento2.md
+++ b/docs/recipe/magento2.md
@@ -251,8 +251,8 @@ settings section
 
 
 ```php title="Default value"
-if (!test('[ -d {{artifact_dir}} ]')) {
-run('mkdir {{artifact_dir}}');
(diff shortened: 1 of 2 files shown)


Schrank:
Fine for me, @peterjaap ?

peterjaap:
Agree! @antonmedv merge at will.

PR #3462, opened 2023-01-25
Update default cachetool version to 9.0.0 (PHP 8.1 compatible)

schmengler:
- [ ] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

      $ php bin/docgen


Schrank:
What do you think about getting the PHP version and depending on the version we load the right cachetool version?

schmengler:
I wouldn't need it but certainly nice to have

dreadnip:
I proposed something similar a while ago: https://github.com/deployphp/deployer/pull/3375

It was merged into the V8 branch, which was deleted shortly after.

antonmedv:
Separate development in v8 branch was a mistake. Very different to track changes.

PR #3471, opened 2023-01-30
Simplify loop in releases_list

ocean90:
* Avoid calling count() in loop
* Remove extra variable for release name
* Reverse the array before for clarity


Schrank:
I love that!

PR #3477, opened 2023-02-02
import() can now load URLs and recipes from third party and private composer repositories

marvinhinz:
This is a quick proof of concept for this discussion/idea: https://github.com/deployphp/deployer/discussions/3475

It extends the existing include() function, so that URLs can be included in the deploy.php file. Also files from composer repositories can be included too. The included packages will be loaded automatically by creating a composer.json or adding them to an existing one.

Example usage in deploy.php with urls:
```
<?php
namespace Deployer;
import("https://pastebin.com/raw/bUpKm80E");
task("deploy", function() {
    writeln("deploy!");
    invoke("test");
});
localhost();
```
Output of dep deploy:
```
mhinz@dev:/srv/htdocs/deployer$ bin/dep deploy
 Do you really want to trust this remote recipe: https://pastebin.com/raw/bUpKm80E? (yes/no) [yes]:
 > y
task deploy
[localhost] deploy!
task test
[localhost] test!
```

Example usage in deploy.php with private composer repo:
```
<?php
namespace Deployer;
import("recipe/test.php", "some/custommodule:dev-master", "https://composer.example.org");
task("deploy", function() {
    writeln("deploy!");
    invoke("test");
});
localhost();
```
Output of dep deploy:
```
mhinz@dev:/srv/htdocs/deployer$ bin/dep deploy
 Do you really want to trust this remote recipe: recipe/test.php? (yes/no) [yes]:
 > y
task deploy
[localhost] deploy!
task test
[localhost] test!
```

This is especially useful when running a lot of projects with similar custom recipes (e.g. in agencies). The automated composer require is probably not really needed, as we could just throw an exception for missing composer packages. But i thought its a bit nicer for the dev experience.

There is still a lot of room for improvement, for example: the deploy.php configuration is loaded before the main symfony console application is running, this results in various little problems (to ask a user an interactive question, a command has to be run manually. Deployer::isWorker() always returns false, because the value is set too late (when the worker command is executed). 

Should the user even be asked for confirmation? In theory it is a risk, but in comparison running the usual `wget URL | bash` commands are risky too. 

- [ ] Bug fix #…?
- [x] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

### DIFF
--- a/src/Command/ImportCommand.php
+++ b/src/Command/ImportCommand.php
@@ -19,7 +19,6 @@ use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Question\ConfirmationQuestion;
 use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\Process\Exception\RuntimeException;
-use Symfony\Component\Process\PhpProcess;
 use Symfony\Component\Process\Process;
 
 class ImportCommand extends Command
@@ -28,8 +27,9 @@ class ImportCommand extends Command
 
     private InputInterface $input;
     private OutputInterface $output;
+    private string $cwd;
 
-    protected function configure()
+    protected function configure(): void
     {
         $this
             ->setName('import')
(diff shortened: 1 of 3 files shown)


antonmedv:
Can we add some tests? Something like e2e?

Schrank:
We can add a test, which loads a file from this composer repository :-)

marvinhinz:
Sure will do :)

antonmedv:
And update to the documentation ;)

PR #3486, opened 2023-02-14
Contrib supervisord monitor

mwagena:
- [ ] Bug fix #…?
- [x] New feature?
- [ ] BC breaks?
- [x] Tests added?
- [x] Docs added?

      $ php bin/docgen

This pull requests adds the possibility to restart Supervisord processes from within deployer when cli rights are insufficient.


mwagena:
Sorry for the commit messages :) had a major conflict with the autoformatter of my IDE...

contrib/supervisord-monitor.php
```diff
@@ -0,0 +1,153 @@
...
+{
+    checkConfig();
+
+    $authResponseInfo = [];
+    Httpie::post(get('supervisord_uri'))->header('Authorization', getBasicAuthToken())->send($authResponseInfo);
+
+    return $authResponseInfo['http_code'] === 200;
+}
+
+function action($name, $action = 'stop')
```
antonmedv, line 122: This very broad name in Deployer namespace. 

Please refactor all function to use as variables via set(name, function).

antonmedv (changes requested):

contrib/supervisord-monitor.php
mwagena: I renamed the function names to be preficed with the recipe name.

What do you mean by refactoring all functions to use as variables via set(name, function)
There should only be one action available: ```supervisord-monitor:restart```

mwagena: @antonmedv updated the code with one last change so that stop and start are also available tasks and also now wrapped so that it won't interfere with the other code inside the Deployer namespace

antonmedv (approved):

PR #3326, opened 2022-09-30
Split asset compilation

renttek:
This PR adds the improvements from #3311 
The new variables have comments explaining the supported configuration options/structure

PR #3510, opened 2023-02-24
Adds Comments to magento2 recipe and modifies DocGen, to be able to explain artifact deployment

wilfriedwolf:
Hi there,

I added the documentation to the tasks and settings. Also I modified the DocGen class a little to allow the artifact deployment tasks to be explained more prominently.
This can be reused in every recipe, so if (grouped) tasks `artifact:build` and `artifact:deploy` exist in a recipe an additional section will be added to the docs.

Best regards,

Willi


recipe/magento2.php
```diff
@@ -98,12 +98,38 @@
 set('enable_zerodowntime', true);
 
 // Tasks
+
+// To work correctly with artifact deployment, it is necessary to set the MAGE_MODE correctly in `app/etc/config.php`
```
akosglue, line 102: Does it get higher priority than the env file?
Or it has nothing to do with the environment value and not will override. (i.e. magento ignores it)
(in fact this could have maybe an other name like DEPLOYER_MAGE_MODE) ?

recipe/magento2.php
```diff
@@ -231,23 +257,33 @@
 
 after('deploy:failed', 'magento:maintenance:disable');
 
-// artifact deployment section
-// settings section
+// artifact deployemnt section
```
peterjaap, line 260: Missing capitcal, and typo

wilfriedwolf: Just changed and pushed...

recipe/magento2.php
wilfriedwolf: @peterjaap Am I correct about the usage of MAGE_MODE and the dev-Settings (i.e. the commands where they are used in) I was not quite sure although I know that they are required in order to work correctly.

akosglue: where can I find actually the docs/source that it is needed for deployer? and also, is it is common magento practice? where should it stay exactly, is it under system maybe?

wilfriedwolf: Well, I just wrote something but now I am less sure than before... I vaguely remember, that if you do not set the correct i.e. the target MAGE_MODE, than it will not work. Anyhow calling `setup:di:compile` without `MAGE_MODE` set also works (not sure about the outcome though) Looking up the code to find where DI compilation depends on MAGE_MODE I could not find anything. Maybe @peterjaap knows more...

akosglue: I have only experience with MAGE_MODE in the env.php file or in .htaccess (or nginx config), not yet heard of it in the config.php actually.

peterjaap: MAGE_MODE has nothing to do with Deployer as such, but to actually be able to deploy artifacts correctly, Magento needs to be in production mode.

peterjaap: And seeing that `env.php` files are usually not present in a build environment, adding it to `config.php` makes sense.

peterjaap:
@antonmedv LGTM, merge at will

recipe/magento2.php
peterjaap: @wilfriedwolf another option would be this; https://github.com/deployphp/deployer/pull/3326/files#r1009708393

PR #3512, opened 2023-02-24
Fixed bin/console calls in shopware recipe

dahormez:
- [ x ] Bug fix: **Shopware bin/console calls with default bin/php and not with {{bin/php}}**
I was running the recipe with {{bin/php}} = /opt/local/bin/php8.2 but my default cli version is php7.4 the deployment stopped caused of the wrong php version by calling bin/console directly.


Schrank:
Awesome, thanks @dahormez! <3

PR #3522, opened 2023-03-01
Added support for ntfy.sh notifications, contrib/ntfy

HereticPilgrim:
- [ ] Bug fix #…?
- [x] New feature: added support for ntfy.sh notifications
- [ ] BC breaks?
- [ ] Tests added? --> No. If you can give me some pointers on how to do it, I will gladly add them.
- [x] Docs added?

      $ php bin/docgen


HereticPilgrim:
All other contrib recipes use `{{target}}` for the default notification.
This results in notifications like:
```_HereticPilgrim_ deploying `main` to *main*```

I used `{{hostname}}` as it's more intuitive for me, resulting in:
```_HereticPilgrim_ deploying `main` to *example.com*```

If you want, I can standardize on `{{target}}`

antonmedv:
Lets fix info messages)

HereticPilgrim:
Can you elaborate on what you mean by "fix"?
Would you prefer `{{hostname}}` in the other contrib-recipes or should I change my version to include `{{target}}`?

antonmedv:
Can you show an example of what do you think?

Issue #296, opened 2015-05-26
The command "deploy" is not defined.


ZhukV:
I install the Deployer to `/usr/local/bin/dep`, and create a deploy.php in root directory with content:

```php
server('production', 'foo-bar.com', 22)
    ->user('foobar')
    ->forwardAgent()
    ->stage('production')
    ->env('deploy_path', '/var/www/foo-bar.com');

set('repository', 'git@bitbucket.org:foo/bar.git');
```

But, if i run `dep deploy production` i see a error:

```
  [InvalidArgumentException]        
  Command "deploy" is not defined.
```

Why? Thank.

elfet:
Require recipe at beginning of file:
```php
require 'recipe/common.php';
```

ZhukV:
Thank. All ok.

iamrutvik:
But I have added recipe file.
Here it is : 
`


require 'vendor/deployer/deployer/recipe/laravel.php';



server('prod', '192.168.91.128', 22)
    ->user('ubuntu')
	->password('e2m@123')
    ->stage('production')
    ->env('deploy_path', '/var/www/html'); // Define the base path to deploy your project to.

set('svnrepo', 'https://192.168.1.148/svn/dev-autodeployment/trunk/autodeploy');
set('svnuser', 'rutvikb');
set('svnpass', 'rutv@123');

task('deploy:update_code', function () {
    $svn = 'C:/Program Files/TortoiseSVN/bin';
    $repository = trim(get('svnrepo'));
    $user = trim(get('svnuser'));
    $pass = trim(get('svnpass'));
    run("$svn export --force --username $user --password $pass $repository {{release_path}} 2>&1");
})->desc('Updating code');`

PR #3526, opened 2023-03-02
Added missing deploy task to recipe/common

HereticPilgrim:
`recipe/common.php` currently does not define a deploy task and is the only recipe that's missing it.
Newcomers may stumble over this behavior, as the documentation asks you to run `dep deploy` as one of the first steps.
This PR fixes this.

It's compatible with recipes that are based on `common`, as overwriting tasks is allowed (e.g. common > symfony > contao).

Note: It appears that this task was previously defined (see #296).

- [ ] Bug fix #…?
- [x] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [x] Docs added? --> updated.

      $ php bin/docgen

### DIFF
--- a/docs/recipe/common.md
+++ b/docs/recipe/common.md
@@ -225,8 +225,21 @@ This task is group task which contains next tasks:
 * [deploy:success](/docs/recipe/common.md#deploysuccess)
 
 
+### deploy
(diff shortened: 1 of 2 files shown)


antonmedv:
Makes sense.

Issue #3290, opened 2022-09-07
Spaces in ssh_arguments are not respected


Dachande663:
- Deployer version: 7.0.1
- Deployment OS: Ubuntu 22.04

````
host('123.123.123.123')->set('ssh_arguments', ['-o ProxyCommand="ssh user@bastion.host"']);
````

When setting an SSH argument on a host that includes a space, the connectionOptionsArray method calls explode on each argument with a space as the split character. Unfortunately, this explode doesn't respect spaces within protected whitespace e.g. between the "ssh" and "user@bastion" in the example above. This means the argument is split here, escaped again, and results in a bash error:

`/bin/sh: 1: Syntax error: Unterminated quoted string`

Is it possible to remove the explode, or if it's required, make sure it respects spaces within quote fences?

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/deployphp) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/deployphp/deployer/issues/3290">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/deployphp/deployer/issues/3290/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/deployphp/deployer/issues/3290/pledge.svg">
</picture>
</a>

antonmedv:
I see. Probably we need to refactor this part to keep old syntax and remove explode.

alinalexandru:
see my comments on https://github.com/deployphp/deployer/issues/3558

PR #3534, opened 2023-03-08
Fixing ssh_arguments to not explode to support arguments that contain spaces

Luwdo:
Fixing ssh_arguments items to not be exploded which has a side effect of causing escapeshellarg to quote into arguments values containing spaces.

For example:

```
->setSshArguments(["-o ProxyCommand='ssh -A jump@1.2.3.4-W %h:%p'"])
```

Fixes this issue: https://github.com/deployphp/deployer/issues/3290

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 *.phar
 .phpunit.result.cache
 docker-compose.override.yml
+/.idea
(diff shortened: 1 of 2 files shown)


antonmedv:
I'm not sure this will fix the issue. As in the https://github.com/deployphp/deployer/blob/f3bafec466cc945a8249899c5d7f65072d3e32b2/src/Component/Ssh/Client.php#L65

we need the array with arguments. I think here is the problem with the API. I think we can fix it in the v8 by replacing API to 

```php
->setSshArguments(["ProxyCommand" => "ssh -A jump@1.2.3.4-W %h:%p"])
```

Issue #3402, opened 2022-12-14
Add the specific Laravel artisan specific command for notify Bugsnag of a deployment


andrecolza:
Laravel Bugsnag package has the artisan command `php artisan bugsnag:deploy` to notify Bugsnag of new deployment, which is easier to use than `bugsnag:notify` task in the Bugsnag contrib recipe.
It would be useful if a new Laravel-specific `artisan:bugsnag:deploy` task was added (in Bugsnag recipe?) to allow the use of this command in Laravel project deployments.

Schrank:
Hi @andrecolza thanks for the idea! A good one. Can you provide a PR for it?

antonmedv:
Closing as inactive

PR #3557, opened 2023-03-29
Added `artisan:bugsnag:deploy` task

andrecolza:
Added task to notify BugSnag of releases using the Laravel Artisan console.

- [ ] Bug fix #…?
- [x] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [x] Docs added?

      $ php bin/docgen

BugSnag provides a Laravel Artisan `bugsnag:deploy` command to report releases. This PR adds a task in `contrib/bugsnag.php` to allow use that command when deploying.

Note: I required the Laravel recipe to allow the use of `artisan` function, not sure if is the right way.

This is related to #3402

### DIFF
--- a/contrib/bugsnag.php
+++ b/contrib/bugsnag.php
@@ -3,23 +3,39 @@
 
 ## Configuration
 
-- *bugsnag_api_key* – the API Key associated with the project. Informs Bugsnag which project has been deployed. This is the only required field.
+- *bugsnag_api_key* – the API Key associated with the project. Informs BugSnag which project has been deployed. This is the only required field.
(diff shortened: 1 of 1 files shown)


antonmedv:
Let’s put in the laravel recipe.

andrecolza:
@antonmedv Ok, I can open a second PR to put the new task it in the Laravel recipe, but I think it shouldn't be placed there: the Laravel recipe only contains tasks related to first-party commands, while this is a command provided by a third-party package (BugSnag).

The only other case I've found in Deployer is about Statamic, which has its own recipe that extends the Laravel recipe.
Are you sure you prefer this task to be added in the Laravel recipe?

antonmedv:
Definitely for importing laravel in contrib.

PR #3567, opened 2023-04-06
DD#0000: feat: Magento2: Add support for preload keys

valguss:
- Also updated to remove the config set and instead manually add the hocks to your deployer file
- And move var-exporter into the main composer require
- https://github.com/deployphp/deployer/pull/3453#issuecomment-1497248859

- [ - ] Bug fix https://github.com/deployphp/deployer/pull/3453#issuecomment-1497248859
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ - ] Docs added?

      $ php bin/docgen


composer.json
```diff
@@ -50,6 +50,7 @@
         "symfony/polyfill-php80": "^1.22",
         "symfony/polyfill-php81": "^1.26",
         "symfony/process": "^5",
+        "symfony/var-exporter": "^5.4",
```
antonmedv, line 53: Why is got here?

valguss: Yeah I noticed i'd added it to the dev packages where infact I actually use it in the recipe so need it as part of the main packages

recipe/magento2.php
```diff
@@ -443,7 +440,11 @@ function magentoDeployAssetsSplit(string $area)
 /**
  * Update cache id_prefix on deploy so that you are compiling against a fresh cache
  * Reference Issue: https://github.com/davidalger/capistrano-magento2/issues/151
- * use set('use_redis_cache_id') in your deployer script to enable
+ * use add the following to your deployer scripts:
```
schmengler, line 443: Grammar: "To use this feature, add the following to your deployer scripts:"

recipe/magento2.php
```diff
@@ -454,6 +455,17 @@ function magentoDeployAssetsSplit(string $area)
     //set prefix to `alias_releasename_`
     $prefixUpdate = get('alias') . '_' . get('release_name') . '_';
 
+    //check for preload keys and update
+    if (isset($envConfigArray['cache']['frontend']['default']['backend_options']['preload_keys'])) {
+        $oldPrefix = $envConfigArray['cache']['frontend']['default']['id_prefix'];
+        $preloadKeys = $envConfigArray['cache']['frontend']['default']['backend_options']['preload_keys'];
+        $newPreloadKeys = [];
+        foreach ($preloadKeys as $preloadKey) {
+            $newPreloadKeys[] = str_replace($oldPrefix, $prefixUpdate, $preloadKey);
```
schmengler, line 464: to not accidentally replace parts of the cache keys if they match the old prefix, I would consider only replacing the prefix at the beginning of the string. Unlikely, I know, but using `substr_replace` or a regular expression seems more appropiate to me

schmengler (changes requested):

recipe/magento2.php
valguss: Thanks @schmengler 
That's now updated

recipe/magento2.php
valguss: Thanks @schmengler 
That's now updated

recipe/magento2.php
schmengler: "use add the following [...]" is still an incomplete sentence. If you want to keep it short: "Usage: add the following [...]"

recipe/magento2.php
```diff
@@ -440,7 +440,7 @@ function magentoDeployAssetsSplit(string $area)
 /**
  * Update cache id_prefix on deploy so that you are compiling against a fresh cache
  * Reference Issue: https://github.com/davidalger/capistrano-magento2/issues/151
- * use add the following to your deployer scripts:
+ * To use this feature, add the following to your deployer scripts:
```
schmengler, line 443: forget my last comment, this is good now :smile:

schmengler (approved):

antonmedv:
Ok for me. Let's get 👍🏻 from someone with Magento experience and merge.

peterjaap:
@antonmedv 👍

schmengler (approved):

PR #3570, opened 2023-04-06
Install composer if outdated

Ernestopheles:
I added a check in vendors.php: If composer already installed but versiion < 2.0.0., than install composer newly.

I need this for Contao CMS as composer >= 2.0.0 is required.


Schrank:
Sounds reasonable to me. But can you please NOT delete composer.lock? Thanks :-)

Ernestopheles:
Sorry, this was not intended. Would you, please, be so kind to fix that for me?

recipe/deploy/vendors.php
```diff
@@ -1,23 +1,28 @@
...
-// composer version to `.dep/composer.phar`. To use specific composer version
-// download desired phar and place it at `.dep/composer.phar`.
+// If Composer is installed and version >= 2.0.0.: Returns Composer binary path in found.
+// Otherwise try to install latest composer version to `.dep/composer.phar`.
+// To use specific composer version, download desired phar and place it at `.dep/composer.phar`.
 set('bin/composer', function () {
     if (test('[ -f {{deploy_path}}/.dep/composer.phar ]')) {
-        return '{{bin/php}} {{deploy_path}}/.dep/composer.phar';
+        $composerVersion = run('{{bin/php}} {{deploy_path}}/.dep/composer.phar --version');
+        if (version_compare($composerVersion, '2.0.0', '>=')) {
```
antonmedv, line 15: Not sure if it is a good solution. Let's keep the version what user installed by default.

antonmedv (changes requested):

Ernestopheles:
Most often, I assume, it's the Hoster who not updated Composer. But anyway, may be some users want to keep an outdated version. So I agree to close this PR.

I can as good implement my own task to update composer.

PR #3586, opened 2023-04-29
changes for using latest shopware flex template

xavgru12:
- Bug fix: the theme is missing on deployed version. Compiling the theme in the sw:deploy task fixes it. Another fix: in CI systems there is no way to compile the theme locally since a database connection is needed. Adding the theme compile skip option at the build-js.sh fixes it as well.

### DIFF
--- a/recipe/shopware.php
+++ b/recipe/shopware.php
@@ -8,12 +8,20 @@
  * set('repository', 'git@github.com:shopware/production.git');
  * ```
  *
+ * configure host:
+ * host('SSH-HOSTNAME')
+ *     ->setRemoteUser('SSH-USER')
+ *     ->set('deploy_path', '/var/www/shopware') // This is the path, where deployer will create its directory structure
+ *     ->set('http_user', 'www-data') // Not needed, if the `user` is the same user, the webserver is running with 
+ *     ->set('writable_mode', 'chmod');
+ * 
  * :::note
  * Please remember that the installation must be modified so that it can be
  * [build without database](https://developer.shopware.com/docs/guides/hosting/installation-updates/deployments/build-w-o-db#compiling-the-storefront-without-database).
  * :::
  */
 namespace Deployer;
+
 
 require_once __DIR__ . '/common.php';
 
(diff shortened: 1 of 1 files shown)


UlrichThomasGabor:
What exactly was not working? I don't understand from this issue here.

The recipe was implemented having this site in mind: [build SW without DB](https://developer.shopware.com/docs/guides/hosting/installation-updates/deployments/build-w-o-db.html), or alternatively this one [official PaaS documentation](https://developer.shopware.com/docs/products/paas/theme-build.html). Building should happen locally and it was working without a database connection before, because the mentioned files were obtained from the server. From what I understand, you have disabled part of the building? When will building the theme happen now?

From what I see at the moment, I think merging this PR broke things, and without further explanation I think it should be reverted.

@antonmedv @Schrank

Schrank:
Unfortunately I'm not using the shopware recipe at the moment, but a very old version which works for us.

@xavgru12 did you follow any of the mentioned links by @UlrichThomasGabor? Because to build it locally you need to adjust your configuration.

I tend to revert it but add a paragraph to the config about the two links Ulrich provided, so people see in the recipe how to use it. Any comments on this idea?

UlrichThomasGabor:
Sounds good to me.

UlrichThomasGabor:
> When will building the theme happen now?

I just saw that you added a remote task for that. While I guess this is working, it is in my opinion unnecessary to compile the theme on each deployed instance, when all data is already available to compile locally (saves time and energy).

Without further explanation on what problem this issue is actually solving, I would still vote for a revert.

xavgru12:
I have followed the guide of compiling without Database connection. 
Using without database connection, I had to skip the theme compilation as it was described in the guide.

This resulted in the deployed version not having the theme compiled. Adding the remote task resolves this.

UlrichThomasGabor:
You are referring to the last part?
https://developer.shopware.com/docs/guides/hosting/installation-updates/deployments/build-w-o-db.html#partially-compiling-the-storefront

This is optional. You can add that via a switch, but I don't think it is good practice to change the behavior for everyone.

xavgru12:
Compiling the storefront without database does not work for me following the guide. Using the theme skip works fine, however.

UlrichThomasGabor:
Feel free to open an issue or ask on Stackoverflow or SW Slack for help, if you encounter problems with the setup. You can also change your recipe locally (redefine the tasks). But in my opinion, the standard case should be the default one. It is definitely working, since we are using this process for 2 years now or something like that.

Issue #3478, opened 2023-02-02
Deployment breaks when done from different platforms (e.g. Windows and Mac)


klaeser:
- Deployer version: 7.1.3
- Deployment OS: Windows and Mac

      Please, provide a minimal reproducible example of deploy.php
it can be any deploy.php which uses the default "deploy:release" task

Steps:
1. do a deployment from Windows
2. do a deployment of the same project from Mac to the same destination
=> Error: "undefined array key 2" in release.php on line 49

Explanation:
For each deployment, a line is added to the file .dep/release_log
This is done in vendor/deployer/deployer/recipe/deploy/release.php in line 126 with the following code
`    $json = escapeshellarg(json_encode($metainfo));`
Unfortunately, in Windows, escapeshellarg replaces all double quotes in the string by spaces.
https://www.php.net/manual/en/function.escapeshellarg.php
This results for example in the following entry:
`{ created_at : 2023-01-12T09:44:54+0000 , release_name : 7 , user : Firstname Lastname , target : HEAD }`
When this line is read and parsed by deployer on a different OS (e.g. Mac), it can not be parsed.
It can only be parsed when the double quotes are contained:
`{"created_at":"2023-01-12T09:44:54+0000","release_name":"7","user":"Firstname Lastname","target":"HEAD"}`

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/deployphp) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/deployphp/deployer/issues/3478">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/deployphp/deployer/issues/3478/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/deployphp/deployer/issues/3478/pledge.svg">
</picture>
</a>

fritzmg:
I can confirm the problem. However, in my case it does not really seem to be related to the platform, as I _always_ deploy from Windows and _always_ from the Git Bash. Nevertheless, I ended up with a `release_log` like this:

```
{"created_at":"2022-12-20T11:29:41+0000","release_name":"1","user":"someusername","target":"somebranch"}
{"created_at":"2022-12-20T11:32:02+0000","release_name":"2","user":"someusername","target":"somebranch"}
[...]

PR #3603, opened 2023-05-18
Escape shell arguments correctly

fritzmg:
- [x] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

This is my take on fixing https://github.com/deployphp/deployer/issues/3478. The current usage of `escapeshellarg` is wrong, as it will executed on the local system and thus not escape the shell argument appropriately - since the argument is used for a command that is run on the _remote_ system.

This PR basically creates an `escape_shell_argument` that can be made aware of the remote host's operating system. It is a copy of [Symfony\Component\Process\Process::escapeArgument](https://github.com/symfony/symfony/blob/825fd035c1732b62d688fec8bada9f9ad0448835/src/Symfony/Component/Process/Process.php#LL1561C1-L1578C6) and allows to override the directory separator.

So this solution first resolves the directory separator on the remote system and then escapes the shell argument based on that detection.


antonmedv:
Lets use quote function from my other project:

https://github.com/google/zx/blob/d1cd1aa02e3fab7ca4f048d9571b77b6ce25d559/src/util.ts#L31-L48

fritzmg:
> Lets use quote function from my other project:
> 
> https://github.com/google/zx/blob/d1cd1aa02e3fab7ca4f048d9571b77b6ce25d559/src/util.ts#L31-L48

But that code would create the following command, always:

```
echo '{"created_at":"2023-05-19T14:29:03+0000","release_name":"16","user":"foobar","target":"HEAD"}' >> .dep/releases_log
```

This would again produce a wrong `releases_log` if the target server is Windows based.

fritzmg:
@antonmdev I have adjusted the PR now to the new escaping method, under the assumption that deployments onto a Windows server are not really supported (and not common place) anyway.

antonmedv:
> if the target server is Windows based 

This is true. Only Linux.

antonmedv:
Let's add some test for this function.

fritzmg:
I added some basic tests.

s4muel:
at a first glance it won't work with single quotes in string, will it? i mean, the escaped string will be OK (not making problems when running the shell command), but it won't produce valid JSON. taking Erik's name as an example
```
escape_shell_argument(json_encode(['name' => "Erik D'Ercole"]));
```
this will produce `'{"name":"Erik D\'Ercole"}'` which again is not JSON and will cause the same problem

fritzmg:
> at a first glance it won't work with single quotes in string, will it?

It will. Single quotes will be replaced by `\'`.

> which again is not JSON

What makes you say that? Single quotes must be escaped by `\` in JSON.

s4muel:
well, AFAIK there is no escaping of single quotes in JSON. this is valid JSON:
```
{
  "name": "Erik D'Ercole"
}
```
this is invalid JSON:
```
{
  "name": "Erik D\'Ercole"
}
```
not a pro here, just used random JSON online validator in combination with https://www.json.org/json-en.html
i might be wrong

fritzmg:
> well, AFAIK there is no escaping of single quotes in JSON.

While it's not technically necessary, it's good practise to do so. PHP's `json_encode` does it as well. The info will be read by PHP anyway.

s4muel:
```
function escape_shell_argument(string $argument): string
{
    if ('' === $argument || preg_match('~^[a-z0-9/_.\-@:=]+$~i', $argument)) {
        return $argument;
    }

    return "'".str_replace(
        ['\\', "'", "\f", "\n", "\r", "\t", "\v", "\0"],
        ['\\\\', "\\'", '\\f', '\\n', '\\r', '\\t', '\\v', '\\0'],
        $argument,
    )."'";
}

$json = escape_shell_argument(json_encode(['name' => "Erik D'Ercole"]));

echo $json, PHP_EOL;
var_dump(json_decode($json));

echo 'Last error: ', json_last_error_msg(), PHP_EOL, PHP_EOL;


$json2 = <<<EOI
{"name": "Erik D\'Ercole"}
EOI;

echo $json, PHP_EOL;
var_dump(json_decode($json));

echo 'Last error: ', json_last_error_msg(), PHP_EOL, PHP_EOL;
```
produces:
```
'{"name":"Erik D\'Ercole"}'
NULL
Last error: Syntax error

'{"name":"Erik D\'Ercole"}'
NULL
Last error: Syntax error
```

you sure? am i doing something wrong or missing something?

fritzmg:
Sorry, you are right. And escaping with `\'` on the command line level is not sufficient.

Instead, the command needs to look like this:

```
echo '{"name":"Erik D'\''Ercole"}' >> .dep/releases_log
```

which will result in

```
{"name":"Erik D'Ercole"}
```

in the `.dep/releases_log`.

And this is exactly what [`Symfony\Component\Process\Process::escapeArgument`](https://github.com/symfony/symfony/blob/825fd035c1732b62d688fec8bada9f9ad0448835/src/Symfony/Component/Process/Process.php#LL1561C1-L1578C6) does.

I have updated the PR accordingly and boiled it down to the strictly necessary pieces (assuming a Linux server target).

@s4muel can you please test these changes again?

fritzmg:
@antonmedv this should be the correct solution. Any chance of getting this merged and released?

ttk:
I've been hit by this bug.  I've reviewed this PR and it looks good to me!

PR #3639, opened 2023-07-16
fix completion templates skipping

### LINKED ISSUE #3366: `deb completion` no supported shells
I have installed Deployer using the [installation instructions](https://deployer.org/docs/7.x/installation) to an existing Laravel project.

```
composer require --dev deployer/deployer
./vendor/bin/dep init
```

If I try to generate the bash completion file I get the following error:

```
./vendor/bin/dep completion bash
Detected shell "bash", which is not supported by Symfony shell completion (supported shells: "").
```

Looking at the [Symfony code](https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/Console/Command/DumpCompletionCommand.php) it's looking for 

```
__DIR__.'/../Resources/completion.*'
```

as the supported shells but can't find that.

I've tried running `dep` from different directories, with and without an alias but always with the same results.

The `completion.bash` file does exist:

```
$ ls vendor/symfony/console/Resources/completion.bash 
vendor/symfony/console/Resources/completion.bash
```

Also `artisan completion` does work as expected.

RecklessTimon:
- [x] Bug fix #3366?

### DIFF
--- a/bin/build
+++ b/bin/build
@@ -5,7 +5,6 @@
  * For the full copyright and license information, please view the LICENSE
  * file that was distributed with this source code.
  */
-
 if (ini_get('phar.readonly') === '1') {
     throw new \Exception('Writing to phar files is disabled. Change your `php.ini` or append `-d phar.readonly=false` to the shebang, if supported by your `env` executable.');
 }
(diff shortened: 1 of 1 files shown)


CodeByZach:
I still run into issue #3366 on version 7.3.1

antonmedv:
We need to upgrade Symfony components for this to work. This is planned for v8.

CodeByZach:
Appreciate the quick response! So you're saying it's likely not something wrong with my config. And I should just wait for v8, right?

MartinMa:
Meanwhile, is there a workaround to manually add the completion templates to zsh (default terminal on MacOS)?

PR #3649, opened 2023-07-23
SshCommand: Fix host selection

gharlan:
- [x] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

### DIFF
--- a/src/Command/SshCommand.php
+++ b/src/Command/SshCommand.php
@@ -69,7 +69,7 @@ class SshCommand extends Command
             }
 
             if (count($hostsAliases) === 1) {
-                $host = current($this->deployer->hosts->all());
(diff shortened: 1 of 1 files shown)


src/Command/SshCommand.php
```diff
@@ -69,7 +69,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
             }
 
             if (count($hostsAliases) === 1) {
-                $host = current($this->deployer->hosts->all());
+                $host = $this->deployer->hosts->get($hostsAliases[0]);
```
gharlan, line 72: I have two hosts:
- host 0: `Localhost`
- host 1: remote `Host`

The Localhost is removed from `$hostAliases`. But in this line it was chosen anyway.

Issue #3542, opened 2023-03-21
Please login as the user \"ubuntu\" rather than the user \"root\ during provisioning


florentdestremau:
- Deployer version: 7.2.1
- Deployment OS: ubuntu 20.04

I have a OVH VPS that I provisioned. I can only login via ssh on the `ubuntu` user with sudo enabled. I tried the `dep provision -o become=root -o remote_user=ubuntu` even while having the `deployer` user in my deploy.php config (as stated in the docs).

Everything went fine until `provision:website` where it seems to force the deployer user but I can't access the server with this user. When investigating I ended up finding that my rsa key was added like this:

```
# /home/deployer/.ssh/authorized_keys
no-port-forwarding,no-agent-forwarding,no-X11-forwarding,command="echo 'Please login as the user \"ubuntu\" rather than the user \"root\".';echo;sleep 10;exit 142" ssh-rsa XXXX-MY-KEY-XXXXXX
ssh-rsa XXXX-MY-KEY-XXXXXX
```
As you can see my key is added twice AND with a guard preventing me to login.

Looking further it seems it's copying my standard root user `authorized_keys` instead of creating a usable configuration.

dvdinbox:
@antonmedv @florentdestremau 

Is there any workaround to this? I'm trying to deploy to an aws ec2 and it reports

```
[ec2-****.us-east-2.compute.amazonaws.com]  error  in website.php on line 16:
[ec2-****.us-east-2.compute.amazonaws.com] run [ -d ~/public ] || mkdir ~/public 
[ec2-****.us-east-2.compute.amazonaws.com] Please login as the user "ubuntu" rather than the user "root".
```

Here's some things I tried

1) Allowed root ssh into the instance. The strange thing is I'm now able to ssh as root, but deployer still returns  the 'Please login as the user "ubuntu" rather than the user "root"' error

2) Tried to manipulate the problem line of code in provision:website. I thought I'd be able to make changes in the local vendor\deployer\deployer\recipe\provision\website.php file and see the effects of the changes in real time. I can't figure out to make changes to code and test the effect of the changes. Note that I'm relatively new to this php ecosystem.

florentdestremau:
> > Looking further it seems it's copying my standard root user authorized_keys instead of creating a usable configuration.
> 
> Didn’t get this part.

[...]

PR #3669, opened 2023-08-12
Expanded Caddyfile - www redirect, rewrites, encoding, transport headers

nikspyratos:
Relates to #3666 

Expands Caddyfile to include a www redirect, rewrites for sensitive files, encoding, and transport headers - all from [here](https://ma.ttias.be/caddyfile-config-example-for-laravel/). The only change not added is defaulting to index.php as 1. I believe the existing config already covers that, and 2) I'm not sure every PHP framework may be using that file (I assume most do, but I'm not sure).

Only change to existing functionality is doing a `str_ireplace` to remove `www.` from domain inputs, so that the new config will cover the www redirect.

This is a draft because I've been struggling to test it. How does Deployer work when installed in repositories? No matter what changes I make to the files in `vendor` in my application, the changes are never followed. 

Side note: Additionally, it's a mission to get provisioning and deployment working out of the bat due to #3542 and #3147 - I'm testing on EC2. The workaround is to essentially follow bad security practices and allow root login into my server to get this to work, which isn't great.

### DIFF
--- a/docs/recipe/provision/website.md
+++ b/docs/recipe/provision/website.md
@@ -18,7 +18,7 @@ require 'recipe/provision/website.php';
 
 
 ```php title="Default value"
-return ask(' Domain: ');
+return str_ireplace('www.', '', ask(' Domain: '));
 ```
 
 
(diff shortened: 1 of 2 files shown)


nikspyratos:
Think I have to redo this - seems these directives are Caddy 1 directives and are not valid in Caddy 2.

PR #3682, opened 2023-09-08
feat: add bun recipe

innocenzi:
In the same vein as https://github.com/deployphp/deployer/pull/3677, this pull request adds support for installing dependencies and running the build command with [bun](https://bun.sh/), which just hit 1.0.


innocenzi:
Cleaning up my PRs, someone else feel free to contribute that, we're now using our own recipe.

PR #3684, opened 2023-09-11
Allow multiple cachetool calls

infabo:
- [x] Bug fix #…? No
- [x] New feature? Yes
- [x] BC breaks? I would say no. Possibly breaking for people that modified the recipe's tasks.
- [x] Tests added? No tests I am aware of. So none adjusted.
- [x] Docs added? Docs regenerated

      $ php bin/docgen


contrib/cachetool.php
```diff
@@ -64,36 +66,51 @@
...
     $options = get('cachetool');
     $fullOptions = get('cachetool_args');
+    $return = [];
 
     if (strlen($fullOptions) > 0) {
-        $options = "{$fullOptions}";
+        $return = ["{$fullOptions}"];
     } elseif (strlen($options) > 0) {
-        $options = "--fcgi={$options}";
+        if (!is_array($options)) {
```
Schrank, line 74: You can shorten that by just casting it to array - if it is a string it becomes ['string'], if it is array it stays as it is :-)

contrib/cachetool.php
```diff
@@ -64,36 +66,51 @@
 set('cachetool_options', function () {
     $options = get('cachetool');
     $fullOptions = get('cachetool_args');
+    $return = [];
 
     if (strlen($fullOptions) > 0) {
-        $options = "{$fullOptions}";
+        $return = ["{$fullOptions}"];
```
Schrank, line 72: Is there a reason for "{ because imho you can remove it and just have `[$fullOptions]`

Schrank (changes requested):
That looks great! Interesting idea :-)

contrib/cachetool.php
infabo: It was there before. Did not change it. But yes, it is not needed.

Schrank: Thanks for the cleanup :-)

Schrank (approved):
Looks good to me!

yard-mschwartz:
It turns out we had a copy of the default opcache clear command in `deploy.php`, which broke the our deployment pipeline since v7.3.2, that includes this change. Only God knows why. :upside_down_face: 

The error being: Array to string conversion in Configuration.php on line 136

Anyway, this is now solved by removing this copy, that shouldn't have been there in the first place.

PR #3708, opened 2023-10-12
Crontab: Rewrite the crontab recipe to work with sections

ardentsword:
- [ ] Bug fix #…?
- [x] New feature? 
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

Rewrite of the crontab contrib recipe. It creates a 'section' in the crontab without modifying the original crontab (before the whole crontab was wiped). If the cronjobs are modified it only touches the section. This removes the need to remove old 'unused' cronjobs when changing the command.

### DIFF
--- a/contrib/crontab.php
+++ b/contrib/crontab.php
@@ -2,7 +2,8 @@
 /*
 Recipe for adding crontab jobs.
 
-It checks for duplicates by the command part of the job. Changing the schedule will update the crontab. So when you change the command part you have to manually remove the old one. Use `crontab -e` on the server to remove it.
+This recipe creates a new section in the crontab file with the configured jobs.
+The section is identified by the *crontab:identifier* variable, by default the application name.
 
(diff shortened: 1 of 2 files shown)


antonmedv:
Is it a breaking change?

ardentsword:
It is not breaking for the config, but it is indeed breaking in that it duplicates the rules if it is already there.
I could write some code to check for duplicates and remove those 'outside' of the section, would that be the preferred solution?

antonmedv:
Yes, lets try this.

ardentsword:
I added the new code, let me know what you think of it :)

SimJoSt:
@ardentsword thank you for the rewrite. It makes working with crontab much smoother and structured. Great work!

ardentsword:
@SimJoSt Thanks, you're welcome, glad to hear it is getting used :)

PR #3717, opened 2023-10-27
Fix test function

nlemoine:
Prevents line breaks that might be outputed in `run` return to break condition.

I had a case with WP CLI (using [remote SSH features](https://make.wordpress.org/cli/handbook/guides/running-commands-remotely/#using-before_ssh-hook-on-client-machine)) where the output of run contains a line break that makes the test command fail whereas it shouldn't have.

Basically, instead of getting `+accurate`, I got `\n+accurate` which [makes the test](https://github.com/deployphp/deployer/blob/6611816c89524433bd5eb592457a62c3c8ee96f5/src/functions.php#L480) evaluate to `false`.

- [x] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?


nlemoine:
Thanks you @antonmedv ! I think `runLocally` should also be fixed the same way.

PR #3725, opened 2023-11-10
Add no_throw check to CLI running

allanmfx:
The `no_throw` option is ignored in the CLI process running.

- [x] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [x] Docs added?

      $ php bin/docgen

### DIFF
--- a/src/Component/ProcessRunner/ProcessRunner.php
+++ b/src/Component/ProcessRunner/ProcessRunner.php
@@ -73,13 +73,15 @@ class ProcessRunner
             $process->mustRun($callback);
             return $process->getOutput();
         } catch (ProcessFailedException $exception) {
-            throw new RunException(
-                $host,
-                $command,
-                $process->getExitCode(),
-                $process->getOutput(),
-                $process->getErrorOutput()
-            );
+            if (!$config['no_throw']) {
+                throw new RunException(
+                    $host,
+                    $command,
+                    $process->getExitCode(),
+                    $process->getOutput(),
+                    $process->getErrorOutput()
+                );
+            }
         } catch (ProcessTimedOutException $exception) { // @phpstan-ignore-line can be thrown but is absent from the phpdoc
             throw new TimeoutException(
                 $command,


src/Component/ProcessRunner/ProcessRunner.php
```diff
@@ -73,13 +73,15 @@ public function run(Host $host, string $command, array $config = []): string
...
             return $process->getOutput();
         } catch (ProcessFailedException $exception) {
-            throw new RunException(
-                $host,
-                $command,
-                $process->getExitCode(),
-                $process->getOutput(),
-                $process->getErrorOutput()
-            );
+            if (!$config['no_throw']) {
```
antonmedv, line 76: Not sure what is it for/

antonmedv (changes requested):

src/Component/ProcessRunner/ProcessRunner.php
allanmfx: @antonmedv During a code deploy workflow using deployer sometimes we don't want a certain command to throw an error and stop the whole deploy if some error happens, there is a no_throw config on deployer but currently is only working for the SSH process runner:

image

See \Deployer\Component\Ssh\Client class.

And then on my deployer script I can run something like that:

`run('varnishadm vcl.discard magento2', no_throw: true);`

allanmfx: The no_throw config is passed through the run function and sent along to the process runner:

image

(\Deployer\run) function

antonmedv: What about wrapping a run call in try-catch?

allanmfx: @antonmedv Not a good idea to encapsulate in a try-catch as I will need to introduce logic to a step-by-step script. The best is to have consistency across process runners regarding available parameters.

Deployer supports it for SSH, so it must also support the same parameter in the CLI runner.

allanmfx: @antonmedv Deployer allows to deploy either to an SSH server (using the SSH process runner) or a local machine (using CLI runner, the machine of course must have deployer installed).

For example: I've a **Staging** environment, and I can deploy from my machine to this environment. Deployer will connect to this server (SSH) and deploy, the **no_throw** config works because it's supported by the SSH process runner.

But let's imagine I've deployer installed inside the **Staging** environment and my code deployment workflow connects to this server and there runs the deployer, it's like running deployer locally in a machine, and for this case, deployer uses the CLI runner, not the SSH and the **no_throw** config will not work.

PR #3742, opened 2023-12-06
fix: run cachetool without args

### LINKED ISSUE #3741: Breaking change in v7.3.2 & v7.3.3 for `cachetool` task
- Deployer version: 7.3.2, 7.3.3
- Deployment OS: Debian 11

Hi, we recently updated from 7.3.1 to 7.3.3 and found that our very "default" cachetool usage stopped working. (The only cachetool config we have set is `set('bin/cachetool', 'vendor/bin/cachetool')` so use the composer installed version.) The issue was introduced by #3684 – where is caused a loud failure – and then tweaked by cb28eb82ff7549c888127ae29cb49024d2f2ca00 – where it now causes a silent failure.

The issue is that the `cachetool` setting is `''` (empty string) by default (see https://github.com/deployphp/deployer/blob/master/contrib/cachetool.php#L50) but the new changes in `cachetool_options` (see diff of 7.3.1...master at https://github.com/deployphp/deployer/compare/v7.3.1...master#diff-3ff1b6689a573050ee941e9cf0474c9f96a4ff7697a386d9c1af6b320fc10ce6) mean that this default value leads to *no* invocations of `cachetool`:

1. `cachetool` starts as `''`
2. it is then cast to an array, turning into `['']`
3. the empty element is skipped (by the `foreach ... if`)
4. an empty array is returned from `cachetool_options`
5. task `cachetool:clear:opcache` iterates over this array, doing nothing because it's empty

Based on the code, it seems that this will affect any of the `cachetool:*` tasks, as they all iterate over `cachetool_options`, but I have only experienced it with `cachetool:clear:opcache`

**Workaround**
I have found that `set('cachetool_args', ' ');` is an effective workaround for the time being. **Note** that it needs the second arg to be a non-empty string of only whitespace.

**Possible Fixes**
I'm sorry, but I don't have the time or wherewithal to test and submit a fix for this at this time. The options that jump out at me, though, are:
1. don't cast `(array)get('cachetool');`, and do `if ($options === '') { return [''] }`, then cast `(array) $options` when it's given to the foreach
2. leave the cast in place, but change the `foreach` to something like `$return[] = empty($option) ? '' : "--fcgi={$option}"`

Thank you!

infabo:
fixes a regression bug introduced with https://github.com/deployphp/deployer/commit/266554d3c7fea23c27ece48e3256d0120f88b138

- [x] Bug fix #3741
- [x] New feature? No
- [x] BC breaks? No, restores backward compatibility
- [ ] Tests added?
- [x] Docs added?

      $ php bin/docgen

### DIFF
--- a/contrib/cachetool.php
+++ b/contrib/cachetool.php
@@ -3,14 +3,14 @@
 
 ## Configuration
 
-- **cachetool** *(optional)*: accepts a *string* or an *array* of strings with the unix socket or ip address to php-fpm. If `cachetool` is not given, then the application will look for a `cachetool.yml` file and read the configuration from there.
+- **cachetool** *(optional)*: accepts a *string* or an *array* of strings with the unix socket or ip address to php-fpm. If `cachetool` is not given, then the application will look for a configuration file. The file must be named .cachetool.yml or .cachetool.yaml. CacheTool will look for this file on the current directory and in any parent directory until it finds one. If the paths above fail it will try to load /etc/cachetool.yml or /etc/cachetool.yaml configuration file.
 
     ```php
     set('cachetool', '/var/run/php-fpm.sock');
     // or
     set('cachetool', '127.0.0.1:9000');
     // or
-    set('cachetool', ['/var/run/php-fpm.sock', '/var/run/php-fpm-other.sock]);
+    set('cachetool', ['/var/run/php-fpm.sock', '/var/run/php-fpm-other.sock']);
     ```
 
 You can also specify different cachetool settings for each host:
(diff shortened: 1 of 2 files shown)


antonmedv:
@claytonrcarter please rake a look.

contrib/cachetool.php
```diff
@@ -72,13 +72,13 @@
...
         foreach ($options as $option) {
-            if ($option !== '') {
+            if (is_string($option) && $option !== '') {
                 $return[] = "--fcgi={$option}";
             }
         }
     }
 
-    return $return;
+    return $return ?: [''];
```
claytonrcarter, line 81: I had to double check, but `[] ?: ['']` returns `['']`, so this makes sense and it should resolve the issue I'm seeing in 7.3.3.

Thank you! 👍

claytonrcarter (approved):
This should do the trick; thank you for the prompt resolution!

Issue #3667, opened 2023-08-10
Interactive command output not shown


ochorocho:
- Deployer version: 7.3.1

- Deployment OS: php:8.1-cli-alpine (Docker)

I have the following task. This task will simply run `./vendor/bin/typo3 setup` on the remote server using symfony/process:
```
task('typo3:setup-if-needed', function () {
    cd('{{release_path}}');
    $decorator = output()->isDecorated();
    set('decorator', false);
    $remote = get('remote_user') . '@' . get('hostname');
    $binary = get('release_path') . '/vendor/bin/typo3';
    $process = new Process(['ssh', '-t', $remote,  $binary, 'setup']);
    $process->setTty(true);
    $process->run();
    set('decorator', $decorator);
});

```

This command is interactive, so user input ist required.

I tried using the `run('{{bin/typo3}} setup');` and set `ssh_arguments: ['-tt']` function provided by deployer, but this leaves the cli stuck without showing the interactive questions:
<img width="276" alt="Bildschirmfoto 2023-08-10 um 09 15 09" src="https://github.com/deployphp/deployer/assets/4623070/48eee91c-6815-4aba-b440-6bd12c34a546">

Without the `ssh_arguments: ['-tt']` the task fails:

```
[staging] ssh '-A' '-o' 'ControlMaster=auto' '-o' 'ControlPersist=60' '-o' 'ControlPath=~/.ssh/redittemplate@template.knallimall.org' 'redittemplate@template.knallimall.org' ': a9ba5c417eb370a0e4ae; bash -ls'
[staging] run cd /var/www/clients/client1/web89/home/redittemplate/web/typo3/releases/8 && (vendor/bin/typo3 setup)
[staging] Configuration already exists do you want to overwrite it [default: no] ?
[staging]  error  in deploy.php on line 101:
[staging] exit code 1 (General error)
done typo3:setup-if-needed 12s 510ms
```

Is there a way to run interactive commands on the server?

antonmedv:
Right now it is not possible. As each task is executed in a separate process.

PR #3754, opened 2023-12-17
[TASK] Add TYPO3 rsync recipe

ochorocho:
- [ ] Bug fix #…?
- [x] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [x] Docs added?

Origin MR:
https://github.com/deployphp/deployer/issues/3667


ochorocho:
I've added docs, but don't know why the "doc / docgen" fails as i do not have modified files locally after running `php  bin/docgen`

PR #3763, opened 2024-01-11
Switch to owner change instead of shh user change

MartijnCuppens:
Fixes https://github.com/deployphp/deployer/issues/3542


alanzheng88:
hi @MartijnCuppens, this assumes that we're able to log in as deployer user via ssh to deploy. if we need to log in as a different user to deploy, how should we go about doing this?

PR #3765, opened 2024-01-12
Move Caddy file in to separate file

MartijnCuppens:
- Move `Caddyfile` in to separate file to increase readability.
- Make use of `{{vars}}` instead of `$vars` (`realpath` had to be set for `deploy_path`)

This might be the first step for a better configurable `Caddyfile` in the future.

### DIFF
--- a/recipe/provision/Caddyfile
+++ b/recipe/provision/Caddyfile
@@ -1,0 +1,22 @@
+{{domain}} {
+
+	root * {{deploy_path}}/current/{{public_path}}
+	file_server
+	php_fastcgi * unix//run/php/php{{php_version}}-fpm.sock {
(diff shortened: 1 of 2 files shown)


SimJoSt:
Since this change, I am running into the following error when provisioning a new website:
```
ErrorException  in website.php on line 24:
    file_get_contents(./Caddyfile): Failed to open stream: No such file or directory
```
Did the change come with additional requirements that need to be fulfilled so the new system works?
If not, I'll create a proper bug report issue.
Downgrading to v7.3.3 resolved the problem.

SimJoSt:
Migrated to https://github.com/deployphp/deployer/discussions/3901

PR #3783, opened 2024-02-26
Add release name to deploy:info

Propaganistas:
I always find myself adding a task to just show the release name.
Why not add it to the built-in `deploy:info` task?

### DIFF
--- a/recipe/deploy/info.php
+++ b/recipe/deploy/info.php
@@ -3,5 +3,5 @@ namespace Deployer;
 
 desc('Displays info about deployment');
 task('deploy:info', function () {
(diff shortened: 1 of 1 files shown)


Schrank:
Nice  idea! But stupid question: where is the release_name taken from? 😅

Propaganistas:
@Schrank 
https://github.com/deployphp/deployer/blob/51e8e82c0726b0386d2989c42df003899d35af8a/recipe/deploy/release.php#L9-L14

It just reads the (potentially) already existing releases log.

Propaganistas:
Added a failsafe for initial deployments

PR #3822, opened 2024-04-26
DD#0000: feat: Update cache prefix to use stage if availble for multi…

### LINKED ISSUE #3790: Magento 2 Recipe: Cache prefix generated different cache prefixes when deploying to multiple hosts
**Situation:**

We have Magento 2 projects which are load balanced between different servers. For instance 2 frontend web servers for visitors and a seperate backend service for the Magento backoffice. All servers point to the same Redis instance which is located on the database server. We're deploying with the configuration: 

`after:
    deploy:shared: magento:set_cache_prefix
    deploy:magento: magento:cleanup_cache_prefix`

**Actual result:**

The set cache prefix is different per host because it will take the alias and release number. So for one server it will be:

`production_web35_12_`

And on the other server the prefix is set to: 

`production_web40_9_`

**Issue:** 

The issue is that cache will not be cleared correctly. The prefix should be the same for all 3 webservers in order for 1 webserver to clear the cache for the other two as well. 

**Expected result:**

The generated cache key should be the same for all hosts within the same "stage" and release. So if we deploy to "stage" production to 3 webservers, the generated key should be the same for all 3 servers for Magento to work correctly. 

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/deployphp) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/deployphp/deployer/issues/3790">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/deployphp/deployer/issues/3790/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

valguss:
… server deploys

- [x] Bug fix #3790?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [x] Docs added?

      $ php bin/docgen

### DIFF
--- a/docs/recipe/magento2.md
+++ b/docs/recipe/magento2.md
@@ -33,7 +33,7 @@ The [deploy](#deploy) task of **Magento 2** consists of:
   * [deploy:setup](/docs/recipe/deploy/setup.md#deploysetup) – Prepares host for deploy
   * [deploy:lock](/docs/recipe/deploy/lock.md#deploylock) – Locks deploy
   * [deploy:release](/docs/recipe/deploy/release.md#deployrelease) – Prepares release
-  * [deploy:update_code](/docs/recipe/shopware.md#deployupdate_code) – 
+  * [deploy:update_code](/docs/recipe/deploy/update_code.md#deployupdate_code) – Updates code
   * [deploy:shared](/docs/recipe/deploy/shared.md#deployshared) – Creates symlinks for shared files and dirs
   * [deploy:writable](/docs/recipe/deploy/writable.md#deploywritable) – Makes writable dirs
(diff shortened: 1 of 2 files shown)


valguss:
Hi @antonmedv not sure what's going on here but the docgen think I need to make an update to `docs/magento2.md` but i've already done that locally. Is there any special config I need to apply?

Cheers

antonmedv:
@peterjaap

barryvdh:
Not sure if relevant for this PR, but there are restrictions to the prefix.

```
Invalid id or tag 'staging.mydomain.com_129_GLOBAL__DICONFIG' : must use only [a-zA-Z0-9_]
```

So probably good to add a regex to check the output?

PR #3828, opened 2024-05-10
M2 recipe : handle cron stop in try catch

gbobts:
This change handle the case when cron:remove can not be executed properly, for example when setup:upgrade or Magento 2 installation wasn't done properly.

Try/Catch prevents breaking of deployment (as cron stopping is nice to have, but not required).

### DIFF
--- a/recipe/magento2.php
+++ b/recipe/magento2.php
@@ -494,6 +494,34 @@ task('magento:cleanup_cache_prefix', function () {
     run('{{bin/symlink}} {{deploy_path}}/shared/' . ENV_CONFIG_FILE_PATH . ' {{release_path}}/' . ENV_CONFIG_FILE_PATH);
 });
 
+/**
+ * Remove cron from crontab and kill running cron jobs
+ * To use this feature, add the following to your deployer scripts:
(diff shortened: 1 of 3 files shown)


peterjaap:
I'm not entirely sure this should be in the default recipe at all. Installing the cron is a one-off task and shouldn't occur on every deployment.

recipe/magento2.php
```diff
@@ -494,6 +494,38 @@ function magentoDeployAssetsSplit(string $area)
...
+task('magento:cron:stop', function () {
+    if (has('previous_release')) {
+        try {
+            run('{{bin/php}} {{previous_release}}/{{magento_dir}}/bin/magento cron:remove');
+        } catch (RunException $e) {
+            // do nothing
+        }
+    }
+
+    run('pgrep -U "$(id -u)" -f "bin/magento +(cron:run|queue:consumers:start)" | xargs -r kill');
```
antonmedv, line 514: I don't like this approach. What if cron job does not handle SIGTERM?

antonmedv (changes requested):

PR #3835, opened 2024-05-19
Feature/UI enhancements

midhunmonachan:
- [ ] ~~**Bug fix:** No bugs were fixed in this release.~~

- [x] **New feature:** Add some UI enhancements for provision. 

  - mainly affects `recipe/provision`, `recipe/laravel`, `provision/php`
  - [`recipe/laravel`] add default laravel `public_path` config value
  - [`provision/php`] set default `php_version` from `composer.json` _(if available)_ in the ask prompt
  - [`recipe/provision`] remove unnecessary print lines for `provision:configure`
  - [`recipe/provision`] avoid prompt for database info if database type is `none`
  - [`recipe/provision`] add confirmation to continue provision in unsupported OS

- [ ] ~~**BC breaks:** No backward compatibility breaks in this release.~~

- [ ] ~~**Tests added:** No new tests were added in this release.~~

- [ ] ~~**Docs added:** No new documentation was added in this release.~~

### DIFF
--- a/recipe/provision.php
+++ b/recipe/provision.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace Deployer;
 
 require __DIR__ . '/provision/databases.php';
@@ -67,12 +68,6 @@ task('provision:configure', function () {
         'db_name',
         'db_password',
     ];
-    $code = "\n\n    host(<info>'{{alias}}'</info>)";
-    foreach ($params as $name) {
-        $code .= "\n        ->set(<info>'$name'</info>, <info>'…'</info>)";
-    }
-    $code .= ";\n\n";
-    writeln($code);
     foreach ($params as $name) {
         get($name);
     }
--- a/recipe/provision.php
+++ b/recipe/provision.php
@@ -64,12 +64,19 @@ task('provision:configure', function () {
         'public_path',
         'php_version',
         'db_type',
+    ];
+    $dbparams = [
         'db_user',
         'db_name',
         'db_password',
     ];
     foreach ($params as $name) {
         get($name);
+    }
+    if (get('db_type') !== 'none') {
+        foreach ($dbparams as $name) {
+            get($name);
+        }
     }
 });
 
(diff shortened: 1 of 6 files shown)


recipe/provision.php
```diff
@@ -63,19 +67,20 @@
...
         'php_version',
         'db_type',
+    ];
+    $dbparams = [
         'db_user',
         'db_name',
         'db_password',
     ];
-    $code = "\n\n    host(<info>'{{alias}}'</info>)";
-    foreach ($params as $name) {
```
antonmedv, line 71: Should we keep printing those code parts?

midhunmonachan: Looks like it was left over from development.

Always prints this:
``` bash

host('your_hostname_here')
 ->set('sudo_password', '…')
 ->set('domain', '…')
 ->set('public_path', '…')
 ->set('php_version', '…')
 ->set('db_type', '…')
 ->set('db_user', '…')
 ->set('db_name', '…')
 ->set('db_password', '…');

```

midhunmonachan: > Should we keep printing those code parts?

Hi @antonmedv, I just noticed this code was put back in the master. Just curious, what's the purpose of printing those in the output?

Thanks!

antonmedv: So users can cope pase those and put into deploy.php.

midhunmonachan: If that's the case, I want to point out couple things.

We are printing lines for only $params array now. I split the `$params` array to `$dbparams` to aid in asking database info only if the database type is not `none`

So, if you need all the parameters output like before, I would modify it to:


```php


    $params = [
        'sudo_password',
        'domain',
        'public_path',
        'php_version',
        'db_type',
    ];
    $dbparams = [
        'db_user',
        'db_name',
        'db_password',
    ];
    $code = "\n\n    host(<info>'{{alias}}'</info>)";
    foreach (array_merge($params, $dbparams) as $name) {
        $code .= "\n        ->set(<info>'$name'</info>, <info>'…'</info>)";
    }
    $code .= ";\n\n";
    writeln($code);
});


```

In this, the `array_merge` function will combine `$params` and `$dbparams` into a single array, which is then iterated over to generate the configuration code. This way, all parameters from both arrays will be included in the output.

---

Also, I think it would be helpful to include a message just before printing it. 

For example, "To avoid entering parameters during script execution, include the following config with values in your deploy.php"

---

Let me know what you think.

antonmedv: Good suggestions. PR?

PR #3838, opened 2024-05-21
[Magento] add task to set permissions on files and directories to prevent all access to everyone

lohoy:
When I create artifact and deliver the package on server, the zip artifact keep permissions of locale directory.

I add task to set good permissions before zip artifact to set 644 permissions on files (754 on executable files in bin directory) and 755 permissions on directories.


antonmedv:
I think this is not very common problem.

peterjaap:
Agreed

PierreGauthier:
@antonmedv I think you should reconsider this fix
Because the artifact file permission should not depend on the permissions of the system that builds the artifact.
Without this fix, if you build the artifact on a CI like github workflow or gitlab ci, you will deploy file with 777 permissions on the server, which is not very secure.

peterjaap:
@PierreGauthier the file permission should not depend on the permissions of the system that builds the artifact, that is true.

The file permission should also not depend on a task that is configured in said system.

The file permissions however *should* depend on its file permission configuration in the repository, which Git supports.

PR #3906, opened 2024-10-01
feat: add new `local_archive` update code strategy

calebdw:
- [ ] Bug fix #…?
- [x] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [x] Docs added?

Hello!

### Description
This adds a new `local_archive` update code strategy which creates a tar from the local repository that is then streamed to the remote server and unpacked. This removes the need for the remote server to fetch the changes from the remote repository.

### Motivation

We have a privately hosted Gitlab instance, and we deploy to cloud servers via our ci/cd pipelines. This has been working just fine in the past. However, we've recently encountered an issue where the external server is not able to connect via ssh to the internal Gitlab server---which means the `git remote update` fails and our deployments are failing. The root issue is non-trivial and our IT personnel has been working with Cisco techs for a couple of weeks now, but they still have not been able to fix the issue.

I temporarily fixed our deployments by overriding the `deploy:update_code` task to simply use the local repository in the pipeline which works perfectly:

```php
    $host = currentHost()->connectionString();
    $target = get('target');

    $targetWithDir = $target;
    if (!empty(get('sub_directory'))) {
        $targetWithDir .= ':{{sub_directory}}';
    }

    runLocally(<<<BASH
        git archive {$targetWithDir} | ssh {$host} "tar -x -f - -C {{release_path}} 2>&1"
        BASH);

    $rev = escapeshellarg(runLocally("git rev-list $target -1"));
    run("echo $rev > {{release_path}}/REVISION");
});
```

I figured it would be a good idea to contribute this upstream in case anyone else would like to use this strategy :)

Thanks!


calebdw:
@antonmedv, could you please explain why you didn't like this?

calebdw:
@antonmedv, I've rebased :+1:

antonmedv:
Hmm, please recreate PR. I can't reoopen in. 
image

PR #3913, opened 2024-10-11
Remove collapsed true from Gitlab CI output

peterjaap:
- [x] Bug fix #…?
The Gitlab CI output now shows this;

image

I then have to expand that task output, to see the second task;

image

..etc etc.

This is very annoying to reach the end with 30+ tasks.

Apparently the collapsed=true is not needed to make a section, which is by default collapsed and causes this behavior. Removing it should fix this issue.


antonmedv:
@peterjaap what docker image do you use for Deployer?

peterjaap:
@antonmedv `php:${PHPVERSION}-buster`, for example https://hub.docker.com/layers/library/php/8.2-buster/images/sha256-7a320abde776a92a14bf7ee1371ab7781da23f4ff72b3175f600429e2f1350b0?context=explore

antonmedv:
Working on an official Deployer docker image.

PR #3923, opened 2024-10-17
Fix: change slack success and fail messages

Aquive:
- [x] Bug fix (Slack Messages)
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

Slack message was as follows:

```
MyProject
Deploy to *HEAD* successful
```

Where something like below was intended.

```
MyProject
Deploy to *myproject.com* successful
```
I changed it to:
     
```
MyProject
Deploy of `HEAD` to *myproject.com* successful
```

Same for failed message.

### DIFF
--- a/contrib/slack.php
+++ b/contrib/slack.php
@@ -76,8 +76,8 @@ set('slack_title', function () {
 
 // Deploy message
 set('slack_text', '_{{user}}_ deploying `{{target}}` to *{{hostname}}*');
-set('slack_success_text', 'Deploy to *{{target}}* successful');
-set('slack_failure_text', 'Deploy to *{{target}}* failed');
+set('slack_success_text', 'Deploy of `{{target}}` to *{{hostname}}* successful');
+set('slack_failure_text', 'Deploy of `{{target}}` to *{{hostname}}* failed');
 set('slack_rollback_text', '_{{user}}_ rolled back changes on *{{target}}*');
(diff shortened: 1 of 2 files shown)


antonmedv:
This will work strangely on multiple hosts deploys. I think we need some sort of notion of "stages".

contrib/slack.php
```diff
@@ -76,8 +76,8 @@
 
 // Deploy message
 set('slack_text', '_{{user}}_ deploying `{{target}}` to *{{hostname}}*');
-set('slack_success_text', 'Deploy to *{{target}}* successful');
-set('slack_failure_text', 'Deploy to *{{target}}* failed');
+set('slack_success_text', 'Deploy of `{{target}}` to *{{hostname}}* successful');
```
antonmedv, line 79: Showing hostname is not the best practise. We ascully want to show something like Deploy prod. Or deploy staging.

antonmedv (changes requested):

contrib/slack.php
antonmedv: > Deploying to two host in one go?

Yes. 

OR just a host name with IP. And we better to use aliaas in this case.

antonmedv: I guess we need a special magic config to determine the value of deployed hosts.

PR #3856, opened 2024-06-30
[Symfony] Add dump-env task

richardhj:
Add task to run dump-env on deployment

- [ ] Bug fix #…?
- [x] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [x] Docs added?

      Please, regenerate docs by running next command:
      $ php bin/docgen

As recommended on https://symfony.com/doc/current/deployment.html#b-configure-your-environment-variables

Run `composer dump-env` on every deployment.

The `$APP_ENV` shall be automatically when using `set('dotenv', '{{deploy_path}}/shared/.env.local');` in the deploy.php file.

PR #3932, opened 2024-10-22
[Symfony] Rename deploy:env to deploy:env:optimize

richardhj:
due to a naming conflict

- [x] Bug fix https://github.com/deployphp/deployer/pull/3856#issuecomment-2429681879
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

      $ php bin/docgen

This PR renames a task thats name is already used in the deploy.php (even though the task name was not used Symfony recipes before).

### DIFF
--- a/recipe/symfony.php
+++ b/recipe/symfony.php
@@ -64,7 +64,7 @@ task('deploy:cache:clear', function () {
 });
 
 desc('Optimize environment variables');
-task('deploy:env', function () {
+task('deploy:env:optimize', function () {
     within('{{release_or_current_path}}', function () {
         run('{{bin/composer}} dump-env "${APP_ENV:-prod}"');
     });


recipe/symfony.php
```diff
@@ -64,7 +64,7 @@
 });
 
 desc('Optimize environment variables');
-task('deploy:env', function () {
+task('deploy:env:optimize', function () {
```
Seb33300, line 67: Why not using the same word?
```suggestion
task('deploy:env:dump', function () {
```

mrhackcz:
And what about running this task after `deploy:vendors` ? :)

richardhj:
> And what about running this task after `deploy:vendors` ? :)

It is not intended to add this command to the default task at all (IMHO, because it depends on symfony/flex).
The case it ran for you was because you have had it in the deploy task (due to the bad naming) :)

YetiCGN:
> It is not intended to add this command to the default task at all (IMHO, because it depends on symfony/flex). The case it ran for you was because you have had it in the deploy task (due to the bad naming) :)

Nah, it was added to the [`update:prepare` common workflow](https://github.com/deployphp/deployer/commit/9db20b3c0d5fd89dc1ca5b2f6bff2ef93ceb524d) which is [causing us problems](https://github.com/deployphp/deployer/discussions/3938).

antonmedv:
I will release a new version with a fix.

richardhj:
Thank you @antonmedv

PR #3952, opened 2024-11-07
Update cachetool to latest version

webflo:
Cachetool has newer release. This PR updates the version number. It would be also possible to use "https://github.com/gordalina/cachetool/releases/latest/download/cachetool.phar" for the latest version. See https://docs.github.com/en/repositories/releasing-projects-on-github/linking-to-releases


Schrank:
It would be awesome to have a check which PHP version runs and then install the right cachetool version. Hardcoding the newest version is great (and better then any old version), but I prefer a version which runs with the damn old PHP version I have to use.

No pressure @webflo but if you have some boredom 🙈 


Thanks for the PR! <3

PR #3979, opened 2024-12-13
Added: bitbucket pipelines example

james2doyle:
Just added some docs for bitbucket. The example is a stripped-down version of what I'm using for a Laravel deployment, so it should be verbose enough for most starting points.

- [ ] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [x] Docs added?

### DIFF
--- a/docs/ci-cd.md
+++ b/docs/ci-cd.md
@@ -76,6 +76,53 @@ deploy:
     - master
 ```
 
+## Bitbucket Pipelines
+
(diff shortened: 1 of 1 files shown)


antonmedv:
Nice! Thanks.

james2doyle:
Great! Just a heads up that the image used in the example is a 7.x image.
Is that alright?

antonmedv:
v7 is ok

PR #4016, opened 2025-02-10
Lint sources on PHP 7.x via Github Action

staabm:
adds a new Github Action workflow via lints against PHP 7.x to prevent problems like https://github.com/deployphp/deployer/discussions/3976 which got fixed in https://github.com/deployphp/deployer/commit/f1f7b6169a1b48d7f715a10e8eb8323703cb9ab7

the newly added check skips the following files, as these currently contain PHP8+ only syntax:
```
Script parallel-lint handling the __exec_command event returned with error code 1
Error: Parse error: syntax error, unexpected 'o755' (T_STRING), expecting ')' in src/Documentation/DocGen.php on line 311
Error: Parse error: syntax error, unexpected 'o755' (T_STRING) in src/Component/PharUpdate/Update.php on line 104
Error: Parse error: syntax error, unexpected ':', expecting ')' in tests/legacy/recipe/named_arguments.php on line 8
```

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,41 @@
+name: test
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+
+jobs:
+  unit:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        php-versions: [ '7.4' ]
+    steps:
+      - uses: actions/checkout@v4
+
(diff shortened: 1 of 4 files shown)


antonmedv:
Nice! Should we fix php8 file to be php7?

> tests/legacy/recipe/named_arguments.php
> src/Documentation/DocGen.php

Can be skipped. 

But `src/Component/PharUpdate/Update.php` seems important.

staabm:
sure - here you are

PR #4017, opened 2025-02-10
feat: add new `local_archive` update code strategy

calebdw:
- [ ] Bug fix #…?
- [x] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [x] Docs added?

Supersedes #3906

Hello!

### Description
This adds a new `local_archive` update code strategy which creates a tar from the local repository that is then streamed to the remote server and unpacked. This removes the need for the remote server to fetch the changes from the remote repository.

### Motivation

We have a privately hosted Gitlab instance, and we deploy to cloud servers via our ci/cd pipelines. This has been working just fine in the past. However, we've recently encountered an issue where the external server is not able to connect via ssh to the internal Gitlab server---which means the `git remote update` fails and our deployments are failing. The root issue is non-trivial and our IT personnel has been working with Cisco techs for a couple of weeks now, but they still have not been able to fix the issue.

I temporarily fixed our deployments by overriding the `deploy:update_code` task to simply use the local repository in the pipeline which works perfectly:

```php
    $host = currentHost()->connectionString();
    $target = get('target');

    $targetWithDir = $target;
    if (!empty(get('sub_directory'))) {
        $targetWithDir .= ':{{sub_directory}}';
    }

    runLocally(<<<BASH
        git archive {$targetWithDir} | ssh {$host} "tar -x -f - -C {{release_path}} 2>&1"
        BASH);

    $rev = escapeshellarg(runLocally("git rev-list $target -1"));
    run("echo $rev > {{release_path}}/REVISION");
});
```

I figured it would be a good idea to contribute this upstream in case anyone else would like to use this strategy :)

Thanks!


recipe/deploy/update_code.php
```diff
@@ -70,55 +71,68 @@
...
-        goto start;
-    }
-
-    run("$git remote update 2>&1", env: $env);
+    if ($strategy === 'local_archive') {
+        $host = currentHost()->connectionString();
 
+        // Copy to release_path.
+        runLocally(<<<BASH
+            git archive {$targetWithDir} | ssh {$host} "tar -x -f - -C {{release_path}} 2>&1"
```
antonmedv, line 87: Let's use deployer's upload function.

antonmedv (changes requested):

recipe/deploy/update_code.php
calebdw: Do you want to use the `upload` function for the tar archive or all the files? I was using `git archive` to respect any `export-ignore`s listed in the `.gitattributes` file.

Are you saying I should:
- locally create the archive
- use the `upload` function to upload the archive
- remotely untar the archive?

calebdw: I've updated

adpeyre:
Hi,
I'm also interested in something like this!
I deploy locally from a GitLab Runner installed on my target machine, so the code update consists of copying files to the release path (the repository is already cloned by GitLab).
Do you think your solution could work for me?

Currently, I have to use this temporary solution.

``` yaml
tasks:
  deploy:update_code_from_runner:
        -   select: 'from=runner'
        -   upload:
                src: '.'
                dest: '{{release_path}}'
        -   run: 'echo $(git rev-parse HEAD) > {{release_path}}/REVISION'

    # Déploiement classique si on n'est pas dans le contexte d'un runner
  deploy:update_code_from_git:
        -   select: 'from!=runner'
        -   run: 'deploy:update_code'

  deploy:prepare:
        - 'deploy:info'
        - 'deploy:setup'
        - 'deploy:lock'
        - 'deploy:release'
        - 'deploy:update_code_from_runner'
        - 'deploy:update_code_from_git'
        - 'deploy:shared'
        - 'deploy:writable'

```` yaml
hosts:
  production:
        <<: *production
        local: true
        labels: { from: runner }
````

PR #4019, opened 2025-02-11
Add array_is_list polyfill for backward compatibility

NickVulkers:
Add array_is_list polyfill as requested in https://github.com/deployphp/deployer/discussions/4007#discussioncomment-12145348.

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -980,16 +980,3 @@ function fetch(string $url, string $method = 'get', array $headers = [], ?string
     }
     return $http->send($info);
 }
-
-/**
- * Polyfill for the array_is_list function introduced in PHP 8.1.
- *
- * @param array $array The array to check.
- * @return bool Returns true if the array is a list, false otherwise.
- */
-if (!function_exists('array_is_list')) {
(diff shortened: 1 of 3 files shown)


src/functions.php
```diff
@@ -980,3 +980,16 @@ function fetch(string $url, string $method = 'get', array $headers = [], ?string
...
 }
+
+/**
+ * Polyfill for the array_is_list function introduced in PHP 8.1.
+ *
+ * @param array $array The array to check.
+ * @return bool Returns true if the array is a list, false otherwise.
+ */
+if (!function_exists('array_is_list')) {
+    function array_is_list(array $array): bool {
```
antonmedv, line 991: I think this will just add array_is_list to Deployer namespace.

antonmedv (changes requested):

src/functions.php
NickVulkers: My bad, should've looked a little better. I think this is the correct approach looking at your other code, I've now added it to `Deployer\Support\helpers` and used it in the magento2 recipe.

PR #4026, opened 2025-02-27
Remove duplicate Content-type header

webflo:
The Grafana plugin is broken, because the generated request has two http header (Content-**t**ype and Content-**T**ype). The later one, it the correct header and its added by ```\Deployer\Utility\Httpie::jsonBody```

- [x] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

      $ php bin/docgen

### DIFF
--- a/contrib/grafana.php
+++ b/contrib/grafana.php
@@ -61,7 +61,6 @@ task('grafana:annotation', function () {
 
     Httpie::post($config['url'])
         ->header('Authorization', 'Bearer ' . $config['token'])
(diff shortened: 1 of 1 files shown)


blankse:
@webflo I think the target branch should be `7.x` because it is a bug.

webflo:
https://github.com/deployphp/deployer/pull/4049 created for 7.x Thanks!

PR #4034, opened 2025-03-09
Fix performance of task "releases"

kszymukowicz:
Fix performance of task `releases` by checking `cat releases/$release/REVISION` only for existing releases and not all past releases stored in `releases_log`.

A real example data to compare results.
In `releases_log` there are 160 releases logs.   
In `releases_list` there are 5 releases folders.

Before fix: 57 seconds.
After fix: 7 seconds.

- [x] Bug fix
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

### DIFF
--- a/recipe/deploy/release.php
+++ b/recipe/deploy/release.php
@@ -179,14 +179,16 @@ task('releases', function () {
             } else {
                 $status = "<info>$release</info>";
             }
+            try {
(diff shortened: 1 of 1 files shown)


antonmedv (approved):

antonmedv:
Nope =/

PR #4041, opened 2025-03-24
Disable the http:cache:warm:up step for Shopware 6.6+

poespas:
We've seen a few complaints about how Shopware 6.6+ appeared to have dropped support for http:cache:warm:up. This is why we've added conditional logic to check the current shopware version, very similar to how Magento's version check works [as can be seen here](https://github.com/deployphp/deployer/blob/master/recipe/magento2.php#L111).

The change in Shopware was loosely mentioned here: https://github.com/shopware/shopware/blob/932d232e999f06cf340ed00708047b9721b253b0/UPGRADE-6.6.md?plain=1#L1035

- [x] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

      $ php bin/docgen

### DIFF
--- a/docs/recipe/shopware.md
+++ b/docs/recipe/shopware.md
@@ -166,11 +166,23 @@ Please note that the files in `config/jwt/*` receive special attention in the `s
 ```
 
 
+### shopware_version
+[Source](https://github.com/deployphp/deployer/blob/master/recipe/shopware.php#L75)
+
+This sets the shopware version to the version of the shopware console command.
+
+```php title="Default value"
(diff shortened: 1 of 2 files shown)


antonmedv (approved):

phizab:
Hi @antonmedv,

would be nice to have this change in version 7, because version 8 is still in alpha. Are there any plans for version 7.5.13?

antonmedv:
Not right now. I'm planning to move with v8 release soon.

PR #4046, opened 2025-03-31
Fix #4036: Escape shell argument when setting remote crontab (7.x)

silentvick:
- [x] Bug fix #4036 
- [ ] New feature
- [ ] BC breaks
- [ ] Tests added
- [x] Docs added

      $ php bin/docgen


antonmedv:
Releasing v7 kinda complex)

PR #4055, opened 2025-04-06
Php composer parsing

Max13:
`composer.json` can contain: `"php": "^8.2.0"`, in this case `php8.2.0` will be used for provisioning, and `php8.2.0-*` packages don't exist.

Also added `8.3` in the choices (as it was already the default)


Max13:
@antonmedv Would it be agreed if I PR this same code (and #4054) to 7.x as well ?

antonmedv:
Nope. All new features go only to v8.

antonmedv:
I think so. Releaseing v7 is difficult compared to v8.

Max13:
I thought that contribs, as being single files, could be integrated in v7 and v8. No problem

PR #4059, opened 2025-04-22
Shopware recipe: Share theme folder for atomic themes

M-arcus:
Shopware has decided to make their theme compilation atomic: https://developer.shopware.com/docs/resources/references/adr/2023-01-10-atomic-theme-compilation.html

This means, on each deployment, during the execution of `theme:compile` (when building remotely, code is [here](https://github.com/deployphp/deployer/blob/master/recipe/shopware.php#L114-L118)), the storefront page is broken, because the new theme files only exists in an unreleased release, until `deploy:publish` happens.

### DIFF
--- a/docs/recipe/shopware.md
+++ b/docs/recipe/shopware.md
@@ -134,6 +134,7 @@ These directories are shared among all releases.
     'files',
     'var/log',
     'public/media',
+    'public/theme',
     'public/thumbnail',
     'public/sitemap',
 ]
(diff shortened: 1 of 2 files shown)


Schrank:
Thank you @M-arcus!

M-arcus:
@Schrank 
I have checked the code again, in this commit https://github.com/deployphp/deployer/commit/7f905c21558f5d9e20e27c38999b48f7ea2a7855 the recipe was changed to local compilation of the theme. This means, the compiled theme would get removed and the theme folder symlinked to the shared folder.

I think we need to revert the PR.

PR #4060, opened 2025-04-22
Revert "Shopware recipe: Share theme folder for atomic themes"

Schrank:
Reverts deployphp/deployer#4059

@M-arcus Thanks for telling me, looked good from my point of view :-)


M-arcus:
@Schrank 
Apologies. I only checked for the existence of the remotely executed theme:compile task, but not if it actually got executed, which previously it was, but 2 years ago :D

Issue #2802, opened 2021-11-30
Git subodules not working with `clone` strategy


matsn0w:
Hey there,

Previously, there was a `git_recursive` option available in `deploy:update_code`, but that was [removed in March](https://github.com/deployphp/deployer/commit/a200c1f9b0159c30ae846c28b84b9a306f611fd7#diff-603dddef04f285d9a2ff024cb7bbe3d8c242011d0f7bca49a00446d769033325) :(

Why is this removed? Can it be restored?

Thanks in advance!

PS: some additional background info, I'm building a Laravel application using [Laravel Modules](https://github.com/nWidart/laravel-modules) where every module has its own repo. These need to be pulled upon deployment.. 

When using the `update_code_strategy` option `clone`, the `.git` directory is preserved. But when I then manually run `git submodule update` in a different task, I'm getting an error:

```
[production] run cd ***/releases/1 && (
    # clone the submodules
    git submodule init
    git submodule update
)
[production] Submodule 'Modules/[MODULE-1]' (***/.dep/[MODULE-1].git) registered for path 'Modules/[MODULE-1]'
[production] Submodule 'Modules/[MODULE-2]' (***/.dep/[MODULE-2].git) registered for path 'Modules/[MODULE-2]'
[production] fatal: repository '***/.dep/[MODULE-1].git' does not exist
[production] fatal: clone of '***/.dep/[MODULE-1].git' into submodule path '***/releases/1/Modules/[MODULE-1]' failed
[production] Failed to clone 'Modules/[MODULE-1]'. Retry scheduled
[production] fatal: repository '***/.dep/[MODULE-2].git' does not exist
[production] fatal: clone of '***/.dep/[MODULE-2].git' into submodule path '***/releases/1/Modules/[MODULE-2]' failed
[production] Failed to clone 'Modules/[MODULE-2]'. Retry scheduled
[production] fatal: repository '***/.dep/[MODULE-1].git' does not exist
[production] fatal: clone of '***/.dep/[MODULE-1].git' into submodule path '***/releases/1/Modules/[MODULE-1]' failed
[production] Failed to clone 'Modules/[MODULE-1]' a second time, aborting
[production] fatal: repository '***/.dep/[MODULE-2].git' does not exist
[production] fatal: clone of '***/.dep/[MODULE-2].git' into submodule path '***/releases/1/Modules/[MODULE-2]' failed
[production] Failed to clone 'Modules/[MODULE-2]' a second time, aborting
[...]

PR #4063, opened 2025-04-29
`update_code` recipe - 3rd option `update_code_strategy` that loads submodule

alxndr-w:
This adds the ability to use submodules with `update_code` set to `clonse_submodule` while staying backwards compatible.

Issue: <https://github.com/deployphp/deployer/issues/2802>

[Using submodules were possible in Deployer 6 but not 7](https://github.com/deployphp/deployer/issues/148#issuecomment-2047203925) and could be reintroduced like this.

- [ ] Bug fix 
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

> Please, regenerate docs by running next command:
> $ php bin/docgen

### DIFF
--- a/recipe/deploy/update_code.php
+++ b/recipe/deploy/update_code.php
@@ -45,6 +45,7 @@ set('target', function () {
 // Can be one of:
 // - archive
 // - clone (if you need the origin repository `.git` dir in your {{release_path}})
+// - close_submodule (same as `clone`, but running `git submodule init` and `git submodule update` as well)
 set('update_code_strategy', 'archive');
 
 // Sets environment variable _GIT_SSH_COMMAND_ for `git clone` command.
@@ -114,6 +115,13 @@ task('deploy:update_code', function () {
         run("$git clone -l $bare .");
         run("$git remote set-url origin $repository", env: $env);
         run("$git checkout --force $target");
+    } elseif (get('update_code_strategy') === 'clone_submodule') {
+        cd('{{release_path}}');
+        run("$git clone -l $bare .");
+        run("$git remote set-url origin $repository", env: $env);
+        run("$git checkout --force $target");
+        run("$git submodule init");
+        run("$git submodule update");
     } else {
         throw new ConfigurationException(parse("Unknown `update_code_strategy` option: {{update_code_strategy}}."));
     }


recipe/deploy/update_code.php
```diff
@@ -45,6 +45,7 @@
 // Can be one of:
 // - archive
 // - clone (if you need the origin repository `.git` dir in your {{release_path}})
+// - close_submodule (same as `clone`, but running `git submodule init` and `git submodule update` as well)
```
cafferata, line 48: Typo:

```suggestion
// - clone_submodule (same as `clone`, but running `git submodule init` and `git submodule update` as well)
```

cafferata (changes requested):

recipe/deploy/update_code.php
alxndr-w: Thanks!

recipe/deploy/update_code.php
```diff
@@ -114,6 +115,13 @@
         run("$git clone -l $bare .");
         run("$git remote set-url origin $repository", env: $env);
         run("$git checkout --force $target");
+    } elseif (get('update_code_strategy') === 'clone_submodule') {
+        cd('{{release_path}}');
+        run("$git clone -l $bare .");
+        run("$git remote set-url origin $repository", env: $env);
+        run("$git checkout --force $target");
+        run("$git submodule init");
```
antonmedv, line 123: Maybe it's better to define it as a separate task which can be attached if needed.

alxndr-w: I'm sorry, I'm not much into Deployer. I don't know how to solve this differently. Also, a separate task just for basically one line would be more complex.

antonmedv:
```php
task('deploy:git:submodules', function () {
    run('git submodule init');
    run('git submodule update');
});
```

antonmedv:
```php

host('production')
    ->set('needs_submodules', true);

after('deploy:update_code', function () {
    if (get('needs_submodules', false)) {
        invoke('deploy:git:submodules');
    }
});
```

PR #4066, opened 2025-05-09
Feature - Laravel - Customizable artisan path

Skullsneeze:
- [x] New feature
- [x] Docs added

**Problem**
With certain setups the default path for artisan does not resolve to the correct directory (eg. monorepo with laravel + another application).

**Solution**
This PR adds the possibility to set a custom path to `artisan`, with a fallback to the default path if no custom path is defined.

**Notes**
This PR is somewhat similar in nature to https://github.com/deployphp/deployer/pull/3772 but aims to provide an easier way of adjusting the path

### DIFF
--- a/recipe/laravel.php
+++ b/recipe/laravel.php
@@ -21,8 +21,9 @@ set('writable_dirs', [
     'storage/logs',
 ]);
 set('log_files', 'storage/logs/*.log');
+set('bin/artisan', '{{release_or_current_path}}/artisan');
 set('laravel_version', function () {
-    $result = run('{{bin/php}} {{release_or_current_path}}/artisan --version');
+    $result = run("{{bin/php}} {{bin/artisan}} --version");
     preg_match_all('/(\d+\.?)+/', $result, $matches);
     return $matches[0][0] ?? 5.5;
 });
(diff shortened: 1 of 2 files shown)


recipe/laravel.php
```diff
@@ -79,6 +80,15 @@ function artisan($command, $options = [])
     };
 }
 
+function artisan_path()
```
antonmedv, line 83: Lets just define a new config option: `{{bin/artisan}}`.

recipe/laravel.php
```diff
@@ -22,7 +22,8 @@
 ]);
 set('log_files', 'storage/logs/*.log');
 set('laravel_version', function () {
-    $result = run('{{bin/php}} {{release_or_current_path}}/artisan --version');
+    $artisan = artisan_path();
+    $result = run("{{bin/php}} $artisan --version");
```
antonmedv, line 26: We can use bin/artisan here:

```php
$result = run("{{bin/artisan}} --version");
```

antonmedv (changes requested):

recipe/laravel.php
Skullsneeze: Yeah I suppose that would make more sense! Will push some changes soon

Skullsneeze:
Thanks @antonmedv, I agree this is a cleaner solution! I've updated the PR and squashed the commits

antonmedv (approved):

PR #4082, opened 2025-06-13
Bugfix dotenv test in laravel.php

aladei-dean-lucassen:
The test to check if .env file is present was not checking if the dotenv variable was set and always used the default. Added getter for dotenv variable.

### DIFF
--- a/recipe/laravel.php
+++ b/recipe/laravel.php
@@ -58,12 +58,15 @@ function artisan($command, $options = [])
             return;
         }
 
+        // Get the dotenv path or use default.
+        $dotenv = get('dotenv', '{{release_or_current_path}}/.env');
+
         // Ensure we warn or fail when a command relies on the ".env" file.
-        if (in_array('failIfNoEnv', $options) && !test('[ -s {{release_or_current_path}}/.env ]')) {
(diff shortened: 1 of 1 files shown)


antonmedv:
I think we also need to change dotenv itself, so it works during deploy:

```diff
-set('dotenv', '{{current_path}}/.env');
+set('dotenv', '{{release_or_current_path}}/.env');
```

aladei-dean-lucassen:
@antonmedv where should this be added? I cannot find any occurrence of the `set('dotenv', '{{current_path}}/.env');` other than a comment on line 78 in the common recipe.

aladei-dean-lucassen:
@antonmedv I'm happy to help but could you point me in the right direction? As stated in my previous comment I cannot find the line of code you shared besides a comment.

> @antonmedv where should this be added? I cannot find any occurrence of the `set('dotenv', '{{current_path}}/.env');` other than a comment on line 78 in the common recipe.

antonmedv:
This code: https://deployer.org/docs/8.x/recipe/common#dotenv

aladei-dean-lucassen:
@antonmedv just added the commit.

antonmedv (approved):

PR #4091, opened 2025-06-30
Fix command options mode (InputOption::VALUE_REQUIRED)

gharlan:
- [x] Bug fix #…?
- [ ] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

When using these options, a value is required.

### DIFF
--- a/src/Command/ConfigCommand.php
+++ b/src/Command/ConfigCommand.php
@@ -30,7 +30,7 @@ class ConfigCommand extends SelectCommand
     protected function configure()
     {
         parent::configure();
-        $this->addOption('format', null, InputOption::VALUE_OPTIONAL, 'The output format (json, yaml)', 'yaml');
+        $this->addOption('format', null, InputOption::VALUE_REQUIRED, 'The output format (json, yaml)', 'yaml');
         $this->getDefinition()->getArgument('selector')->setDefault(['all']);
     }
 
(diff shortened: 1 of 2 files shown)


src/Command/InitCommand.php
```diff
@@ -28,7 +28,7 @@ protected function configure()
         $this
             ->setName('init')
             ->setDescription('Initialize deployer in your project')
-            ->addOption('path', 'p', InputOption::VALUE_OPTIONAL, 'Recipe path');
+            ->addOption('path', 'p', InputOption::VALUE_REQUIRED, 'Recipe path');
```
antonmedv, line 31: Well, this is optional!

src/Command/ConfigCommand.php
```diff
@@ -30,7 +30,7 @@ public function __construct(Deployer $deployer)
     protected function configure()
     {
         parent::configure();
-        $this->addOption('format', null, InputOption::VALUE_OPTIONAL, 'The output format (json, yaml)', 'yaml');
+        $this->addOption('format', null, InputOption::VALUE_REQUIRED, 'The output format (json, yaml)', 'yaml');
```
antonmedv, line 33: It is optional and yaml by default.

antonmedv (changes requested):

antonmedv (approved):

PR #4114, opened 2025-09-09
Skip releases marked with DIRTY_RELEASE

pableu:
A canceled deployment (network disconnection, command aborted with ctrl+c) can leave an incomplete state in `.../releases/<number>`. 

The next `dep rollback` will roll back to that dirty release. 

I found the `releases` task mentioning (and showing) releases as "dirty" by looking for the DIRTY_RELEASE file. So I though it might be a good start to support that also during rollback. 

We could create and remove the file in appropriate steps during the usual deployment tasks in a separate PR. But users could also implement that themselves if needed.


antonmedv:
What we need is a way to track "finished" releases.

pableu:
I'll make a new PR.

PR #4115, opened 2025-09-16
Mark finished release and don't consider un-finished releases in rollback

pableu:
A better solution, as figured out in #4114. 

We want to track successfully finished releases, and only roll back to these in `dep rollback`. 

Otherwise, we might roll back to a semi-released deployment, e.g., one we aborted with CTRL+C or where something else went wrong.


antonmedv (approved):

pableu:
Alrighty, done @antonmedv.

PR #4118, opened 2025-09-25
Feature: Reset PHP OPcache

MarcoRemy:
- [x] New feature
- [ ] Docs added // TODO

Hello there,

This PR is a proposal. It addresses the PHP OPcache reset after a new release is published.

The idea was inspired by the [TYPO3/Surf](https://github.com/TYPO3/Surf) deployment project. It features two [tasks](https://github.com/TYPO3/Surf/tree/main/src/Task/Php) which create and execute a simple PHP script to reset the OPcache. As the OPcache memory space of the web server differs from that of the CLI interpreter, the script must be executed by the web server. To achieve this, the PHP script is placed in the document root of the release and fetched from the web server after the release is published.

```php
// TYPO3/Surf workflow
->beforeStage('transfer', \TYPO3\Surf\Task\Php\WebOpcacheResetCreateScriptTask::class, $application)
->afterStage('switch', \TYPO3\Surf\Task\Php\WebOpcacheResetExecuteTask::class, $application);
```

For Deployer, the mechanics were adapted into an recipe which consists of two grouped task

```php
desc('Reset PHP OPcache');
task('deploy:reset_opcache', [
    'deploy:reset_opcache:prepare',
    'deploy:reset_opcache:execute',
]);
```

These tasks have not yet been included in any deployment recipe because they are still in the experimental stage.

There are several considerations to take into account:
* The script must be placed in the public document root directory. This may not be the same as the release directory. Application recipes need to set `reset_opcache_document_root` to their specific path.
* The document root should be writable for www user
* In order to request the script from the web server, the URL must be set by `reset_opcache_app_url`. This configuration option likely needs improvement.
* The `fetch()` function does not offer any options for handling untrusted HTTPS connections in development environments. Direct use of `Httpie` would be an alternative approach to solve this issue.
* It works with Apache HTTP Server + PHP-FPM. However, I don't know how it applies to other web server software, e.g. nginx.

Please let me know what you think.

### DIFF
--- a/recipe/common.php
+++ b/recipe/common.php
@@ -12,6 +12,7 @@ require __DIR__ . '/deploy/info.php';
 require __DIR__ . '/deploy/lock.php';
 require __DIR__ . '/deploy/push.php';
 require __DIR__ . '/deploy/release.php';
+require __DIR__ . '/deploy/reset_opcache.php';
 require __DIR__ . '/deploy/rollback.php';
 require __DIR__ . '/deploy/setup.php';
 require __DIR__ . '/deploy/shared.php';
--- a/recipe/deploy/reset_opcache.php
+++ b/recipe/deploy/reset_opcache.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace Deployer;
+
+set('reset_opcache_nonce', bin2hex(random_bytes(8)));
+set('reset_opcache_document_root', '{{release_or_current_path}}');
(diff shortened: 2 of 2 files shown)


MarcoRemy:
Thank you for your feedback. I was unaware of the cachetool recipe. As it already handles the opcache, I am going to close the PR.

Schrank:
Sorry for the time invested, although I'm happy that it solves your problem! <3

PR #4190, opened 2026-04-12
Update docker image due to missing/depracated dependencies

featdd:
The current docker image fails because of the missing sockets extension for PHP.

For successfull build and execution "linux-headers" & "curl-dev" needed to be added.

I build a custom image for v8.0.0-rc, see [featdd/deployerphp:v8.0.0-rc](https://hub.docker.com/repository/docker/featdd/deployerphp/tags/v8.0.0-rc) and now it works fine for me.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,8 @@
 FROM php:8.4-cli-alpine
 
-RUN apk add --no-cache bash git openssh-client rsync zip unzip libzip-dev \
+RUN apk update && apk add --no-cache linux-headers bash git openssh-client rsync zip unzip libzip-dev curl-dev
 
-RUN docker-php-ext-install mbstring mcrypt pcntl sockets curl zip
+RUN docker-php-ext-install pcntl sockets curl zip
 
 COPY --chmod=755 deployer.phar /bin/dep
 


Dockerfile
```diff
@@ -1,8 +1,8 @@
-FROM php:8.4-cli-alpine
+FROM php:8.5-cli-alpine
```
antonmedv: Lets split to a separate PR.

Dockerfile
```diff
@@ -1,8 +1,8 @@
...
 
-RUN apk add --no-cache bash git openssh-client rsync zip unzip libzip-dev \
+RUN apk update && apk add --no-cache linux-headers bash git openssh-client rsync zip unzip libzip-dev curl-dev
 
-RUN docker-php-ext-install mbstring mcrypt pcntl sockets curl zip
```
antonmedv, line 5: why those are dropped?

antonmedv (changes requested):

antonmedv:
Why do we need "linux-headers" & "curl-dev"?

featdd:
Hi @antonmedv,

### linux-headers
The "linux-headers" is necessary to compile the socket extension for PHP, otherwise the docker build ends up with this error:
`cc -I. -I/usr/src/php/ext/sockets -I/usr/local/include/php -I/usr/local/include/php/main -I/usr/local/include/php/TSRM -I/usr/local/include/php/Zend -I/usr/local/include/php/ext -I/usr/local/include/php/ext/date/lib -fstack-protector-strong -fpic -fpie -O2 -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -DHAVE_CONFIG_H -fstack-protector-strong -fpic -fpie -O2 -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -D_GNU_SOURCE -DZEND_ENABLE_STATIC_TSRMLS_CACHE=1 -DZEND_COMPILE_DL_EXT=1 -c /usr/src/php/ext/sockets/sockets.c -MMD -MF sockets.dep -MT sockets.lo  -fPIC -DPIC -o .libs/sockets.o

/usr/src/php/ext/sockets/sockets.c:58:12: fatal error: linux/sock_diag.h: No such file or directory`

The file necessary is part of the linux-header package, see: [linux/sock_diag.h](https://github.com/torvalds/linux/blob/master/include/linux/sock_diag.h)

### curl-dev
Similiar issue happens when you try to compile the curl PHP extension (`Package 'libcurl' not found`), it requires libcurl which is installed through "curl-dev" (It's kind of an equivalent for alpine linux)

### mbstring
This can be dropped as the extension is [enabled by default in the base image already](https://github.com/docker-library/php/blob/master/8.4/alpine3.23/cli/Dockerfile#L136)

### mcrypt
The extension is deprecated and shouldn't be used anymore, it fails to be installed with pecl or docker-php-ext-install, by searching the code I didn't see a hard dependency on this anyway or did I oversee something here?
Error: `error: /usr/src/php/ext/mcrypt does not exist`

### PHP 8.5
Should I just drop/amend this from this commit? (It works fine in the image I built)

Greetings
Daniel

Dockerfile
featdd: see my reply here: https://github.com/deployphp/deployer/pull/4190#issuecomment-4231824858

Dockerfile
featdd: I removed this and amended this to the pull request commit

antonmedv (approved):

Dockerfile
```diff
@@ -1,8 +1,8 @@
 FROM php:8.4-cli-alpine
 
-RUN apk add --no-cache bash git openssh-client rsync zip unzip libzip-dev \
+RUN apk update && apk add --no-cache linux-headers bash git openssh-client rsync zip unzip libzip-dev curl-dev
 
-RUN docker-php-ext-install mbstring mcrypt pcntl sockets curl zip
+RUN docker-php-ext-install pcntl sockets curl zip
```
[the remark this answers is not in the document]
antonmedv: @featdd looks like --no-cache is enought. Could you please verify?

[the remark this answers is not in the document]
featdd: @antonmedv ah not only that, it seems curl is also already available by default.

I just let codex do some further optimizations and could save 13MB in the final image, see: [featdd/deployerphp](https://hub.docker.com/repository/docker/featdd/deployerphp/tags/v8.0.0-rc)

I'll create another pull request

PR #4199, opened 2026-04-26
Feat: add thin typed wrappers for set('path', ...) and get('path', ...)

m1rm:
- [ ] ~Bug fix #…?~
- [x] New feature?
- [ ] ~BC breaks?~
- [x] Tests added?
- [x] Docs added?

**Changes:**
This PR adds two thin typed wrappers ` setPath(string $path): self` and `getPath(): ?string`, which are wrapping `$this->config->set('path', $path);` and `$this->config->get('path', null);` respectively.

**Motivation:**
Improve IDE/static analysis support by providing a typed function to use.

### DIFF
--- a/docs/hosts.md
+++ b/docs/hosts.md
@@ -76,6 +76,16 @@ host('example.org')
     ->setRemoteUser('deployer');
 ```
 
+### Path
+
+You can store an arbitrary path string in the `path` config (for example to reuse in `{{path}}` placeholders). On a host, prefer `setPath()` over `set('path', …)` so the value is typed as a string and IDEs or static analysis can catch mistakes such as passing a number, which the generic `set()` method does not flag. In the recipe file (outside a host definition), use the global [set()](api.md#set) function—for example `set('path', '/var/www/myapp')`.
+
+```php
+host('example.org')->setPath('/var/www/myapp');
+
+set('path', '/var/www/myapp');
+```
+
 ---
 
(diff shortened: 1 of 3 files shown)


antonmedv:
I'm not really like the ability to have this path without explicitly explaining what it is. I think for type options we need to find another solution.

m1rm:
maybe I can improve my pull request? 

all I thought of was that `setPath` is not that much 'worse' than `set('path', ...)`, but maybe a better name for `setPath` (and `getPath`respectively) would already do the job?

antonmedv:
I just think the path is pretty generic and can confuse a lot of users of deployer because they're gonna see a path in autocomplete and gonna think what is this path refers to.

m1rm:
I absolutely agree :+1: , give me a day or two to come up with something better :slightly_smiling_face: I think the benefits are worth it

m1rm:
what do you think of "setPathConfig"/"getPathConfig" @antonmedv . Making it clearer that it is a configured setting and it is up to the user what to use it for? 

On second thought however I am not sure if we want to expose that at all/encourage the usage. I approached it solely from a "technical" point of view, which might have been not ideal.

I renamed the functions to the best name I could come up with.

antonmedv:
I don't think it's a solution create special helpful indicators for custom usage just to have better types. So I all four types I really like TypeScript and Haskell but this is not a solution.

PR #4201, opened 2026-04-29
feat: npm copy previous release

ALameLlama:
Yarn does this and it should improve performance

https://github.com/deployphp/deployer/blob/master/contrib/yarn.php#L24-L29

- [ ] Bug fix #…?
- [X] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?


antonmedv:
I do not think this is the right approach. 

I also think we should remove cp from yarn too!

Both yarn/npm have own cache.

PR #4222, opened 2026-05-21
PHPStan: cleanup setup

staabm:
cleaning up tooling setup to ignore less errors. preparing the way to raise to level 6 in a followup

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -9,9 +9,16 @@ parameters:
         - recipe
         - contrib
 
+    bootstrapFiles:
+        - tests/constants.php
+
+    dynamicConstantNames:
+        DEPLOYER_VERSION: string
+        DEPLOYER_BIN: string
+        MASTER_ENDPOINT: string
+        MASTER_TOKEN: string
+
     ignoreErrors:
-        - "#^Constant DEPLOYER_VERSION not found\\.$#"
-        - "#^Constant DEPLOYER_BIN not found\\.$#"
         - "#^Constant MASTER_ENDPOINT not found\\.$#"
         - "#^Constant MASTER_TOKEN not found\\.$#"
         - "#CpanelPhp#"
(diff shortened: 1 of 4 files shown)


phpstan.neon
```diff
@@ -9,9 +9,16 @@ parameters:
...
         - contrib
 
+    bootstrapFiles:
+        - tests/constants.php
+
+    dynamicConstantNames:
+        DEPLOYER_VERSION: string
+        DEPLOYER_BIN: string
+        MASTER_ENDPOINT: string
+        MASTER_TOKEN: string
```
staabm, line 19: make PHPStan aware which types we expect the constants to have. those will be validated against when code uses `define()` or `const`.

see
- https://phpstan.org/config-reference#constants
- https://x.com/markusstaab/status/2056439956986269742

phpstan.neon
```diff
@@ -9,9 +9,16 @@ parameters:
...
+
     ignoreErrors:
-        - "#^Constant DEPLOYER_VERSION not found\\.$#"
-        - "#^Constant DEPLOYER_BIN not found\\.$#"
         - "#^Constant MASTER_ENDPOINT not found\\.$#"
         - "#^Constant MASTER_TOKEN not found\\.$#"
```
staabm, line 23: `DEPLOYER_VERSION`, `DEPLOYER_BIN` will be detected by phpstan bootstrap.

`MASTER_ENDPOINT`, `MASTER_TOKEN` are only conditionally defined, therefore still needs the ignore

src/Configuration.php
```diff
@@ -17,6 +17,9 @@
 use function Deployer\Support\is_closure;
 use function Deployer\Support\normalize_line_endings;
 
+/**
+ * @implements \ArrayAccess<string, mixed>
```
staabm, line 21: reflects the signature of 
`function set(string $name, mixed $value)`

tests/constants.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+define('DEPLOYER_BIN', __DIR__ . '/../bin/dep');
```
staabm, line 3: extracting constant definitions in a separate file, so PHPStan can discover them

antonmedv (approved):

PR #4224, opened 2026-05-21
ENV vars and secrets have scalar values, and a string name

staabm:
this PR tries to reflect assumptions already in the code.
I think we can narrow these types.

or do use-cases exist with e.g. a array-value in ENV or secret?


antonmedv:
I think scalar is right here.

PR #4225, opened 2026-05-21
Most RunParams are readonly

staabm:
make value objects immutable where possible


antonmedv:
For some reason phpstan started failing ring after this: https://github.com/deployphp/deployer/commits/master/

staabm:
I will have a look,...

fix in https://github.com/deployphp/deployer/pull/4229

PR #4227, opened 2026-06-02
Add shared_dirs_populate option to populate shared dirs from release

adpeyre:
Adds a new `shared_dirs_populate` option (default: `false`) to the `deploy:shared` task.

  When enabled, each deploy copies any new files and subdirectories present in the release into the already-existing shared directory, without overwriting content already there. This is useful when new
  subdirectories are added to the repository over time and need to be propagated to the shared directory on the server.

  The copy uses `--no-clobber` (existing files are never overwritten) and `--no-dereference` (directory symlinks in the source are preserved as symlinks).

- [ ] Bug fix #…?
- [x] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [x] Docs added?

### DIFF
--- a/docs/recipe/deploy/shared.md
+++ b/docs/recipe/deploy/shared.md
@@ -23,8 +23,20 @@ set('shared_dirs', ['storage']);
 
 
 
+### shared_dirs_populate
+[Source](https://github.com/deployphp/deployer/blob/master/recipe/deploy/shared.php#L18)
+
+When true, populate existing shared dirs with any new files/dirs present in the release.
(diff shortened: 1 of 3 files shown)


antonmedv:
Would be cool to add test for this feature.

antonmedv:
And manybe just make it on by default.

adpeyre:
@antonmedv I added a test and changed the default value.

adpeyre:
@antonmedv Is it ok for you ?

PR #4229, opened 2026-06-03
Fix phpstan build

staabm:
the ignored errors no longer appear in latest phpstan, which got rid of the false positives

### DIFF
--- a/tests/phpstan-baseline.neon
+++ b/tests/phpstan-baseline.neon
@@ -1,21 +1,2 @@
 parameters:
 	ignoreErrors:
-		-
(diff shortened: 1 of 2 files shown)


tests/phpstan-baseline.neon
```diff
@@ -1,21 +1,2 @@
 parameters:
 	ignoreErrors:
-		-
-			message: "#^Comparison operation \"\\>\" between 100|125|200|100000 and 0 is always true\\.$#"
```
staabm, line 4: All these errors are related to `goto`. Phpstan learned how to analyze it recently

antonmedv:
Awesome! Thanks a lot @staabm

PR #4231, opened 2026-06-10
More precise `ObjectProxy` doctypes

staabm:
to prevent errors like 

> Call to an undefined method Host|ObjectProxy::setHostname().

in https://phpstan.org/r/e94b62bf-c573-45a5-9503-3fd6eff433b5 on PHPStan level 7+ in builder patterns like

```
host('production-aws1')
	->setHostname('myhost');
```

### DIFF
--- a/src/Support/ObjectProxy.php
+++ b/src/Support/ObjectProxy.php
@@ -10,13 +10,21 @@ declare(strict_types=1);
 
 namespace Deployer\Support;
 
+use Deployer\Host\Host;
+
+/**
+ * @mixin Host
+ */
 class ObjectProxy
 {
     /**
-     * @var array
+     * @var array<Host>
      */
     private $objects;
 
+    /**
+     * @param array<Host> $objects
+     */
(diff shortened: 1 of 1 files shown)


src/Support/ObjectProxy.php
```diff
@@ -10,13 +10,21 @@
 
 namespace Deployer\Support;
 
+use Deployer\Host\Host;
+
+/**
+ * @mixin Host
+ */
 class ObjectProxy
```
staabm, line 18: while this class has a pretty generic name and implementation, within deployer we only use it in combination with `Host` (or `LocalHost` which extends `Host`).

for better static analysis at the consumer side, I would love we can narrow this class with the proposed phpdocs

antonmedv (approved):

src/Support/ObjectProxy.php
antonmedv: We can rename to HostsProxy

PR #4234, opened 2026-06-15
[Microsoft Teams] Add Support for sending notifications through Workflows in Microsoft Teams

stefanzweifel:
This PR adds a new `ms-teams-workflows` recipe that adds support for sending notifications through Workflows in Microsoft Teams.

The existing `ms-teams` recipe relies on the "Microsoft 365 Connectors" API which has been deprecated by Microsoft [^1]. (The API actually still works, eventhough Microsoft mentions in their posts that the API has been disabled)

As the webhook URLs between "Microsoft 365 Connectors" and "Workflows in Teams" / "Power Automate" differ, I've opted to create a new recipe that provides an easy way for users to switch without creating a breaking change for users.

Migration is simple as changing the import and the variable.

```diff
-require 'contrib/ms-teams.php';
+require 'contrib/ms-teams-workflows.php';

-set('teams_webhook', 'https://******.webhook.office.com/***');
+set('teams_workflows_webhook', 'https://******.api.powerplatform.com:443/powerautomate/******');
```

We could also keep the names of variables and scripts the same ("teams" instead of "teams_workflows"), but that wouldn't allow using both implementations in parallel in case users would like to test both paths.

---

- [ ] Bug fix #…?
- [x] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [x] Docs added?

      $ php bin/docgen

[^1]: https://devblogs.microsoft.com/microsoft365dev/retirement-of-office-365-connectors-within-microsoft-teams/


antonmedv (approved):

stefanzweifel:
@antonmedv Would you be able to cut a new release for this? I know there aren't many changes between the last release (v8.0.5) and the default branch, but we're seeing more errors than usual from the MS Teams API and would like to switch to the new approach.

We could use our fork, but it would save us a lot of time and updating, if this change would just be released in deployer itself.

brianfreytag:
@antonmedv I'm also in need of this being added to a Release. Not getting Teams notifications anymore is becoming an issue.

brianfreytag:
@antonmedv We're still looking for a tagged release with this update, please.

antonmedv:
Will do!

brianfreytag:
> Will do!

Do you have an ETA for this? Thanks!

PR #4247, opened 2026-07-24
Add Fedora 43/44 provision recipe

devrfc83:
Mirrors recipe/provision.php (Ubuntu) using dnf instead of apt: native Fedora PHP packages (no Remi), MariaDB only (no MySQL), PostgreSQL via the official PGDG repo, firewalld instead of ufw, and Valkey/Memcached services. Verified end-to-end against a real Fedora 44 droplet, including PHP-FPM/Caddy socket permissions and SELinux (httpd_t) content labeling for deploy_path.

- [x] New feature?
- [ ] BC breaks?
- [ ] Tests added?
- [ ] Docs added?

### DIFF
--- a/recipe/provision-fedora.php
+++ b/recipe/provision-fedora.php
@@ -1,0 +1,216 @@
+<?php
+
+namespace Deployer;
+
+require __DIR__ . '/provision-fedora/databases.php';
+require __DIR__ . '/provision-fedora/nodejs.php';
+require __DIR__ . '/provision-fedora/php.php';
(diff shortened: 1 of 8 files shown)


devrfc83:
A little more context:

I have been using deployer for years now, but I always find myself struggling with provisioning, as I often deploy on Fedora instead of Ubuntu. Tested it against a Digital Ocean droplet and it provisioned the machine correctly. Don't hesitate to ping me if it requires some changes.

